PR #28, opened 2020-04-21
Enable user namespaces for non-root users

liorokman:

This PR makes it possible for non-root users to use user namespaces.

User namespaces are used by rootless container tools. For example, [buildah](https://github.com/containers/buildah/), [img](https://github.com/genuinetools/img/), [rootless podman](https://github.com/containers/libpod), and more. 

Most distributions have this support automatically enabled - the Debian project disables this in their kernel builds. Enabling this will restore parity with CoreOS and SuSE.

**Release note**:

```improvement user
User namespaces are enabled for non-root users.
```


marwinski:
also approved by @gehoern

PR #42, opened 2020-04-30
Adapt tekton task to integration test --iaas option

dguendisch:
Adapts the tekton task for the integration tests to respect the newly introduced `--iaas` flag.
The taskrun now demonstrates the usage with the `aws` infrastructure.

Only when we integrate the tasks into a pipeline can we run e.g. an aws- and a gcp-task in parallel.

### DIFF
--- a/ci/integrationtest-run.yaml
+++ b/ci/integrationtest-run.yaml
@@ -1,11 +1,14 @@
 apiVersion: tekton.dev/v1beta1
 kind: TaskRun
 metadata:
-  generateName: it-test-
+  generateName: it-test-aws-
 spec:
   serviceAccountName: build-bot
   taskRef:
     name: integrationtest-gardenlinux-task
+  params:
+  - name: iaas
+    value: aws # or gcp
   resources:
     inputs:
     - name: repo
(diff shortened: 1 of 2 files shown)


ci/integrationtest-run.yaml
```diff
@@ -1,11 +1,14 @@
...
 metadata:
-  generateName: it-test-
+  generateName: it-test-aws-
 spec:
   serviceAccountName: build-bot
   taskRef:
     name: integrationtest-gardenlinux-task
+  params:
+  - name: iaas
+    value: aws # or gcp
```
msohn, line 11: how about also running it on gcp ?

dguendisch: The taskrun is only an example to verify that the task _can_ execute in tekton.
To integrate it properly we need a pipeline (there you can run one task for aws and one for gcp in parallel).
However the pipeline integration with tekton does not make too much sense currently, as we don't have the means to get e.g. the ami-id or the gcp image-name out. Only with this information we can finally include the tests in some meaningful pipeline.

msohn:
lgtm

PR #63, opened 2020-05-27
Enable project quotas on the ROOT filesystem

### LINKED ISSUE #62: Make it possible to use Project Quotas on GardenLinux

Kubernetes allows defining limits on the amount of disk space used by ephemeral storage. Kubelet implements this feature in two ways:
1. A periodic task that scans directories
1. Using a kernel feature called Project Quotas where the kernel keeps track of the disk usage.

Full details are here: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#local-ephemeral-storage 

In order to make it possible to enable the second method, the filesystem needs to have the project quota feature turned on. This needs to be done when the filesystem is unmounted, so it can't be done after GardenLinux has already booted.

**Why is this needed**:

Directory scanning is not a very accurate way to account for ephemeral filesystem usage. It is racy, and it fails to account for files that are deleted from the filesystem, but are still held open by any process (i.e. the dnode is still allocated, but the inode is already free).

Adding this feature makes it possible for projects using GardenLinux to opt in to the second method supported by Kubelet. A project that wants this must still opt-in by enabling the relevant featuregate.

liorokman:

This PR enables project quotas on the ROOT filesystem for GardenLinux images.

Fixes #62 

**Release note**:

```improvement user
GardenLinux now supports enabling the kubelet `LocalStorageCapacityIsolationFSQuotaMonitoring` feature gate. 
```

### DIFF
--- a/bin/makef.sh
+++ b/bin/makef.sh
@@ -104,7 +104,7 @@ mkfs.ext4 -L ROOT -O quota -E lazy_itable_init=0,lazy_journal_init=0,quotatype=u
 #tune2fs -c 0 -i 0 ${loopback}p3
 
 echo "### mounting filesystems"
-mkdir -p ${dir_name}          && mount ${loopback}p4 -o prjquota ${dir_name}
+mkdir -p ${dir_name}          && mount ${loopback}p4 ${dir_name}
 mkdir -p ${dir_name}/boot/efi && mount ${loopback}p2 ${dir_name}/boot/efi
 mkdir -p ${dir_name}/usr      && mount ${loopback}p3 ${dir_name}/usr
 
--- a/bin/makef.sh
+++ b/bin/makef.sh
@@ -25,6 +25,24 @@ help() {
 	exit 1
 }
 
+
+# Run fsck 3 times on an ext4 image. Sometimes once is just not enough, but give up 
+# if it takes more than 3 times.
+fsckExt4() {
+
+  local device=$1 ; shift
+  local retries=0
+
+  while ! fsck.ext4 -f -a -E discard,inode_count_fullmap -D ${device} ; do
+    echo Errors detected, retrying the fsck.
+    retries=$(($retries + 1))
+    if [ $retries -gt 3 ]; then
+      return 1
+    fi
+  done
+
+}
+
 target="bios,uefi"
 fs_check=1
 force=0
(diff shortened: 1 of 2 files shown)


marwinski:
/lgtm

I have built an image and started a VM. This looks ok but as I don't really understand quota (at least not while in a parallel meetings) I would ask @gehoern to decide and merge.

bin/makef.sh
```diff
@@ -132,8 +150,8 @@ sync
 
 echo "### final fsck, just to be sure"
 fsck.vfat -f -a ${loopback}p2
-fsck.ext4 -f -a -E discard,inode_count_fullmap -D ${loopback}p3
```
gehoern, line 135: did you had to remove this ? the inode_count_fullmap makes sure the inode table is completely reordered. Does it cause problems ?

liorokman: I didn't remove it. I simply moved it to the function.

bin/makef.sh
```diff
@@ -81,12 +99,12 @@ echo "### reconnected loopback to ${loopback}"
 echo "### creating filesystems"
 mkfs.vfat -n EFI ${loopback}p2
 mkfs.ext4 -L USR -E lazy_itable_init=0,lazy_journal_init=0 ${loopback}p3
-mkfs.ext4 -L ROOT -E lazy_itable_init=0,lazy_journal_init=0 ${loopback}p4
+mkfs.ext4 -L ROOT -O quota -E lazy_itable_init=0,lazy_journal_init=0,quotatype=usrquota:grpquota:prjquota  ${loopback}p4
 # part of debian-cloud-images, I am sure we want that :-) -> it is default
 #tune2fs -c 0 -i 0 ${loopback}p3
 
 echo "### mounting filesystems"
-mkdir -p ${dir_name}          && mount ${loopback}p4 ${dir_name}
```
gehoern, line 89: here on mount of the fs i think we do not need this right? it would be needed later in the OS like you specified later ... but not here

liorokman: I wasn't sure. I added the mount option just to verify that even when the image is being build, the project quota would be kept up to date. Seems to be not expensive to do, and might prevent issues later.

bin/makef.sh
gehoern: oh sorry was blind :-D

bin/makef.sh
gehoern: yeah but is not supported in my next pr :-D

liorokman: Let me check, if it's not required I'll remove it.

gehoern: top

liorokman: If the filesystem is mounted without the `prjquota` flag, then anything added to the filesystem at that time won't be tracked. 

If the filesystem is remounted with the flag, any files added later are correctly tracked.

For this use-case, that behavior is good enough, so I'll remove the flag from the `mount` operation in the script.

Issue #52, opened 2020-05-19
Can't log on with admin on Alicloud


jia-jerry:
**What happened**:
Can't log on with admin on Alicloud. Cloud-init failed.
**What you expected to happen**:
**How to reproduce it (as minimally and precisely as possible)**:
Latest master branch.
**Anything else we need to know**:
I used to build an image with alicloud branch. It worked. Below is the comparison of the cloud-init log.
Success one:
```
2020-05-18 17:57:35,896 - handlers.py[DEBUG]: finish: init-network/check-cache: SUCCESS: no cache found
2020-05-18 17:57:35,896 - util.py[DEBUG]: Attempting to remove /var/lib/cloud/instance
2020-05-18 17:57:35,897 - stages.py[DEBUG]: Using distro class <class 'cloudinit.distros.debian.Distro'>
2020-05-18 17:57:35,897 - __init__.py[DEBUG]: Looking for data source in: ['AliYun', 'None'], via packages ['', 'cloudinit.sources'] that matches dependencies ['FILESYSTEM', 'NETWORK']
2020-05-18 17:57:35,899 - __init__.py[DEBUG]: Searching for network data source in: ['DataSourceAliYun', 'DataSourceNone']
2020-05-18 17:57:35,900 - handlers.py[DEBUG]: start: init-network/search-AliYun: searching for network data from DataSourceAliYun
2020-05-18 17:57:35,900 - __init__.py[DEBUG]: Seeing if we can get any data from <class 'cloudinit.sources.DataSourceAliYun.DataSourceAliYun'>
2020-05-18 17:57:35,900 - __init__.py[DEBUG]: Update datasource metadata and network config due to events: New instance first boot
2020-05-18 17:57:35,900 - util.py[DEBUG]: Running command ['systemd-detect-virt', '--quiet', '--container'] with allowed return codes [0] (shell=False, capture=True)
2020-05-18 17:57:35,905 - util.py[DEBUG]: Running command ['running-in-container'] with allowed return codes [0] (shell=False, capture=True)
2020-05-18 17:57:35,907 - util.py[DEBUG]: Running command ['lxc-is-container'] with allowed return codes [0] (shell=False, capture=True)
2020-05-18 17:57:35,908 - util.py[DEBUG]: Reading from /proc/1/environ (quiet=False)
2020-05-18 17:57:35,908 - util.py[DEBUG]: Read 153 bytes from /proc/1/environ
[...]

PR #64, opened 2020-05-27
Alicloud fix

### LINKED ISSUE #53: It takes about 20 mins for Gardenlinux to be booted on Alicloud
**What happened**:
It takes about 20 mins for Gardenlinux to be booted on Alicloud
It takes less than 1 min for Gardenlinux to be booted on Alicloud
Use master branch
Alicloud found it is stuck in step `initialize debian`. They suggested we follow the steps in this [doc](https://www.alibabacloud.com/help/doc-detail/48226.htm?spm=a2c63.p38356.b99.203.5f7a751eVcTS1f) to build a customized image. I'm not sure virtio is complied with[ what Alicloud suggested](https://www.alibabacloud.com/help/doc-detail/62423.htm?spm=a2c63.p38356.879954.8.7ac11541cVs3Nd#concept-dvq-cqs-xdb). 
**Environment**: Alicloud

jia-jerry:
Fix bootstrap issues on Alicloud
Fixes #53 
#52 

**Release note**:

```improvement operator

```

### DIFF
--- a/bin/make-ali-ami
+++ b/bin/make-ali-ami
@@ -5,10 +5,10 @@ VERSION=43
 region=eu-central-1
 aliyun ecs ImportImage \
  --Description="gardenlinux-dev-ali-$VERSION" \
- --Platform=Debian \
+ --Platform='Others Linux' \
  --ImageName="gardenlinux-dev-ali-$VERSION" \
  --DiskDeviceMapping.1.OSSBucket=gardenlinux-development \
- --DiskDeviceMapping.1.OSSObject="gardenlinux/garden-linux-dev-ali-43.qcow2" \
+ --DiskDeviceMapping.1.OSSObject="gardenlinux/garden-linux-dev-ali-${VERSION}.qcow2" \
  --DiskDeviceMapping.1.Format=qcow2 \
  --DiskDeviceMapping.1.DiskImSize=5 \
  --RegionId=$region
--- a/features/ali/file.include/etc/cloud/cloud.cfg.d/ds-identity.cfg
+++ b/features/ali/file.include/etc/cloud/cloud.cfg.d/ds-identity.cfg
@@ -1,2 +1,0 @@
-datasource: AliYun
-policy: enabled
--- a/features/ali/pkg.include
+++ b/features/ali/pkg.include
@@ -2,3 +2,4 @@ python3-cffi-backend
 python3-boto
 cloud-init
 dmidecode
+rng-tools-debian
--- a/features/ali/pkg.include
+++ b/features/ali/pkg.include
@@ -2,4 +2,3 @@ python3-cffi-backend
 python3-boto
 cloud-init
 dmidecode
-rng-tools-debian


jia-jerry:
Not to enable `rng-tools-debian.service` due to
```
May 27 02:01:51 gothers systemd[1]: rng-tools-debian.service: Control process exited, code=exited, status=1/FAILURE
May 27 02:01:51 gothers systemd[1]: rng-tools-debian.service: Failed with result 'exit-code'.
May 27 02:01:51 gothers systemd[1]: Failed to start LSB: rng-tools (Debian variant).
```

bin/make-ali-ami
```diff
@@ -5,10 +5,10 @@ VERSION=43
 region=eu-central-1
 aliyun ecs ImportImage \
  --Description="gardenlinux-dev-ali-$VERSION" \
- --Platform=Debian \
+ --Platform='Others Linux' \
```
zanetworker, line 8: just from the issue discription I saw you quoting `OtherLinux` instead, does wording matter, or any other random value other than `Debian` would do that job?

gehoern: hmm infact i think we are a debian ... from the technical site ... so i do not know if the implications of "Others Linux" are bad for us

jia-jerry: > just from the issue discription I saw you quoting `OtherLinux` instead, does wording matter, or any other random value other than `Debian` would do that job?

It should be `Others Linux`. It is a default value. If it is set Debian, it will fail to bootstraped within desired time. Alicloud inject something for different OS types. e.g. I tried to set type `CoreOS`, it could be booted quickly but failed in ssh connection.

jia-jerry: > hmm infact i think we are a debian ... from the technical site ... so i do not know if the implications of "Others Linux" are bad for us

I could check with Alicloud

bin/make-ali-ami
```diff
@@ -5,10 +5,10 @@ VERSION=43
...
  --ImageName="gardenlinux-dev-ali-$VERSION" \
  --DiskDeviceMapping.1.OSSBucket=gardenlinux-development \
- --DiskDeviceMapping.1.OSSObject="gardenlinux/garden-linux-dev-ali-43.qcow2" \
+ --DiskDeviceMapping.1.OSSObject="gardenlinux/garden-linux-dev-ali-${VERSION}.qcow2" \
```
ccwienk, line 11: good catch

bin/make-ali-ami
gehoern: hmm it is an official debian we have nothing from extern - all debian. BUT we change the default behavior e.g. debian has some stuff that is systemd and some is not -> networking is not systemd based it is the old dhclient stuff -> that we switched to pure systemd

gehoern: but all in all it is a debian

marwinski: I am sue it has nothing to do with that. Alicloud should explain what the mean by "Debian" and "other" and what precisely takes 20 minutes. Maybe they try some optimisations for the 4.x kernel that fails because we use a 5.4 kernel !?

jia-jerry: I have already created another ticked to Alicloud. Also put all these pieces of information to the ticket.

jia-jerry: Just got the reply:

> IF you use "Platform=“Others Linux”" on Debian image, it will cause some default setting being unset, such as software source, ntp, network, etc.. Thank you.

marwinski: I don't understand. What will be unset. NTP server is configured in the image, network is handled by dhcp, there is no software source ...

jia-jerry: I did a test for Image Platform=“Others Linux” and make it running in Alicloud Gardener cluster, it works.

PR #70, opened 2020-06-02
set ntp server for Alicloud

jia-jerry:
set ntp server for Alicloud
Fixes #

**Release note**:

```improvement operator

```


gehoern:
would prefer we do this like for the other providers, but good for now

marwinski:
How are we doing this for the other providers. Natural would be to use the DHCP option (which is probably provided by the other cloud providers). Ali does not provide this

PR #80, opened 2020-06-05
Prepare full automation of the Azure Marketplace process

dkistner:
Prepare full automation of the publishing/transport process of `gardenlinux` on the Azure Marketplace.

The entire process has a pretty long test turnaround of a few hours or in the worst case days...

:information_source: As mentioned this is a preparation to enable the full automation of the process.
It is not yet tested once end to end.

@ccwienk, @AndreasBurger  We would need to poll this function `check_offer_transport_state()` e.g. once per hour. Need to be defined how we technically do that.
In addition we need to store the ids of the transport operations to `staging` (publish operation) and `production` (golive operation) to configure `check_offer_transport_state()` properly (which we need to call during the polling). Please checkout the TODOs.

Seems there are a few inconsistencies in the documentation regarding possible operation states, response values etc.
https://docs.microsoft.com/en-us/azure/marketplace/cloud-partner-portal-orig/cloud-partner-portal-api-overview

**Release note**:

```NONE
```


ci/glci/azure.py
```diff
@@ -89,6 +91,17 @@ def get_image_url(self, image_name: str):
         return f"https://{self.sa_name}.blob.core.windows.net/{self.container_name}/{image_name}?{container_sas}"
 
 
+class Azmp_operation_state(Enum):
```
ccwienk, line 94: nit: according to PEP8, use CamelCase for class names (snake_case for variables..)

ci/glci/azure.py
```diff
@@ -162,6 +175,51 @@ def publish_offer(self, publisher_id: str, offer_id: str, notification_mails):
             )
         response.close()
 
+    def fetch_ongoing_operation_id(self, publisher_id: str, offer_id: str, transport_dest: Azmp_transport_dest):
+        '''Fetches the id of an ongoing Azure Marketplace transport operation to a certain transport destination.'''
+
+        response = requests.get(
+            f"https://cloudpartner.azure.com/api/publishers/{publisher_id}/offers/{offer_id}/submissions",
```
ccwienk, line 182: could you factor out the api-routes (deduplicate base URL, separate URL construction from code using them)? e.g. like so: https://github.com/gardener/cc-utils/blob/master/checkmarx/client.py#L30 (there are probably also other patterns one could apply, of course)

ci/glci/azure.py
```diff
@@ -269,11 +327,73 @@ def update_offer(
...
     )
 
+    # TODO Persist the publish_operation_id to start polling the transport operation status to staging.
+    publish_operation_id = marketplace_client.fetch_ongoing_operation_id(
+        publisher_id,
+        offer_id,
+        Azmp_transport_dest.STAGING,
+    )
+
+def check_offer_transport_state(
```
ccwienk, line 342: should this function not rather return something (e.g. an enum value)?

ci/glci/azure.py
```diff
@@ -89,6 +91,17 @@ def get_image_url(self, image_name: str):
...
 
+class Azmp_operation_state(Enum):
+    NOTSTARETD = "notStarted"
+    RUNNING = "running"
+    COMPLETED = "completed"
+    SUCCEEDED = "succeeded"
+    FAILED = "failed"
+
+class Azmp_transport_dest(Enum):
+    STAGING = "staging"
```
ccwienk, line 102: are those all possible values? are those predefined by azure?

ci/glci/az.py
```diff
@@ -0,0 +1,464 @@
+import requests
```
ccwienk, line 1: nit: order imports alphabetically (but grouped)

ci/glci/az.py
```diff
@@ -0,0 +1,464 @@
...
+    }
+    return spec
+
+
+def remove_image_version_from_plan(spec: dict, plan_id: str, image_version: str, image_url: str):
+    """remove an image version from a given plan and return a modified offer spec."""
+
+    plan_spec = {}
+    for plan in spec["definition"]["plans"]:
+        if plan["planId"] == plan_id:
```
ccwienk, line 262: since this is now the second implementation of plan-lookup, I would actually suggest to factor this out; maybe it could be included into the function or method that returns the initial definition-dict

dkistner:
@ccwienk Thanks for the feedback. I just briefly checked it and not in all detail yet. It seems to me that the suggestions are primarily regarding python syntactic/styling improvements and for reducing boiler plate etc. 

That's valid and I will address it, but could you please check and comment if the Azure logic fit into the gardenlinux ci logic? Are there any pieces missing or wrong so that we can't make use of it in the pipeline? 

I added some dataclasses for Azure in `glci.model` like `AzurePublishCfg` which include `AzureMarketplaceCfg`, `AzureServicePrincipalCfg` and `AzureStorageAccountCfg`.
They are all part of `glci.model.CicdCfg`. I see this config is generated here via a file read https://github.com/gardenlinux/gardenlinux/blob/main/ci/glci/util.py#L25-L40
That means the config file need to contain the proper information. Where is this config file coming from? I guess from the secret server?

I also extended the the `glci.model.ReleaseManifest` class with an Azure specific published image class. This one is a parent class of `glci.model.OnlineReleaseManifest`. Which is passed back from `glci.az.upload_and_publish_image`. How is this post-processed? Does it the information in it just get persistent in on S3?

Sorry for asking that, but the whole process is not completely clear to me. So far I was only able to test the Azure marketplace logic isolated, but not the entire flow. Therefore feedback would be helpful.

ccwienk:
@dkistner : 
> Where is this config file coming from? I guess from the secret server?
yes, exactly; you might want to also get in touch w/ @AndreasBurger 

As for the (Online)`ReleaseManifest`:

The idea (to be documented..) looks like so: Gardenlinux is built in different "flavours" (defined by a combination of n `features` + exactly one `platform` (defined in `features/<feature-name>`). One such flavour would e.g. be an image for Azure. Flavours are grouped into "flavour sets" (defined in `flavours.yaml`).

Each flavour being built in the build pipeline is accompanied by a `manifest` (model class being the `ReleaseManifest` you mentioned), which is ultimately uploaded to our build results storage (S3 bucket currently). The difference towards the `OnlineReleaseManifest` being that the latter also contains the bucket+key coordinates of the store to which they were uploaded (this is not part of the actual manifest, because manifests are intended to be copied, e.g. to express promotion of a snapshot to e.g. a release).

The build pipeline uses the manifests to determine whether or not a given build flavour has already been published; it also uses the contained metadata e.g. for the image uploads.

ci/glci/model.py
```diff
@@ -401,14 +409,32 @@ class BuildCfg:
...
+    name: str
+    container_name: str
+    access_key: str
+
+
+@dataclasses.dataclass(frozen=True)
+class AzurePublishCfg:
+    marketplace: AzureMarketplaceCfg
+    service_principal: AzureServicePrincipalCfg
+    storate_account: AzureStorageAccountCfg
```
ccwienk, line 437: typo; I suppose, this should rather read `storage_account`?

ci/glci/model.py
```diff
@@ -272,6 +272,14 @@ class GcpPublishedImage(PublishedImageBase):
     gcp_project_name: str
 
 
+@dataclasses.dataclass(frozen=True)
+class AzurePublishedImage:
+    transport_state: str
+    urn: str
+    publish_operation_id: str
```
ccwienk, line 279: do we actually need to persist the publish_operation_id?

ci/glci/model.py
```diff
@@ -272,6 +272,14 @@ class GcpPublishedImage(PublishedImageBase):
     gcp_project_name: str
 
 
+@dataclasses.dataclass(frozen=True)
+class AzurePublishedImage:
+    transport_state: str
+    urn: str
```
ccwienk, line 278: I know, the other dataclasses are also very sparse with documentation.

however, could you add a brief doc-str explaining those attrs (e.g. what the urn refers to, which transport_states there are..); also, it might be nice to use an enum for TransportState

ci/glci/model.py
dkistner: Hmm I think so. It is an output of the first step (upload image. + trigger publish process). In the second step we need this id to poll the status of the publishing process. Same for the `golive_operation_id`

ci/glci/az.py
dkistner: done

ci/glci/az.py
```diff
@@ -0,0 +1,464 @@
...
+    marketplace_baseurl = "https://cloudpartner.azure.com/api/publishers"
+
+    def __init__(self, spn_tenant_id: str, spn_client_id: str, spn_client_secret: str):
+        app_client = ConfidentialClientApplication(
+            spn_client_id,
+            authority=f"https://login.microsoftonline.com/{spn_tenant_id}",
+            client_credential=spn_client_secret
+        )
+        token = app_client.acquire_token_for_client(scopes="https://cloudpartner.azure.com/.default")
+        self.req_headers = {
```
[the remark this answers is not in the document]
dkistner: Unfortunately there is no SDK for the Azure Marketplace apis. So we need to use direct http calls.

I added an internal `_request` method.

ci/glci/az.py
```diff
@@ -0,0 +1,464 @@
...
+    def _api_url(self, *parts):
+        url_parts = "/".join(p for p in parts)
+        return f"{self.marketplace_baseurl}/{url_parts}"
+
+    def fetch_offer(self, publisher_id: str, offer_id: str):
+        """Fetch an offer from Azure marketplace."""
+
+        response = requests.get(
+            self._api_url(publisher_id, "offers", offer_id),
+            headers=self.req_headers,
```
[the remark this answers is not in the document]
dkistner: done

ci/glci/az.py
```diff
@@ -0,0 +1,464 @@
...
+            params=self.req_params
+        )
+        if not response.ok:
+            raise RuntimeError(
+                "Fetching of Azure marketplace offer for gardenlinux failed. "
+                f"statuscode={response.status_code}"
```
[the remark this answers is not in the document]
dkistner: Hmm I already consider `response.raise_for_status()` but afaik it does not have the possibility to pass custom error message. I added an own method `_raise_for_status()` for this use case. I think it make sense to pass more context in error cases.

ci/glci/az.py
```diff
@@ -0,0 +1,464 @@
...
+            )
+        offer_spec = response.json()
+        response.close()
```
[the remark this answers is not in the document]
dkistner: done

ci/glci/az.py
```diff
@@ -0,0 +1,464 @@
...
+        return offer_spec
+
+    def update_offer(self, publisher_id: str, offer_id: str, spec: dict):
+        """Update an offer with a give spec."""
+
+        headers = self.req_headers.copy()
```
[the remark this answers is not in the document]
dkistner: done

ci/glci/az.py
```diff
@@ -0,0 +1,464 @@
...
+                )
+
+
+def add_image_version_to_plan(
+    spec: dict,
+    plan_id: str,
+    image_version: str,
+    image_url: str
+):
+    """Add a new image version to a given plan and return a modified offer spec."""
```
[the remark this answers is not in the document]
dkistner: done

ci/glci/az.py
dkistner: done

ci/glci/model.py
dkistner: Added some doc-str and add an enum class `glci.model.AzureTransportState`.

ci/glci/model.py
dkistner: done

ci/glci/az.py
```diff
@@ -0,0 +1,464 @@
...
+        self.req_headers = {
+            "Authorization": f"Bearer {token['access_token']}",
+            "Content-Type": "application/json",
+        }
+        self.req_params = {
+            "api-version": "2017-10-31"
+        }
+
+    def _api_url(self, *parts):
+        url_parts = "/".join(p for p in parts)
```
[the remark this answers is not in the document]
dkistner: done

dkistner:
Closes in favour of https://github.com/gardenlinux/gardenlinux/pull/115

PR #82, opened 2020-06-06
Added a default-ulimit setting for max memory lock size of 64Mb.

### LINKED ISSUE #81: Running Golang binaries fails in GardenLinux based Docker containers for Golang >= 1.14
**What happened**:

While running ArgoCD on a GardenLinux cluster, the container eventually crashes with the following error message: 

> runtime: mlock of signal stack failed: 12
> runtime: increase the mlock limit (ulimit -l) or
> runtime: update your kernel to 5.3.15+, 5.4.2+, or 5.5+
> fatal error: mlock failed

The container should not crash.

Run ArgoCD of the later versions (compiled with Golang 1.14.1) on a GardenLinux based Kubernetes cluster.

After a little digging, the root cause of this issue is the following Golang bug: https://github.com/golang/go/issues/37436 coupled with a bad default setting for the Docker runtime engine and coupled with the way that Debian-packaged kernels report their version.

1. Linux kernel version 5.2.something introduced a memory corruption bug that got fixed in version 5.4.2. See https://github.com/golang/go/wiki/LinuxKernelSignalVectorBug for details.
1. Golang 1.14 introduced a feature that got bit by the kernel bug above, assuming the kernel itself was compiled with GCC 9.x (which is the case for Debian Bullseye). 
1. Since versions above 5.4.2 are safe, and versions below are not, Golang introduced a workaround for non-patched kernel versions, where the stack memory is `mlock(2)`-ed into memory.. That workaround is automatically enabled if the kernel version is detected to be a vulnerable kernel version.
1. Debian (and Ubuntu) keep the reported kernel version stable, due to multiple software packages depending on a stable version. The base version reported by `uname -r` is `5.4.0-4-cloud-amd64`, but if you ran `uname -v` you would get the real version (`#1 SMP Debian 5.4.19-1 (2020-02-13)` in this case) which is already patched.
1. Golang looks at the wrong field in the uname response, and misidentifies the kernel version as 5.4.0, which is still buggy. Golang then enables the `mlock` workaround, which fails because the Docker daemon launches all containers with a ulimit of 64KB for max locked memory, instead of the system-wide default which is 64MB.

The fix is to modify the way docker is configured on the host, and adding a "default-ulimits" section to the /etc/docker/daemon.json file with the correct value (64M instead of 64K). 

See item (2) in  https://github.com/golang/go/issues/37436#issuecomment-593315935 for an example of what to put inside the `daemon.json` file, but instead of setting the locked memory limit to unlimited, I suggest keeping it at the system default, which is 64Mb instead of 64Kb.

liorokman:

Fixes #81 

Based on the suggestion in item (2) in https://github.com/golang/go/issues/37436#issuecomment-593315935 .

**Release note**:
```noteworthy user
Allows Golang 1.14 binaries to run correctly by increasing the default max memory lock ulimit size for Docker containers.
```

### DIFF
--- a/features/gardener/file.include/etc/docker/docker.conf
+++ b/features/gardener/file.include/etc/docker/docker.conf
@@ -1,2 +1,9 @@
-{ "storage-driver": "overlay2"}
-
+{ "storage-driver": "overlay2",
+  "default-ulimits": {
+      "memlock": {
+          "name": "memlock",
+          "hard": 65536,
+          "soft": 65536
+      }
+  }
+}


liorokman:
This should solve the issue for Dockerd based installations - not sure about Containerd based installations.

marwinski:
This puzzled me for some time. The go code uses ```uname -r``` which reports ```5.4.0-4-cloud-amd64``` and it thinks it has got the kernel bug and activates the workaround (although it is kernel ```5.4.19```). I hope this workaround for the workaround fixes the workaround.

PR #86, opened 2020-06-08
Remove containerd config

danielfoehrKn:

By installing the debian package containerd, Gardenlinux already contains the file containerd config file in /etc/containerd/config.toml.
Gardener does not override an existing containerd config, and as a consequence does not generate the default configuration file.

This is currently not a problem, but could become one in the future, when Gardener adds support for containerd configuration via the Shoot.yaml.

Fixes #

I tested `containerd` running successfully with `gardenlinux` using a  version of the Gardener that [includes this fix](https://github.com/gardener/gardener/commit/6886fbe56daebba7d7537e13d710747149deb85b)  together with [this PR version of the `os-gardenlinux`](https://github.com/gardener/gardener-extension-os-gardenlinux/pull/5)  extension.
Also `gvisor` seemed to have no issues.

**Release note**:

```improvement operator

```

### DIFF
--- a/features/gardener/exec.config
+++ b/features/gardener/exec.config
@@ -16,3 +16,6 @@ systemctl mask nftables.service
 #
 systemctl disable docker
 systemctl disable containerd
+
+# Delete the default containerd config file. Let Gardener create it.
+rm -rf /etc/containerd/config.toml


marwinski:
I would be reluctant do delete the containerd configuration in the image. I have tested containerd "by hand" which probably would not have been possible without the configuration file. This would mean shipping an image with a broken software package (unless and please correct me if containerd happily runs without that file). I am also curious why Gardener would not adapt/overwrite that file which would be the correct way forward.

danielfoehrKn:
If containerd does not find its config file, it will just use its default configuration (the same that Gardener currently generates anyways). So yes, containerd does not need the config.toml.

> I am also curious why Gardener would not adapt/overwrite that file which would be the correct way forward

I am not 100% sure about the original reason why it was implemented like this. 
In any case, I think there is either the option to completely override it or respect an existing config file. Merging would be impractical.

I can think of the following reason why it might be beneficial to have the ability to manually set the containerd configuration.
 - testing purposes (not really a reason)
 - supporting configurations options for containerd that are not supported (yet) by Gardener or are purposefully not configurable through Gardener.
 - fixing the container runtime configuration for a specific OS version e.g for a "hardened" setup. E.g  forcing `SELinux` to be enabled. This could be an issue for gov cloud deployments.

>  which would be the correct way forward

Why do you think this is the correct way forward for Gardener to go?

PR #91, opened 2020-06-12
Add OpenStack image build upload procedure

dkistner:
Add procedure to upload gardenlinux images to OpenStack environments.

~~For each OpenStack environment we need to create a `Openrc` object with proper access information. This need to be passed for each environment to the function `upload_image()` together with the name of the image, builddate and the path to the image vmdk file in the local filesystem.~~

Updated the PR. For each OpenStack environment we need a `glci.model.CicdCfg` which contain internally an `glci.model.Openrc` object with access information and image properties. The image is now uploaded directly via web url (e.g. S3 object url) to Glance instead of uploading it from file system.
It's now also integrated with the publish process.

cc @ccwienk 

**Release note**:

```improvement operator
Add image upload procedure for OpenStack environments.
```

### DIFF
--- a/ci/glci/model.py
+++ b/ci/glci/model.py
@@ -596,3 +596,13 @@ def feature_by_name(feature_name: str):
         if feature.name == feature_name:
             return feature
     raise ValueError(feature_name)
+
+
+@dataclasses.dataclass(frozen=True)
+class OpenRC:
+    auth_url: str
+    domain: str
+    region: str
+    project_name: str
+    username: str
+    password: str
--- a/ci/glci/model.py
+++ b/ci/glci/model.py
@@ -273,6 +273,13 @@ class GcpPublishedImage(PublishedImageBase):
 
 
 @dataclasses.dataclass(frozen=True)
+class OpenstackPublishedImage:
+    region_name: str
+    image_id: str
+    image_name: str
+
+
+@dataclasses.dataclass(frozen=True)
 class ReleaseManifest(ReleaseIdentifier):
     '''
     metadata for a gardenlinux release variant that can be (or was) published to a persistency
@@ -281,7 +288,7 @@ class ReleaseManifest(ReleaseIdentifier):
     build_timestamp: str
     paths: typing.Tuple[typing.Union[S3_ReleaseFile]]
     published_image_metadata: typing.Union[AlicloudPublishedImageSet,
-                                           AwsPublishedImageSet, GcpPublishedImage, None]
+                                           AwsPublishedImageSet, GcpPublishedImage, OpenstackPublishedImage, None]
 
     def path_by_suffix(self, suffix: str):
         for path in self.paths:
@@ -412,8 +419,25 @@ class AzurePublishCfg:
 
 
 @dataclasses.dataclass(frozen=True)
+class OpenstackEnviroment:
+    auth_url: str
+    domain: str
+    region: str
+    project_name: str
+    username: str
+    password: str
+
+
+@dataclasses.dataclass(frozen=True)
+class OpenstackPublishCfg:
+    openrc: OpenstackEnviroment
+    properties: dict
+
+
+@dataclasses.dataclass(frozen=True)
 class PublishCfg:
     azure: AzurePublishCfg
+    openstack: OpenstackPublishCfg
 
 
 @dataclasses.dataclass(frozen=True)
@@ -597,12 +621,3 @@ def feature_by_name(feature_name: str):
             return feature
     raise ValueError(feature_name)
 
-
-@dataclasses.dataclass(frozen=True)
-class OpenRC:
-    auth_url: str
-    domain: str
-    region: str
-    project_name: str
-    username: str
-    password: str
--- a/ci/glci/openstack_image.py
+++ b/ci/glci/openstack_image.py
@@ -1,82 +1,102 @@
-import os
-import time
-import argparse
-
-from glci.model import OpenRC
+import dataclasses
+import functools
+from datetime import datetime
+from time import sleep
 
 from openstack import connect
 
+import glci
+
 class OpenstackImageUploader:
-  '''OpenstackImageUploader is a client to upload images to Openstack Glance.'''
+    '''OpenstackImageUploader is a client to upload images to Openstack Glance.'''
 
-  def __init__(self, openrc: OpenRC):
-      self.conn = connect(
-          auth_url=openrc.auth_url,
-          project_name=openrc.project_name,
-          username=openrc.username,
-          password=openrc.password,
-          region_name=openrc.region,
-          user_domain_name=openrc.domain,
-          project_domain_name=openrc.domain,
-      )
+    def __init__(self, openrc: glci.model.OpenstackEnviroment):
+        self.openstack_env = openrc
 
-  def upload_image(self, name: str, path: str, meta: dict, timeout_seconds=86400):
-      '''Upload an image from filesystem to Openstack Glance.'''
+    @functools.lru_cache
+    def _get_connection(self):
+        return connect(
+            auth_url=self.openstack_env.auth_url,
+            project_name=self.openstack_env.project_name,
+            username=self.openstack_env.username,
+            password=self.openstack_env.password,
+            region_name=self.openstack_env.region,
+            user_domain_name=self.openstack_env.domain,
+            project_domain_name=self.openstack_env.domain,
+        )
 
-      image = self.conn.image.create_image(
-          name=name,
-          filename=path,
-          disk_format='vmdk',
-          container_format='bare',
-          visibility='community',
-          meta=meta,
-          timeout=timeout_seconds,
-      )
-      return image['id']
+    def upload_image_from_fs(self, name: str, path: str, meta: dict, timeout_seconds=86400):
+        '''Upload an image from filesystem to Openstack Glance.'''
 
-  def wait_image_ready(self, image_id: str, wait_interval_seconds=10):
-      '''Wait until an image get in ready state.'''
+        conn = self._get_connection()
+        image = conn.image.create_image(
+            name=name,
+            filename=path,
+            disk_format='vmdk',
+            container_format='bare',
+            visibility='community',
+            timeout=timeout_seconds,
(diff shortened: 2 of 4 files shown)


ci/glci/model.py
```diff
@@ -480,3 +480,46 @@ def feature_by_name(feature_name: str):
         if feature.name == feature_name:
             return feature
     raise ValueError(feature_name)
+
+
+class OpenRC():
```
ccwienk, line 485: rather use a dataclass
(see e.g. glci.model.py for examples..)
or: https://docs.python.org/3/library/dataclasses.html for full reference

ci/glci/openstack_image.py
```diff
@@ -0,0 +1,70 @@
+import os
+import time
+
+from glci.model import OpenRC
+
+from openstack import connect
+
+class OpenstackImageUploader():
```
ccwienk, line 8: nit: paranthesis are not required for Python3

ci/glci/openstack_image.py
```diff
@@ -0,0 +1,70 @@
...
+  '''OpenstackImageUploader is a client to upload images to Openstack Glance.'''
+
+  def __init__(self, openrc: OpenRC):
+    self.conn = connect(
```
ccwienk, line 12: nit: I would prefer to not establish the connection in the c'tor, but rather do this when required (a common pattern to do this whilst avoiding repetitive code being to add a method creating the function, and memoising it, using the @functools.lru_cache decorator - there should also be examples in `ci/**.py`)

ci/glci/openstack_image.py
```diff
@@ -0,0 +1,70 @@
...
+    },
+  }
+  image_id = uploader.upload_image(image_name, image_path, image_meta)
+  uploader.wait_image_ready(image_id)
+  return image_id
+
+
+def main():
+  openrc = OpenRC()
+  image_name = os.environ['IMAGE_NAME']
```
ccwienk, line 64: please use argv to pass args instead
-> `argparse.ArgumentParser` - some examples can be found in `ci/*.py`

ci/glci/model.py
dkistner: Yep, sounds reasonable to me. Hmm I think for local development we should still keep the option in the constructor to read the values from the environment.

ci/glci/openstack_image.py
dkistner: +1

ci/glci/openstack_image.py
dkistner: Hmm I'm not sure if the Connection object really holds a persistent connection. I need to check.

ci/glci/openstack_image.py
dkistner: +1

ci/glci/model.py
```diff
@@ -480,3 +480,13 @@ def feature_by_name(feature_name: str):
         if feature.name == feature_name:
             return feature
     raise ValueError(feature_name)
+
+
+@dataclasses.dataclass(frozen=True)
+class OpenRC:
+    auth_url: str = os.environ['OS_AUTH_URL']
```
ccwienk, line 487: please do not consume values from ENVIRONMENT here.

If at all, this should be done at the pipeline-step script. However, I would prefer that configuration be passed from `cicd.yaml`

ci/glci/openstack_image.py
```diff
@@ -0,0 +1,75 @@
...
+      '''Upload an image from filesystem to Openstack Glance.'''
+
+      image = self.conn.image.create_image(
+          name=name,
+          filename=path,
+          disk_format='vmdk',
+          container_format='bare',
+          visibility='community',
+          meta=meta,
+          timeout=86400,
```
ccwienk, line 33: please state unit (e.g. timeout_millis, or timeout_seconds, ..)
+ why 86400?

ci/glci/openstack_image.py
```diff
@@ -0,0 +1,75 @@
...
+      )
+      return image['id']
+
+  def wait_image_ready(self, image_id: str):
+      '''Wait until an image get in ready state.'''
+
+      while True:
+          image = self.conn.image.get_image(image_id)
+          if image['status'] == 'queued' or image['status'] == 'saving':
+              time.sleep(10)
```
ccwienk, line 43: nit: add sleep-time as argument with default value?

ci/glci/openstack_image.py
```diff
@@ -0,0 +1,75 @@
...
+              continue
+          if image['status'] == 'active':
+              return
+          RuntimeError(f"image upload to Glance failed due to image status {image['status']}")
+
+
+def upload_image(image_name: str, image_path: str, build_number: str, openrc: OpenRC):
+    uploader = OpenstackImageUploader(openrc)
+    image_meta = {
+        'architecture': 'x86_64',
```
ccwienk, line 53: pass-in as argument (feel free to specify a default value)

ci/glci/model.py
dkistner: Ok, we do not default them via enviroment. How about read the env in the `main()` function? This will be called only for dev/testing purposes.

ci/glci/openstack_image.py
dkistner: One day. The default is one hour. As the images are pretty large I was not sure if this is always sufficient.

ci/glci/openstack_image.py
dkistner: yep will do it.

ci/glci/openstack_image.py
dkistner: That makes sense and give the script more flexibility.

ci/glci/openstack_image.py
```diff
@@ -0,0 +1,82 @@
...
+        },
+    }
+    image_id = uploader.upload_image(image_name, image_path, image_meta)
+    uploader.wait_image_ready(image_id)
+    return image_id
+
+
+def main():
+    openrc = OpenRC(
+        os.environ['OS_AUTH_URL'],
```
ccwienk, line 65: i'm still not too happy with passing those values via ENVIRONMENT - why not use argv?

dkistner: :D As mentioned `main()` is only called for local dev/testing purposes and its a common pattern when interacting with OpenStack clis/scripts to read the env information from environment. Therefore you usually source an `openrc`. file https://docs.openstack.org/mitaka/cli-reference/common/cli_set_environment_variables_using_openstack_rc.html

ci/glci/openstack_image.py
```diff
@@ -0,0 +1,82 @@
...
+          disk_format='vmdk',
+          container_format='bare',
+          visibility='community',
+          meta=meta,
+          timeout=timeout_seconds,
+      )
+      return image['id']
+
+  def wait_image_ready(self, image_id: str, wait_interval_seconds=10):
+      '''Wait until an image get in ready state.'''
```
ccwienk, line 38: also add a timeout?

ci/glci/openstack_image.py
ccwienk: ok - could you at least also add the same options to the `ArgumentParser`, making the ENV values the default, that can be overwritten via CLI (which is, I believe, also a common pattern)?

ci/glci/openstack_image.py
```diff
@@ -0,0 +1,82 @@
...
+          image = self.conn.image.get_image(image_id)
+          if image['status'] == 'queued' or image['status'] == 'saving':
+              time.sleep(wait_interval_seconds)
+              continue
+          if image['status'] == 'active':
+              return
+          RuntimeError(f"image upload to Glance failed due to image status {image['status']}")
+
+
+def upload_image(image_name: str, image_path: str, build_number: str, openrc: OpenRC, architecture='x86_64'):
```
ccwienk, line 50: use `glci.model.Architecture`; translate it to OpenStack's naming scheme. This has the advantage of implicitly validating allowed / known values, and avoid code duplication (with the exception of the translation code, of course)

ci/glci/openstack_image.py
dkistner: done

ci/glci/openstack_image.py
dkistner: obsolete.

ci/glci/openstack_image.py
dkistner: I removed the entire `main` function.

dkistner:
Updated the PR. For each OpenStack environment we need a `glci.model.CicdCfg` which contain internally an `glci.model.Openrc` object with access information and image properties. The image is now uploaded directly via web url (e.g. S3 object url) to Glance instead of uploading it from file system.
It's now also integrated with the publish process.

ci/glci/model.py
```diff
@@ -411,9 +418,26 @@ class AzurePublishCfg:
     storage_account_cfg_name: str
 
 
+@dataclasses.dataclass(frozen=True)
+class OpenRC:
```
ccwienk, line 422: what's an OpenRC?

ci/glci/openstack_image.py
```diff
@@ -0,0 +1,96 @@
+import dataclasses
+
+from datetime import datetime
+from time import sleep
+
+import glci
+
+from openstack import connect
```
ccwienk, line 8: nit: by convention, imports should be grouped, like so:

```
<imports from std python> # e.g. dataclasses, datertime, time, ..

<imports from other 3rdparty> # e.g. requests, openstack, ..

<imports from local> # e.g. glci
```
see https://www.python.org/dev/peps/pep-0008/#imports

ci/glci/model.py
dkistner: Hmm openrc are usually openstack enviroment variable files are called. Maybe we rename that e.g. to `OpenstackEnviroment`.

dkistner: I renamed it to `OpenstackEnviroment`

ci/glci/openstack_image.py
dkistner: Done. Connection is now also cached.

PR #97, opened 2020-06-17
Unrestrict dmesg from non-root users

### LINKED ISSUE #96: Allow non-root users to use dmesg

Remove the restriction on `dmesg` that makes it only usable for root.

**Why is this needed**:

1. Allowing non-root users to use `dmesg` makes it easier to identify various kernel-level errors, the most common of which (in Kubernetes containers) are situations in which the kernel OOM-killer cause processes to be killed.
1. Allows for easier diagnostics in case of driver and hardware issues.
1. Provides compatibility with CoreOS, which didn't restrict `dmesg` from non-root users.

Without this PR, identifying kernel-related issues would require pushing a privileged container to Kubernetes to run dmesg. If the issue is related to stability then it might not be possible to do after the fact, while with the PR any container that is still running on the host could be used to diagnose the issue.

liorokman:

Allows non-root users to use `dmesg`.

Fixes #96 

**Release note**:

```improvement user
Non-root users can now use dmesg.
```

### DIFF
--- a/features/server/file.include/etc/sysctl.d/allow-nonroot-dmesg.conf
+++ b/features/server/file.include/etc/sysctl.d/allow-nonroot-dmesg.conf
@@ -1,0 +1,1 @@
+kernel.dmesg_restrict = 0


marwinski:
lgtm but I'd prefer if this would go in after the branching of the 27 releases

gehoern:
this is a security problem, isnt it?

I mean in dmesg a possibly messages posted that are not needed to be visible for everyone. I think this is a security issue also for SGS

liorokman:
> this is a security problem, isnt it?

It's restricted because  the `dmesg` output may sometimes include addresses in the kernel address-space that can be useful to an attacker under the right circumstances. See https://lwn.net/Articles/414813/ for the details.

The problem with restricting access to the kernel buffer like this is that it becomes a lot harder to debug things like kernel-OOM and general driver issues (not unlike what we have experienced these last few weeks on AWS under CoreOS). Being able to view the kernel buffer has come in handy multiple times in the last few months while working on Gardener-managed clusters. 

It's also only pretend security. Unless GardenLinux is going to enable the kernel "lockdown" feature, the kernel isn't locked down by default and these values can be changed at any time by a sufficiently privileged user.

In the end, I think the value of disabling `dmesg` on the off-chance that something there could be used to attack the server is out-weighed by the difficulty imposed on the developers for gaining access to diagnostics. Most applications running on Kubernetes are going to be running in containers, and don't allow terminal access. It's mostly going to be developers that need access to this facility, and that will access it via `kubectl exec`.

PR #98, opened 2020-06-17
fix bug of alicloud image upload

jia-jerry:
fix bug of alicloud image upload
Fixes #

**Release note**:

```improvement operator

```


ci/glci/alicloud.py
```diff
@@ -31,13 +31,13 @@ def __str__(self):
 
     @staticmethod
     def to_availbel_str_array() -> []:
-        return {
+        return [
             str(AlicloudImageStatus.CREATING),
```
ccwienk, line 35: maybe you could instead just use:
```
return [v.value for v in AlicoundImageStatus]
```

PR #110, opened 2020-07-02
Contribute detailed nginx cache config back to gardenlinux-repo

AndreasBurger:
Includes the latest changes to the live nginx cache config, namely:
- 'OK' and 'moved permanently' responses (200 and 301) are valid for 30d
- 'moved temporarily' responses (302) are valid only for one hour
- any other response expires after ten minutes

### DIFF
--- a/hack/nginx.conf
+++ b/hack/nginx.conf
@@ -14,7 +14,9 @@ server {
             proxy_cache                 STATIC;
             proxy_cache_lock            on;
             proxy_cache_lock_timeout    30s;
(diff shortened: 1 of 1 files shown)


ccwienk:
@gehoern : apparently, w/o this cfg change, the proxy does not prevent repeated http-requests returning http-404, which recently happened (pbly due to a cfg change during build run, or some inconsistent repository state), which effectively again lead to our public IP address(es) being blocked for some time

PR #112, opened 2020-07-06
Enable swap memory accounting

### LINKED ISSUE #111: Enable swap accounting in the memory cgroup

Enable the swap accounting functionality in the memory cgroup controller.

**Why is this needed**:

Debian kernels are compiled by default with the `CONFIG_MEMCG_SWAP_ENABLED` disabled. The result is that the memory cgroup controller doesn't enforce limits on swap memory.

In cloud environments this functionality is required to allow configuring swap on the host but to also prevent any one container from allocating all of the swap memory.

liorokman:

Fixes #111 

**Release note**:

```improvement user

Swap memory accounting is enabled by default.

```


liorokman:
closed because the branch was based on `master` instead of `main`.

PR #113, opened 2020-07-06
Enable swap cgroup accounting

### LINKED ISSUE #111: Enable swap accounting in the memory cgroup

liorokman:

Fixes #111 

**Release note**:

```improvement user
Swap memory accounting is now enabled by default.
```

### DIFF
--- a/features/server/file.include/etc/default/grub.d/enable-swap-cgroup-accounting.cfg
+++ b/features/server/file.include/etc/default/grub.d/enable-swap-cgroup-accounting.cfg
@@ -1,0 +1,1 @@
+GRUB_CMDLINE_LINUX=" cgroup_enable=memory swapaccount=1"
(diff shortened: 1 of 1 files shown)


marwinski:
I generally think that this is a good idea but I am worried about higher resource consumption. This is ok if swap is used but not ok if swap is disabled which is probably accounts for more than 90% of our users. I did not find any information about this. Did you?

liorokman:
I don't know how much overhead this incurs if there is no swap enabled. I do know that Debian (and Debian derivatives) are the only distributions that disable swap accounting by default - so the overhead must be negligible enough for RedHat, Oracle, SuSE, and (RIP) CoreOS to leave it enabled by default. 

Unfortunately this can't be enabled after boot, only during boot.

marwinski:
This is good enough for me. I will test it (unfortunately my machine is broken) and submit. Thanks.

PR #114, opened 2020-07-06
Install open-iscsi by default

### LINKED ISSUE #109: Add open-iscsi to the base image

Add `open-iscsi` to the default installed packages.

**Why is this needed**:

This package is required to allow iSCSI volumes to be mounted on the node.

liorokman:

Fixes #109 

There's no easy way to install the `open-iscsi` package and have the daemon stay down. The `open-iscsi` package isn't integrated with systemd`s just-in-time socket activation functionality.

The overhead for this daemon seems to be small (around 10MiB) when it is just loaded.
```
root@localhost:/root# ps l 407 409
F   UID     PID    PPID PRI  NI    VSZ   RSS WCHAN  STAT TTY        TIME COMMAND
1     0     407       1  20   0   4904   284 -      Ss   ?          0:00 /sbin/iscsid
5     0     408       1  10 -10   5408  5020 -      S<Ls ?          0:00 /sbin/iscsid
```

**Release note**:

```improvement user
GardenLinux supports mounting iSCSI volumes. 

This enables using projects like [Longhorn](https://longhorn.io/) and [OpenEBS](https://openebs.io/) with GardenLinux as the underlying node operating system.

Run `systemctl enable iscsid && systemctl start iscsid` in a privileged container with access to the host filesystem on all hosts that should be capable of mounting iSCSI volumes. This could be done as a `Job` or as a `DaemonSet`.
```

### DIFF
--- a/features/server/exec.config
+++ b/features/server/exec.config
@@ -4,6 +4,8 @@ systemctl enable systemd-resolved
 systemctl enable ssh-keygen
 systemctl enable tmp.mount
 #systemctl enable ssh-moduli
+systemctl disable iscsid
+systemctl disable open-iscsi
 
 for i in $(ls /boot | grep vmlinuz | sed "s/vmlinuz-//"); do
 	systemctl enable kexec-load@$i
--- a/features/server/pkg.include
+++ b/features/server/pkg.include
@@ -27,3 +27,4 @@ less
 tcpdump
 vim-tiny
 quota
+open-iscsi


marwinski:
10MB sounds a bit much knowing that this is only used by a very small fraction of our user base. Can we have that disabled by default?

liorokman:
If we disable it by default, what would be the procedure for whoever needs it to enable it?

marwinski:
Deploy a privileged daemon set with host path with something like this

    chroot /hostPath systemctl enable (some)iscsi

(sorry forgot name). I can add the necessary change.

marwinski:
@gehoern: I guess you don't object if I move this to cloud?

PR #116, opened 2020-07-08
Add OpenStack image build upload procedure

dkistner:
Same as https://github.com/gardenlinux/gardenlinux/pull/91 to different traget branch `main`.

**Release note**:

```improvement operator
The OpenStack image upload procedure is now integrated.
```

### DIFF
--- a/ci/glci/model.py
+++ b/ci/glci/model.py
@@ -308,6 +308,13 @@ class AzurePublishedImage:
 
 
 @dataclasses.dataclass(frozen=True)
+class OpenstackPublishedImage:
+    region_name: str
+    image_id: str
+    image_name: str
+
+
+@dataclasses.dataclass(frozen=True)
 class ReleaseManifest(ReleaseIdentifier):
     '''
     metadata for a gardenlinux release variant that can be (or was) published to a persistency
@@ -320,6 +327,7 @@ class ReleaseManifest(ReleaseIdentifier):
        AwsPublishedImageSet,
        AzurePublishedImage,
        GcpPublishedImage,
+       OpenstackPublishedImage,
        None,
     ]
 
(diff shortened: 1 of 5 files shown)


ci/glci/model.py
```diff
@@ -316,7 +323,8 @@ class ReleaseManifest(ReleaseIdentifier):
     build_timestamp: str
     paths: typing.Tuple[typing.Union[S3_ReleaseFile]]
     published_image_metadata: typing.Union[AlicloudPublishedImageSet,
-                                           AwsPublishedImageSet, AzurePublishedImage, GcpPublishedImage, None]
+                                           AwsPublishedImageSet, AzurePublishedImage,
```
ccwienk, line 326: as discussed yesterday, I recommend formatting list member one entry per line and order alphabetically

ci/glci/model.py
```diff
@@ -465,9 +473,26 @@ class AzurePublishCfg:
...
+    region: str
+    project_name: str
+    username: str
+    password: str
+
+
+@dataclasses.dataclass(frozen=True)
+class OpenstackPublishCfg:
+    openrc: OpenstackEnviroment
+    properties: dict
```
ccwienk, line 489: while this technically works, of course, adding dataclasses also for the `properties` could help to improve readability / maintainability

ci/glci/model.py
msohn: you don't use an auto-formatter like black [1] ?

[1] https://github.com/psf/black

ci/glci/model.py
dkistner: Hmm the issue here we need the properties dict to pass image properties which can be different depending on the openstack environment. As everything within the ci comes from the secret server we should maybe also retrieve them from there. WDYT?

dkistner:
I have updated the PR. The openstack gl ci use now the secret server as input.
This PR require: https://github.com/gardener/cc-utils/pull/446
After this is merged I will remove the draft status.

ci/glci/model.py
dkistner: The information are now retrieved from secret server.

PR #117, opened 2020-07-09
Connect ci and Azure publishing procedure

dkistner:
The ci fetch now the configuration for the Azure publishing procedure from SecretServer correctly.

**Release note**:

```NONE

```

### DIFF
--- a/ci/cicd.yaml
+++ b/ci/cicd.yaml
@@ -18,6 +18,6 @@ cicd_cfgs:
         publisher_id: 'sap'
         plan_id: 'greatest'
         service_principal_cfg_name: 'shoot-operator-dev'
-        storage_account_cfg_name: 'gardelinux-dev'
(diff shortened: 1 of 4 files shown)


dkistner:
@ccwienk Can you please try out? I tested it locally and it fetch the configuration now correctly from the SecretServer, but please double check.

One TODO is still there. The configuration of the mail receipts which should be notified in case something went wrong during the publishing process.

ccwienk:
@dkistner : as for the email recipients, we might just add a list to cicd.yaml, using any of github-usernames, github-team-names, and email-addresses (the first two in particular to allow people to add themselves w/o publishing their email addresses, which not everybody may want - wdyt?

dkistner:
Hmm yes sound like a great idea to use github-team-names. I assume this will not work out of the box. Should we open an issue for that and bring it in with a subsequent pr?

PR #118, opened 2020-07-13
Refactor integrationtests pipeline

dguendisch:
Refactors the tekton part of the integration tests by making parameters like image name configurable and adding an example pipeline.

In addition:
- removes the pipenv dependency and uses only pip to install requirements.
- allows to have a different gcp image project than the project used to create the test VM.

### DIFF
--- a/ci/integrationtests/task.yaml
+++ b/ci/integrationtests/task.yaml
@@ -60,6 +60,7 @@ spec:
           service_account_json:
           service_account_json_path: /tmp/serviceaccount.json
           machine_type: n1-standard-1
+          image_project: sap-se-gcp-gardenlinux
           image_name: $(params.gcp_image_name)
           # user and ssh key used to connect to the vm
           user: admin
--- a/ci/integrationtests/task.yaml
+++ b/ci/integrationtests/task.yaml
@@ -7,6 +7,12 @@ spec:
   params:
   - name: iaas
     type: string
+  - name: ami_id
+    type: string
+    default: replaceme_aws
+  - name: gcp_image_name
+    type: string
+    default: replaceme_gcp
   resources:
     inputs:
     - name: repo
@@ -40,7 +46,7 @@ spec:
           access_key_id: %s
           secret_access_key: %s
           # which AMI to use for creating an ec2 instance
-          ami_id: ami-0e7e5773a975f0aea
+          ami_id: $(params.ami_id)
           ssh_key_filepath: /tmp/id_rsa
           # user and ssh key used to connect to the ec2 instance
           user: admin
@@ -54,7 +60,7 @@ spec:
           service_account_json:
           service_account_json_path: /tmp/serviceaccount.json
           machine_type: n1-standard-1
-          image_name: garden-linux-dev-gcp-38
+          image_name: $(params.gcp_image_name)
           # user and ssh key used to connect to the vm
           user: admin
           ssh_key_filepath: /tmp/id_rsa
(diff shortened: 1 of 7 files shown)


ci/integrationtests/task.yaml
```diff
@@ -71,10 +71,9 @@ spec:
 
         echo "> Installing dependencies"
         apk update && apk add make
-        pip install pipenv
```
msohn, line 67: - why do you want to get rid of pipenv ?
- where are test dependencies installed ?
- it looks like you want to switch from pipenv to pip only for automated test run, correct ?
I wouldn't like to do this for developing integration tests since to my experience installing python dependencies without using a virtual environment is not a good idea

ci/integrationtests/task.yaml
```diff
@@ -71,10 +71,9 @@ spec:
...
+        pip install scp
 
         cd $(resources.inputs.repo.path)/tests
-        pipenv install -dev
```
msohn, line 70: this command would update dependencies, better use pipenv sync, see https://pypi.org/project/pipenv/

ci/integrationtests/task.yaml
dguendisch: Yes, I would like to remove pipenv for automated testruns, why:
- I didn't manage to install pipenv in the cc/job-image image, neither in debian:bullseye
- the tests "only" need the `scp` module in addition to what is available already with the standard cc/job-image
- eventually we should have a docker image containing all required dependencies instead of adding them at runtime and thus unnecessarily prolonging each and every testrun.
- I don't mind/care what is being used for local development, be it a virtual env or be it a docker container, ideally it should be the same dep mgmt as in the testruns, but as said, I did not manage to install pipenv with the above mentioned images
- ideally you can provide us with an image (or dockerfile) where the tests can be executed in the way you prefer, i.e. with pipenv

ci/integrationtests/task.yaml
dguendisch: hmm, you are referring to a deleted line?

msohn: yep, still this was wrong

ci/integrationtests/task.yaml
msohn: which base image do you typically use for docker images ?

ci/integrationtests/task.yaml
dguendisch: It's only a copy of what you described in the resp. readme. Then I guess you'd might want to change it here: https://github.com/gardenlinux/gardenlinux/blame/main/tests/README.md#L15

ci/integrationtests/task.yaml
dguendisch: We regularly use `eu.gcr.io/gardener-project/cc/job-image`.

ci/integrationtests/task.yaml
msohn: mea culpa, I'll fix the readme

ci/integrationtests/task.yaml
msohn: where is the source of this image ?

dguendisch: I guess https://github.com/gardener/cc-utils/blob/master/Dockerfile, @ccwienk is this correct?

dguendisch: ping @msohn, how should we proceed? Merge this PR or wait for other suggestions or docker images from you?

PR #119, opened 2020-07-16
make image shared hiddenly on Alicloud

jia-jerry:
Share gardenlinux image to other accounts
Fixes #

**Release note**:

```improvement operator
GardenLinux images are built for all regions on Alicloud and are made shared with other accounts.
```

### DIFF
--- a/ci/glci/alicloud.py
+++ b/ci/glci/alicloud.py
@@ -10,6 +10,7 @@ import dataclasses
 import glci.model
 from aliyunsdkecs.request.v20140526 import ImportImageRequest
 from aliyunsdkecs.request.v20140526 import DescribeImagesRequest
+from aliyunsdkecs.request.v20140526 import DeleteImageRequest
 from aliyunsdkecs.request.v20140526 import DescribeRegionsRequest
 from aliyunsdkecs.request.v20140526 import CopyImageRequest
 from aliyunsdkecs.request.v20140526 import ModifyImageSharePermissionRequest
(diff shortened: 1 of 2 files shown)


ci/glci/alicloud.py
```diff
@@ -113,6 +115,22 @@ def make_image(self) -> glci.model.OnlineReleaseManifest:
         return dataclasses.replace(
             self.release, published_image_metadata=published_image_set)
 
+    # Share image in a hidden way. The account should apply for whitelist
+    def _share_image(self, region_image_map: {}):
```
ccwienk, line 119: nit: the usual type annotion would be `region_image_map: dict`

ci/glci/alicloud.py
```diff
@@ -113,6 +115,22 @@ def make_image(self) -> glci.model.OnlineReleaseManifest:
...
+        for region, image_id in region_image_map.items():
+            self.acs_client.set_region_id(region)
+            logger.info(
+                f"share image ({region}/{image_id}) as a hidden image"
+            )
+            req = ModifyImageSharePermissionRequest.ModifyImageSharePermissionRequest()
+            req.set_ImageId(image_id)
+            # HIDDEN --> share
+            # PRIVATE --> un-share
+            req.set_LaunchPermission('HIDDEN')
```
ccwienk, line 129: could you:
1. extract the permission values e.g. as an enum
2. pass-in the visibility into this method as a parameter (it is of course OK to assign `HIDDEN` as default value)
?

jia-jerry: Regarding _deletion_, is any reference function? I need input and output.

ccwienk: sure, there is an example for AWS:

https://github.com/gardenlinux/gardenlinux/blob/main/ci/glci/aws.py#L257

Note that this is not yet _used_ (I used the function manually via a REPL shell some times for purging images I published for testing purposes)

PR #123, opened 2020-08-17
Gardener feature: add crictl and disabling containerd and docker

danielfoehrKn:

Installs `crictl` as part of the Gardener feature.  Please see [here](https://github.com/gardener/gardener/issues/2719) for more information regarding the Gardener context.

Also fixes the disabling of containerd and docker systemd services. This would not have worked previously anyways, as Gardener currently always requires docker to be present - even when bootstrapping a containerd only node.

With this [PR](https://github.com/gardener/gardener/pull/2739), the Gardener will 
 - not require Docker any more to bootstrap CRI enabled nodes
- require `crictl` for the bootstrap of CRI enabled nodes

Fixes #

Open as draft PR as the corresponding [PR in the Gardener repository](https://github.com/gardener/gardener/pull/2739) is not merged yet.

**Release note**:

```improvement operator
Adding crictl to the Gardener feature. 
```
```improvement operator
Fixing a bug  that lead to not disabling the docker systemd service properly.
```

### DIFF
--- a/features/gardener/exec.config
+++ b/features/gardener/exec.config
@@ -1,18 +1,22 @@
 #!/bin/bash
 
 # set the default to iptalbes and deactivate nf_tables kernel modules
-#
 update-alternatives --set iptables /usr/sbin/iptables-legacy
 update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 update-alternatives --set arptables /usr/sbin/arptables-legacy
 update-alternatives --set ebtables /usr/sbin/ebtables-legacy
 
 # just to make sure there are no traces left
-#
 systemctl mask nftables.service
 
 # Disable docker and containerd, Gardener will have to enable the
 # one it uses
-#
-systemctl disable docker
-systemctl disable containerd
+systemctl disable docker.service
+systemctl disable docker.socket
+systemctl disable containerd.service
+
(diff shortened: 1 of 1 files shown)


marwinski:
I do have two questions:

* the new iptables feature has been deactivated on request of @vpnachev. I don't see this change related to it. Can you please check with him as it might have an impact on existing clusters. 
* docker- and containerd services appear to be deactivated twice, once in the build script and once in the 01-default-preset.preset file. Is this really necessary?

danielfoehrKn:
>The new iptables feature has been deactivated on request of @vpnachev. I don't see this change related to it. Can you please check with him as it might have an impact on existing clusters.

There are no changes in this PR regarding iptables?

> docker- and containerd services appear to be deactivated twice, once in the build script and once in the 01-default-preset.preset file. Is this really necessary?

I am not 100 % sure if both are required due to issues building the image - I ll follow up on it before I remove the draft.

marwinski:
As for my first point: sorry I was still asleep

danielfoehrKn:
Tested it out, and yes you are right, only `systemctl disable docker.socket` was missing. 
Removed the preset files.

danielfoehrKn:
Closing due to PR in gardener closed - please see [here](https://github.com/gardener/gardener/pull/2739#issuecomment-677667949).

PR #125, opened 2020-09-15
clean alicloud image based on Image name

jia-jerry:
Provide remove images method for AliCloud
Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/ci/glci/alicloud.py
+++ b/ci/glci/alicloud.py
@@ -54,12 +54,10 @@ class AlicloudImageMaker:
         self.acs_client = acs_client
         self.release = release
         self.build_cfg = build_cfg
-        if build_cfg is not None:
-            self.bucket_name = build_cfg.oss_bucket_name
-            self.region = build_cfg.alicloud_region
-        if release is not None:
-            self.image_oss_key = f"gardenlinux-{self.release.version}.qcow2"
-            self.image_name = f"gardenlinux-{self.release.canonical_release_manifest_key_suffix()}"
+        self.bucket_name = build_cfg.oss_bucket_name
+        self.region = build_cfg.alicloud_region
+        self.image_oss_key = f"gardenlinux-{self.release.version}.qcow2"
+        self.image_name = f"gardenlinux-{self.release.canonical_release_manifest_key_suffix()}"
 
     # copy image from S3 to OSS
     def cp_image_from_s3(self, s3_client):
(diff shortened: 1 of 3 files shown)


ci/glci/alicloud.py
```diff
@@ -273,3 +283,24 @@ def copy_image(self, src_image_id: str, dest_region: str) -> str:
 
 def parse_response(response):
     return json.loads(response)
+
+def main():
```
ccwienk, line 287: I suppose this main function is debugging/testing code? not sure whether this should stay in

ccwienk (commented):
lgtm - you might want to rm the debugging / testing code?

ci/glci/alicloud.py
```diff
@@ -54,10 +54,12 @@ def __init__(
         self.acs_client = acs_client
         self.release = release
         self.build_cfg = build_cfg
-        self.image_oss_key = f"gardenlinux-{self.release.version}.qcow2"
-        self.region = build_cfg.alicloud_region
-        self.bucket_name = build_cfg.oss_bucket_name
-        self.image_name = f"gardenlinux-{self.release.canonical_release_manifest_key_suffix()}"
+        if build_cfg is not None:
+            self.bucket_name = build_cfg.oss_bucket_name
```
ccwienk, line 58: I think it would be better to either require build_cfg not to be None, or to initialise the fields with None in that case - otherweise, code using this object will have to guard against the absense of the field

ci/glci/alicloud.py
```diff
@@ -134,22 +136,30 @@ def _share_images(self, region_image_map: dict):
...
 
-            req = DeleteImageRequest.DeleteImageRequest()
-            req.set_ImageId(image_id)
-            self.acs_client.do_action_with_exception(req)
-
-        self.acs_client.set_region_id(self.region)
+        regions = self._list_regions()
+        regions.append(self.region)
+        for region in regions:
+            exist, image_id = self._check_image_existance(region,
```
ccwienk, line 144: I think it would be a good idea to extract the functionality to remove a single image into an own method

ci/glci/alicloud.py
```diff
@@ -134,22 +136,30 @@ def _share_images(self, region_image_map: dict):
...
+                self.acs_client.set_region_id(region)
+                req = ModifyImageSharePermissionRequest.ModifyImageSharePermissionRequest()
+                req.set_ImageId(image_id)
+                req.set_LaunchPermission(str(ImageShareOption.UNSHARE))
+                self.acs_client.do_action_with_exception(req)
+                req = DeleteImageRequest.DeleteImageRequest()
+                req.set_ImageId(image_id)
+                self.acs_client.do_action_with_exception(req)
+            except:
+                logger.warning(f"Fail to delete image {self.image_name}/{image_id} in region {region}")
```
ccwienk, line 160: I would somewhat prefer to raise in this case

maybe like so:

```
def delete_image(self, region, image_id):
  # code to delete one image....

def delete_images(self, keep_going=True):
  did_raise = False
  for region, image_id in existing_images:
    try:
      self.delete_image(region, image_id)
    except:
      if keep_going:
        did_raise = True
        continue
      raise
 if did_raise:
   raise
```

ccwienk (commented):
some more comments

ccwienk (approved):
lgtm - thank you!

PR #135, opened 2020-09-25
rm python2 from azure build

AndreasBurger:
After cursory testing The Microsoft Azure Linux-Agent does not seem to depend on python2 anymore. My current best guess for the required python version is 3.8+, as given by https://github.com/Azure/WALinuxAgent/blob/develop/requirements.txt

### DIFF
--- a/features/azure/pkg.include
+++ b/features/azure/pkg.include
@@ -1,7 +1,5 @@
 # Azure images require waagent 
 waagent
-# Some waagent extensions still need python2.7, and break when it is not present
-python
 # debootstrap does not pull python3-cffi-backend
 #python3-cffi-backend
 #


vpnachev:
I will take care next week to verify that the image can be used by [Gardener](https://gardener.cloud) with python3 only.

> My current best guess for the required python version is 3.8+, as given by https://github.com/Azure/WALinuxAgent/blob/develop/requirements.txt

I was thinking so, but this is conditional requirement for the `distro` package , see https://github.com/Azure/WALinuxAgent/issues/1939.

marwinski (commented):
The build fails with this:

```
executing from azure: exec.config
sed: can't read /etc/waagent.conf: No such file or directory
```

This said, it is probably unrelated to this change as waagent gets removed during the build process. We've just had this in a different place as well due to a python 3.8 upgrade from Debian.

vpnachev:
This PR needed rebase to get the latest fixes from today. Could you try to build again from https://github.com/gardenlinux/gardenlinux/pull/135/commits/80d3865c4b4f53c35c5b51e4d02f93f3c155c825 ?

vpnachev (approved):
/lgtm

I have successfully created a Gardener shoot cluster with image build based on this PR.

PR #137, opened 2020-09-26
Pin docker.io version to 19.03.13+dfsg1-2

vpnachev:
This is needed as a workaround for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=970525 till the fix is delivered to the stream used by Garden Linux.

Fixes #

**Release note**:

```improvement operator

```


marwinski:
Thanks Vladimir!

PR #161, opened 2020-10-31
Set DEBIAN_FRONTEND to noninteractive in Debian docker images

berendt:
Signed-off-by: Christian Berendt <berendt@23technologies.cloud>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ RUN	go get -u golang.org/x/lint/golint \
      && make
 
 FROM 	debian:testing-slim
+ENV     DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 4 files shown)


lightmare:
@gehoern Sorry for jumping into a closed PR, I've accidentally got here when looking for something completely different.
[Docker reference](https://docs.docker.com/engine/reference/builder/#env) advises to use `ARG` instead of `ENV` for this purpose, or alternatively to set the variable only in the RUN instruction.

berendt:
@lightmare Thanks for the tip. Would you like to open a followup so we can correct it?

lightmare:
Sure, I can make a PR.

PR #165, opened 2020-11-03
Use "Garden Linux"

berendt:
Little things. But I think it is important to use the right name of the product everywhere.


gehoern:
Nice. At least someone is watching the brand :-D.
Thanks

PR #174, opened 2020-11-13
Fixed font issue with svg logos

leonardlorenz:
On some devices the original font used in the logos is not supported. This patch converts the text into path data which is font independent.


vpnachev (approved):
/lgtm

PR #175, opened 2020-11-13
for advanced Docker Proxy.

Bobrow:
I’m working with advanced Docker Proxy. 
This requires reconfiguration of Docker on every Node (/etc/systemd/system/docker.service.d/http-proxy.conf),
daemon-reload, and restart docker.service.
Restart – during bootstrapping - crashes already running container on the Node.
To prevent such crash, /etc/docker/daemon.json should have setting “live-restore: true”

### DIFF
--- a/features/gardener/file.include/etc/docker/daemon.json
+++ b/features/gardener/file.include/etc/docker/daemon.json
@@ -1,4 +1,5 @@
-{ "live-restore": true, "storage-driver": "overlay2",
+{ "live-restore": true,
+  "storage-driver": "overlay2",
   "default-ulimits": {
       "memlock": {
           "name": "memlock",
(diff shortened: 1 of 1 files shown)


features/gardener/file.include/etc/docker/daemon.json
```diff
@@ -1,4 +1,4 @@
-{ "storage-driver": "overlay2",
+{ "live-restore": true, "storage-driver": "overlay2",
```
berendt, line 1: I would prefer one argument per line.

marwinski:
@rfranzke: I consider this PR useful but it changes behaviour. Any objections from your side (https://docs.docker.com/config/containers/live-restore/)?

rfranzke:
No

PR #176, opened 2020-11-16
Renamed "debuerreotype-" files to "garden-" files.

### LINKED ISSUE #93: remove the nameing for debuerreotype
Sone names in the bin folder are starting with the prefix "debuerreotype" since the project https://github.com/debuerreotype/debuerreotype is our parent. we changed so much in the recent past that it is time to get rid of the unspellable name brand everything to garden but also making sure we state the debuerreotype guys for the initial project as well so they get all their credit.

mtrx1337:
It renames the debuerreotype prefixed files to use the garden prefix instead.
For example: debuerreotype-chroot -> garden-chroot
The prefix is needed to keep compatibility with the relative paths used, that's why i kept it instead of removing it completely.

This is to easier type out the file name and because we added a bunch of code to them.

Fixes #93 

It built on kvm and I was able to start the VM. Does this mean everything worked fine? Is there any further testing needed?


gehoern (commented):
nice

PR #215, opened 2021-03-23
avoid setting ENV DEBIAN_FRONTEND in Dockerfiles

lightmare:

/kind cleanup
/area os
/os garden-linux
/priority normal

Use ARG DEBIAN_FRONTEND=noninteractive, which defines a build-time variable.
ENV on the other hand preserves that variable in the final image. For an explanation of
why that is undesirable, see:

https://docs.docker.com/engine/reference/builder/#env
https://github.com/moby/moby/blob/docs/docs/faq.md#why-is-debian_frontendnoninteractive-discouraged-in-dockerfiles

Improves upon #161 

**Release note**:

```other operator
NONE
```


gehoern:
interesting point ... i guess that means we also nee to change some follow on processes. PR will be merged but we need to change some more elements. Thanks!

PR #248, opened 2021-04-26
Cleanup pipeline rendering and enable jobs without secrets-server

AndreasBurger:
This PR tries to improve the usability of the existing pipeline. The contained changes are:
- Lint/cleanup the existing pipeline-gear
- Un-hardcode the pipeline's deployment namespace
- Enables the pipeline to run without a deployed secrets-server
- Add flags to create a minified pipeline, i.e. a pipeline that only builds the image.


ci/render_task.py
```diff
@@ -17,49 +17,58 @@
 
 def main():
     parser = argparse.ArgumentParser()
+    parser.add_argument('--use-secrets-server', action='store_true')
     parser.add_argument('--outfile', default='tasks.yaml')
+    parser.add_argument('--minimal', action='store_true', help='omit prebuild and promote steps')
```
ccwienk, line 22: I'm not sure about the "minimal" build semantics tbh. there was the idea of adding "build targets", and to specify what the build does by adding or omitting those. I would prefer to stick to this approach (adding a flag on CLI-level as a convenience-shorthand seems okay-ish. but it should not be passed into the pipeline, i think

ci/render_task.py
```diff
@@ -17,49 +17,58 @@
...
 
     parsed = parser.parse_args()
+    full_build = not parsed.minimal
```
ccwienk, line 25: :-/

as stated above. this will not be very sustainable. it is very imminent that we will have more than two different "build flavours"

ci/tasks.py
```diff
@@ -76,6 +80,7 @@ def promote_task(
 
 
 def build_task(
+    use_secrets_server: bool,
```
ccwienk, line 83: I would prefer to pass the secrets-file here. in general, it was to my experience almost never a good idea to use booleans for this kind of parametrisations

ccwienk (changes requested):

jensh007 (approved):

PR #278, opened 2021-07-05
Error notifications

jensh007:
"/kind" identifiers:     enhancement
/area os
/os garden-linux

This PR adds the capability to send email notfifications from the build pipeline if the build breaks.

**Release note**:
```
NONE
```


ci/render_pipeline_run.py
```diff
@@ -93,6 +93,7 @@ def mk_pipeline_packages_run(
     branch: str,
     cicd_cfg: str,
     committish: str,
+    disable_notification: str,
```
AndreasBurger, line 96:
```suggestion
    disable_notifications: str,
```
Up to you.

ci/render_pipeline_run.py
```diff
@@ -295,6 +305,7 @@ def main():
     parser.add_argument('--git-url', default='https://github.com/gardenlinux/gardenlinux.git')
     parser.add_argument('--flavour-set', default='all')
     parser.add_argument('--version', default=None)
+    parser.add_argument('--disable-notification', default=False)
```
AndreasBurger, line 308:
```suggestion
    parser.add_argument('--disable-notification', action='store_const', const='True', default='False')
```
avoids the cast, but overall more typing. Up to you

ci/render_pipelines.py
```diff
@@ -221,7 +265,7 @@ def mk_pipeline(
         build_task = mk_pipeline_build_task(
             gardenlinux_flavour=glf,
             pipeline_flavour=pipeline_flavour,
-            run_after= [ base_build_task.name ],
+            run_after=[base_build_task.name],
```
AndreasBurger, line 268: +1

ci/steps.py
```diff
@@ -697,3 +697,56 @@ def build_base_image_step(
...
+                git_url,
+                status_dict_str,
+            ],
+        ),
+        volumeMounts=volume_mounts,
+        env=env_vars,
+    )
+
+
+def getlog_step(
```
AndreasBurger, line 729:
```suggestion
def get_logs_step(
```
singular/plural is up to you, but I'd rather have the underscore

ci/steps.py
```diff
@@ -697,3 +697,56 @@ def build_base_image_step(
...
+    pipeline_run: tkn.model.NamedParam,
+    namespace: tkn.model.NamedParam,
+    env_vars: typing.List[typing.Dict] = [],
+    volume_mounts: typing.List[typing.Dict] = [],
+):
+    return tkn.model.TaskStep(
+        name='get-logs',
+        image=DEFAULT_IMAGE,
+        script=task_step_script(
+            path=os.path.join(steps_dir, 'getlogs.py'),
```
AndreasBurger, line 740:
```suggestion
            path=os.path.join(steps_dir, 'get_logs.py'),
```
also, I'd like to be consistent with `get_log` and `get_logs`, whatever it'll end up being

ci/steps/getlogs.py
```diff
@@ -0,0 +1,88 @@
...
+from kubernetes import client, config
+
+
+def getlogs(repo_dir: str, namespace: str, pipeline_run_name: str):
+    @dataclass
+    class TaskRunInfo:
+        name: str
+        pod_name: str
+        steps: Dict[str, str]  # step-name --> step container
+
```
AndreasBurger, line 18: consider moving this to the model

AndreasBurger: also, an "anonymous" class is a bit unusual. Consider moving it out of the function

ci/steps/getlogs.py
```diff
@@ -0,0 +1,88 @@
...
+    # pod_dict = {tr['pipelineTaskName']: tr['status']['podName'] for tr in task_runs.values()}
+
+    task_run_infos = []
+    for tr in task_runs.values():
+        task_steps = {step['name']: step['container'] for step in tr['status']['steps']}
+        task_run_infos.append(
+            TaskRunInfo(
+                name=tr['pipelineTaskName'],
+                pod_name=tr['status']['podName'],
+                steps=task_steps
```
AndreasBurger, line 51:
```suggestion
                steps=task_steps,
```
nit

ci/steps/getlogs.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # create a zip file to store logs:
+    zip_name = os.path.join(repo_dir, 'build_log.zip')
+    with zipfile.ZipFile(zip_name, mode='w') as log_zip:
+        for run_info in task_run_infos:
+            pod_name = run_info.pod_name
+            counter = 0
+            for step, container_name in run_info.steps.items():
+                zip_comp = f'{run_info.name}/{counter:02}_{step}.log'
+                print(f'Getting logs for pod {pod_name}, step {step} in container {container_name}')
```
AndreasBurger, line 63:
```suggestion
                print(f'Getting logs for {pod_name=} {step=} {container_name=}')
```
While I personally do not prefer this syntax, it makes a bit more sense this sort for debug-logging. If this is intended as a nice message for the end-user, disregard this comment

ci/steps/notify.py
```diff
@@ -0,0 +1,136 @@
...
+from email import encoders
+from email.mime.base import MIMEBase
+import glci.notify
+import glci.util
+import json
+import mailutil
+import os
+from string import Template
+import typing
+import urllib
```
AndreasBurger, line 13: nit: sorting (I think its done here as we do it in cc-utils so far)

ci/steps/notify.py
```diff
@@ -0,0 +1,136 @@
...
+):
+    namespace = '$(params.namespace)'
+    pipeline_name = '$(params.pipeline_name)'
+    pipeline_run = '$(params.pipeline_run_name)'
+
+    repo_dir = '$(params.repo_dir)'
+
+    if distutils.util.strtobool('$(params.disable_notification)'):
+        print('Notification is disabled, bot sending email')
+        return
```
AndreasBurger, line 44: These should be passed using the existing Framework

AndreasBurger (changes requested):
generally lgtm

ci/steps/getlogs.py
jensh007: I prefer this one to keep as it is. This dataclass is convenient but exactly used only in this function. Why make it global?

ci/steps/getlogs.py
jensh007: agreed.

ci/steps/getlogs.py
jensh007: hm may be shorter but not more readable.

ci/steps/notify.py
jensh007: sorting is good but where is the order incorrect here?

ci/steps/notify.py
jensh007: not sure about this comment: it uses glci.notify....

ci/render_pipeline_run.py
jensh007: agreed.

ci/render_pipeline_run.py
jensh007: agreed.

ci/steps.py
jensh007: agreed.

ci/steps.py
jensh007: agreed.

PR #299, opened 2021-08-05
Annotate package patches

mbanck:

/kind doc
/area os
/os garden-linux

Adds some comments to package patches.


mbanck:
> @mbanck Label kind/doc does not exist.

well, I think it should!

Issue #253, opened 2021-05-27
GCP build is failing due to missing google-compute-engine package


vpnachev:
**What happened**:
I tried to build GCP image from the main branch (bf4bb69c61f81189a9388c8a23b94591710b9cec) but it is failing with
```bash
$ make gcp 
...
debootstrap --force-check-gpg --variant=minbase --no-merged-usr --keyring=/tmp/tmp.TULKPc8rus/debian-archive-bullseye-keyring.gpg --arch=amd64 --include=auditd,ca-certificates,dnsutils,dracut,iproute2,iptables,iputils-ping,libnss-resolve,libnss-systemd,libpam-systemd,lsb-release,net-
tools,netbase,openssh-server,policykit-1,procps,sudo,sysstat,systemd,systemd-coredump,systemd-cron,traceroute,wget,less,tcpdump,vim-tiny,quota,netcat,fdisk,grub-cloud-amd64,cloud-guest-utils,libpam-cracklib,apparmor,containerd,containernetworking-plugins,docker.io,ethtool,ipvsadm,soc
at,curl,logrotate,arptables,ebtables,xfsprogs,btrfs-progs,jq,ipvsadm,cifs-utils,nfs-common,google-compute-engine,rng-tools-debian --exclude=gcc-9-base,irqbalance bullseye rootfs http://45.86.152.1/gardenlinux/archive/debian/20210527T000000Z
I: Retrieving InRelease
I: Checking Release signature
I: Valid Release signature (key id 0146DC6D4A0B2914BDED34DB648ACFD622F3D138)
I: Retrieving Packages
I: Validating Packages
I: Resolving dependencies of required packages...
I: Resolving dependencies of base packages...
I: Checking component main on http://45.86.152.1/gardenlinux/archive/debian/20210527T000000Z...
E: Couldn't find these debs: google-compute-engine

error: 'debootstrap' failed!

  Full command:

    debootstrap --force-check-gpg --variant=minbase --no-merged-usr --keyring=/tmp/tmp.TULKPc8rus/debian-archive-bullseye-keyring.gpg --arch=amd64 --include=auditd\,ca-certificates\,dnsutils\,dracut\,iproute2\,iptables\,iputils-ping\,libnss-resolve\,libnss-systemd\,libpam-systemd\,ls
[...]

PR #309, opened 2021-08-18
Rebuild Ubuntu source package for google-guest-agent

Natureshadow:
/kind enhancement
/area packages
/os garden-linux

This change uses the Ubuntu source package for google-guest-agent instead of the upstream sources. Upstream sources pull in dependencies from GitHub and need network access for package build. Ubuntu has a working package with the Golang dependencies vendored into the `debian/` directory, which is not nice, but nicer.

I have not tested a full rebuild yet, only the one package script alone.

### DIFF
--- a/packages/manual/google-guest-agent
+++ b/packages/manual/google-guest-agent
@@ -3,23 +3,20 @@ set -euo pipefail
 
 . $(dirname $0)/.helper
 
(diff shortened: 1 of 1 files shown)


Natureshadow:
> I assume this
> 
> fixes #253

I don't think so. This builds `google-guest-agent`, not `google-compute-engine`.

PR #313, opened 2021-08-19
use cloud-init for VM provisioning on Azure

MSSedusch:

/kind enhancement
/area os
/os garden-linux

The PR switches from using the waagent for VM provisioning to cloud-init. This makes the waagent compatible with Pyhton 3.9

Fixes #

**Release note**:

```feature user
* switch to cloud-init for VM provisioning on Azure
```

### DIFF
--- a/features/azure/pkg.include
+++ b/features/azure/pkg.include
@@ -1,6 +1,10 @@
 # Azure images require waagent
-#waagent
+waagent
 # debootstrap does not pull python3-cffi-backend
-#python3-cffi-backend
-#
-#cloud-init
+python3-cffi-backend
+python3-cryptography
+python3-oauthlib
+cloud-init
+cloud-guest-utils
+hyperv-daemons
+isc-dhcp-client
--- a/features/azure/exec.config
+++ b/features/azure/exec.config
@@ -1,7 +1,3 @@
 #!/usr/bin/env bash
 
-wget --quiet -P /tmp http://45.86.152.1/gardenlinux/pool/main/w/waagent_2.2.47-2.2_all.deb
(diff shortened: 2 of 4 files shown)


vpnachev:
Thanks @MSSedusch!
Just one question - is it still possible to use bash script as user data to configure the VMs?

MSSedusch:
@vpnachev yes. I tested the image with Gardener and the shoot came up. I was also able to scale the shoot to 2 nodes.

MrBatschner:
@MSSedusch Could you quickly elaborate how changing from waagent to cloud-init makes waagent compatible with Python 3.9? That part of the commit-message caused some head scratching lately...

features/azure/pkg.include
```diff
@@ -1,6 +1,10 @@
...
-#python3-cffi-backend
-#
-#cloud-init
+python3-cffi-backend
+python3-cryptography
+python3-oauthlib
+cloud-init
+cloud-guest-utils
+hyperv-daemons
+isc-dhcp-client
```
waldiTM, line 10: Well. Using ìsc-dhcp-client` will not work, as this OS uses `systemd-networkd`. Also it is not longer supported.

waldiTM:
We talked and consider another way forward and just use `cloud-init` in the future. `waagent` is restricted to extension usage, which are not really appropriate for immutable systems.

MSSedusch:
should I close this PR or should I change it to only use cloud-init?

marwinski (changes requested):
As discussed we need to agree on a way forward. Ideally we want either cloud-init or waagent, but not both.

PR #347, opened 2021-09-07
Add support for local credentials to example integration-test-framework

AndreasBurger:
/kind test
/area os
/os garden-linux

Adapts the existing exemplary integration-test framework so that it _optionally_ supports authentication using local credentials in a way that is transparent to the test-author. Also adds small readme.mds to document usage.
For this PR this was only done for GCP and Azure, as there were some uncertainties with handling this using pytest. These changes should provide an example for doing similar work on the other cloud provider integration tests.

It might be worth thinking about merging this into your dev-branches instead of merging into main. Either way, rebasing should not be much of an issue.

### DIFF
--- a/integration_tests/azure/conftest.py
+++ b/integration_tests/azure/conftest.py
@@ -1,7 +1,10 @@
 from dataclasses import dataclass
 import pytest
 
-from azure.identity import ClientSecretCredential
+from azure.identity import (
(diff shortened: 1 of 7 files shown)


MrBatschner (approved):
Cool stuff and extremely helpful.
/lgtm

PR #350, opened 2021-09-08
Install extra packages via apt

waldiTM:
Installation of additional packages via debootstrap may or may not work.
The included dependency resolver is limited.  Instead just use apt-get
to do the heavy lifting.

/kind enhancement
/area os
/os garden-linux

### DIFF
--- a/bin/garden-init
+++ b/bin/garden-init
@@ -118,7 +118,6 @@ done
 include_aptversion=
 include_aptfile=
 include_wget=
-include_bootstrap=
 for i in $(filter_comment <<< $include | filter_variables | filter_if); do
     if [[ $i = *.deb ]]; then
 	if [[ $i = *://* ]]; then	include_wget+="$i "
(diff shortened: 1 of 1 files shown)


waldiTM:
There are three changes in the package list:

* `openssl` and `libssl1.1` are pulled from security, as long as Bullseye is used as base.
* No `cron` anymore.

gehoern:
this patch introduces several problems:

1. it does not care about "forced" uninstalls in pkg.exclude (before this was done over --include or --exclude of debbootstap)
2. it does not care about the feature of debootstrap to e.g. uninstall or overwrite "essential" packages. in a running debian system this is not possible but it is a feature of debootstrap
3. the apt-mark feature is patched out ... so all essential packages are marked as explicit requested the clutters completely the list of installed elements (non of this was requested and is handled undercover by apt, without bothering the customer to think about: why he installed ever e.g. libpam0 and not libpam17 what maybe relevant at some later time)

So the concluding question is (even the patch was already accepted): the described problems cause permanent friction an also parts of it got already otherwise implemented. 

- Is there any evidence for this debootstrap problem?
- how would we tackle the "3rd" person installing e.g. an image without apt / dpkg (possible in debootstap, impossible with inline apt)
- can we have an effective impact discussion on this debootstrap issue? is this anywhen about to be solved (i cannot find a mark on the todo/changelog list)

waldiTM:
1. Okay, this is missing.
2. Does any pkg.exclude try to exclude base packages? I see `systemd-timesyncd` (not base) and `irqbalance` (not longer pulled in), but nothing else.
3. If the apt-mark stuff is so essential, maybe this should be in debootstrap?
4. Yes. With the repo switch the dependency solver was not longer able to install everything. Sadly I did not make a copy, but that's why I did the change. It works currently, because you use dpkg to force in own packages.
5. The dependencies in Debian do not allow for dpkg to be removed. apt, yes.
6. Which debootstrap issue?

waldiTM:
1. Hu? Where is this broken? The logic filters all to be excluded packages from the installation list already. Can you please show what you see?
@gehoern

PR #362, opened 2021-09-15
Fix NTP server for GCP

marwinski:

/kind bug
/area os
/os garden-linux

On GCP the wrong file was edited for the correct NTP server. This patch fixes it.

Fixes #

**Release note**:

```feature user
Garden Linux on GCP now uses the Google NTP server.
```


waldiTM:
Any reason why this is not a separate file in `/etc/systemd/timesyncd.conf.d/` instead of modifying existing files?

marwinski:
No there is no particular reason. Does this work?

PR #364, opened 2021-09-16
fix: delete all container images on make clean

MrBatschner:

/kind cleanup
/area os
/os garden-linux

Make clean would delete all containers coming from gardenlinux/build-image
and the image itself, all other container images however would remain.
By calling the clean target of the Makefile within the docker directory,
all container images that were built during a Garden Linux build will be
removed.

Building a Garden Linux release would build both, the build-image and
the integration test image (and others). With this change to the Make
file, only the build container image is created, speeding up the process
significantly. The integration test container image is only built on
demand.

**Release note**:

```feature user
fix: delete all container images on make clean
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,19 +17,23 @@ cert/sign.pub:
 docker:
 	make --directory=docker all
 
+.PHONY: docker-build
+docker-build:
+	make --directory=docker build-image
+
 all_prod: ali aws gcp azure metal openstack vmware kvm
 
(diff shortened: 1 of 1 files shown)


mbanck:
I'm wondering whether this should be a more drastic make target, like `make distclean` or something?

Not exactly sure how often `make clean` is called, but it might be annoying having to redo the build containers each time.

MrBatschner:
`make distclean` it is.

Also another change to speed up build times significantly.

PR #367, opened 2021-09-17
Update iproute2 and python3

5kt:

/kind enhancement
/area os
/os garden-linux

Update python3.9 and iprounte2 packages used.


marwinski (approved):
much appreciated :-)

PR #368, opened 2021-09-17
Add security=selinux kernel parameter

### LINKED ISSUE #333: Add SELinux support non enforcing
see #222 for milestone 2021-Q4. This requires effort in the Garden Linux OS extension.

5kt:
/kind enhancement
/area os
/os garden-linux

Add security=selinux to enable selinux (permissive mode, as default in /etc/selinux/config).

Fixes #333


marwinski:
I don't understand. The security option is deprecated, see https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html. The document rather suggests to use selinux=1

PR #369, opened 2021-09-17
Change dns server in resolv.conf to point to local revolver

### LINKED ISSUE #326: Configure Garden Linux image according to infrastructure provider recommendations [Openstack]

We need to make sure we adhere to the recommendations of the infrastructure providers.

**Why is this needed**:

We have seen subtle issues with DNS resolution, clock diversions, nvme issues, etc. due to wrong or omitted configurations.

marwinski:

/kind bug
/area os
/os garden-linux

Our CC EE colleagues recommend the following options in /etc/resolv.conf:

```
options rotate timeout:1
```

Unfortunately there is no obvious way to configure systemd-resolved to do that, however the systemd-resovled has this functionality built in. This PR changes the symbolic link from /etc/resolv.conf to point to the /run/systemd/resolve/stub-resolv.conf file. 

Fixes #326

**Release note**:

```feature user
Configure openstack image according to the recommendations of our CC EE colleagues.
```

### DIFF
--- a/features/openstack/file.include/etc/systemd/system/systemd-resolved.service.d/resolv.conf
+++ b/features/openstack/file.include/etc/systemd/system/systemd-resolved.service.d/resolv.conf
@@ -1,0 +1,3 @@
+[Service]
+ExecStartPre=+/bin/rm /etc/resolv.conf
+ExecStartPost=+/bin/ln -sf /run/systemd/resolve/stub-resolv.conf /etc/resolv.conf


waldiTM:
Hmm, bekommst du die Änderung nicht direkt im Image hin? Wer überschreibt den Link den?

marwinski:
@waldiTM: good point. The original unit is however in `server/file.include/etc/systemd/system/systemd-resolved.service.d/resolv.conf` and should be changed there. As systemd-resolved is caching we wonder why we don't generally link to `/run/systemd/resolve/stub-resolv.conf`

marwinski:
I tried to include a hardcoded symbolic link to `/run/systemd/resolv/resolv.conf` - build fail. Then I tried to create the symblic link in exec.config:
```
# use resolv.conf maitained by systemd-resolved
rm /etc/resolv.conf
ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf
```

This fails with 

```
rm: cannot remove '/etc/resolv.conf': Device or resource busy
ln: failed to create symbolic link '/etc/resolv.conf': File exists
```

So, not ideal but this PR is probably the way to go for now.

gehoern (approved):

Issue #329, opened 2021-08-23
Disambiguate Image Names for OpenStack


marwinski:
**What would you like to be added**:

Our image flavour `openstack` is really tailored for a specific OpenStack implementation (based on VMware) that is certainly not generic and will most likely not work for other OpenStack environments.

I am not sure whether we would be able to have a more generic OpenStack image but until then I propose to name the OpenStack images

`openstack-<<flavour>>`

**Why is this needed**:

The current name is misleading as the image will most likely not work for arbitrary OpenStack implementations.

marwinski:
Proposal: leave `openstack` as is and add a feature `converged-cloud`

Vincinator:
@marwinski should i follow up on this?

PR #372, opened 2021-09-17
docs: clearly mention OpenStack reference implementation

MrBatschner:

/kind documentation
/area os
/os garden-linux

This PR adds some documentation which clearly mentions that the current OpenStack feature set is a reference
implementation that is designed for the (not so common) CCEE setup with ESXi as hypervisor.

**Release note**:

```feature user
docs: clearly mention OpenStack reference implementation
```

### DIFF
--- a/docs/openstack.md
+++ b/docs/openstack.md
@@ -1,4 +1,12 @@
-# Create OpenStack Image for CC EE
+# GardenLinux for OpenStack
+
+A word of **WARNING**: Since OpenStack is an open environment, we can only provide a reference implementation at this point.
+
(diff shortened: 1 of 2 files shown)


marwinski (changes requested):
We agreed to rename the feature in #329. This would probably not fit then!?

gehoern (approved):

MrBatschner:
@marwinski: we talked about it to put this in for now and @gehoern already approved these changes - once you approve, we can merge it

marwinski (approved):

PR #386, opened 2021-09-27
reenabling netcat for debugging

gehoern:
netcat was deprecated in Debian. We use anyhow most of the time socat, but some SRE requesting netcat to be present in Gardener environments.


marwinski (approved):
thanks

PR #459, opened 2021-10-27
Always upload component-descriptor (if configured)

AndreasBurger:
Previously, the component-descriptor publication was limited even if configured - first to releases, then to release-candidates. This PR finally removes this limitation, as our consumers downstream have been configured to react only to releases.


jensh007 (approved):
re-check if this is still useful to have. Otherwise approved

gehoern (approved):

PR #487, opened 2021-11-08
Ci: Move each branches pipelines to their own namespace

AndreasBurger:
Moves the pipelines created by each Git-branch to their own k8s-namespace. The name of the namespace is derived from the branch-name by replacing all non-alphanumeric characters with `-` (Note: This will still fail if the branch starts/ends with a non-alphanumerical character, but at least the error message should be self-explanatory).

Also provides a script that can autmatically clean-up namespaces that did not see any testrun within the last 20 days (not yet added to pipeline-def)

### DIFF
--- a/ci/cleanup_namespaces
+++ b/ci/cleanup_namespaces
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+own_dir="$(readlink -f "$(dirname "${0}")")"
+repo_root="${own_dir}/.."
+
+source "${own_dir}/lib.sh"
+
+install_kubectl
+install_tkn
+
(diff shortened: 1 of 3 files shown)


ci/cleanup_namespaces
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+own_dir="$(readlink -f "$(dirname "${0}")")"
+repo_root="${own_dir}/.."
+
+source "${own_dir}/lib.sh"
+
+install_kubectl
```
jensh007, line 9: Now we have a second file containing these ugly install xyz commands. I would prefer to ensure that these tools are available in our environment (concourse image) and get rid of those commands.

jensh007 (approved):

PR #504, opened 2021-11-16
docker/Makefile: no longer depending on host having dpkg

nkraetzschmar:

/kind bug
/area os
/os garden-linux

`docker/Makefile` currently depends on `dpkg` being available on the host to determine BUILDARCH env for build container.
This breaks if the build system is not debian or debian based.
This PR should fix it by only using `uname`.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -15,7 +15,7 @@ build-image: needslim
 
 .PHONY: build
 build: needslim
-	@docker build --build-arg BUILDARCH="$$(dpkg --print-architecture | awk -F- "{ print \$$NF }")" -t gardenlinux/build $(ALTNAME_INTERNAL) build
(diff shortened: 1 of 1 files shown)


docker/Makefile
```diff
@@ -15,7 +15,7 @@ build-image: needslim
 
 .PHONY: build
 build: needslim
-	@docker build --build-arg BUILDARCH="$$(dpkg --print-architecture | awk -F- "{ print \$$NF }")" -t gardenlinux/build $(ALTNAME_INTERNAL) build
+	@docker build --build-arg BUILDARCH="$$([ "$$(uname -m)" = "aarch64" ] && echo "arm64" || echo "amd64")" -t gardenlinux/build $(ALTNAME_INTERNAL) build
```
MrBatschner, line 18: Is it safe to assume that anything that is not `aarch64` is `amd64`?

gehoern (approved):

PR #522, opened 2021-11-19
Use serial console as default output for cloud

waldiTM:
/kind bug
/area os
/os garden-linux

Cloud providers use serial console output as kind of a log, often even with write access.
However access to the graphical console is much less provided.
So we should make sure the serial console becomes default and gets the most information.

Making both more equal in information can be done with plymouth.


waldiTM:
This was changed in #35 without any explanation by @marwinski.

Vincinator (approved):

gehoern (approved):

gehoern (approved):

marwinski:
@waldiTM I believe there was a very good reason for that change. Not sure what exactly but I will try to dig it out.

PR #529, opened 2021-11-26
new kernel build script

Vincinator:

* One script to build different kernel versions
* parallelised signing of source files
* Retrieve gpgkeys from kernel.org git [1] as recommended by Luis Claudio R.
Gonçalves
* Pin Debian Kernel Buildenv Version (salsa) together with KERNEL_VERSION in VERSION
file for reproducibility
* Use quilt instead of 'patch -p1' to patch our patches on top of salsa patches.
* Added Linux kernel makefile targets for linux-5.4, linux-5.10 and
linux-5.4-signed, linux-5.10-signed to Makefile.inside
* Added packages/download-linux script and download-linux-* targets to
packages/Makefile. This downloads the same sources as the build script
would, but without triggering the build. As an Debug Option, you can
create a debug container via make manual-debug that has packages/downloads/ volume mapped to
/downloads. Inside that build container you can use for example make
linux-5.10-local to copy already downloaded sources from
/downloads/linux-5.10.84 instead of downloading them again. This is only
a debug option to prevent a re-download of sources when you (need
to) throw away your build container. Also note, as per default the
scripts check if the sources are already available (which would never be
the case for a fresh container).
* fix: in the case we are using a newer RT patchset then salsa, and
this patchset was signed by a new maintainer's pubkey not trusted by the older
salsa debian/upstream/rt-sigining-key.pgp, we add trusted gpg keys
specified in manual/linux.d/gpgkeys to this keyring. pubkeys are
downloaded from kernel.org git.

Version Upgrades
* Upgrade to Linux 5.10.84
* Upgrade to Linux 5.10.83-rt58 patchset for linux 5.10.84 kernel
* Upgrade to Linux 5.4.154
* Added preview Linux 5.15 (low hanging fruit - because of salsa supporting 5.15)

[1] https://lore.kernel.org/linux-rt-users/CAD8J--_aDCF1fjtoqiC2vqOS65LPc0tos+gcveMkQnEskrjfqQ@mail.gmail.com/T/#t

* make targets for linux kernel packages of `packages/Makefile.inside` still the same. 
* Trigger kernel build as before
```
cd packages
make manual
# inside container to build 5.10 kernel:
make linux-5.10
make linux-5.10-signed
```


packages/manual/VERSION
```diff
@@ -0,0 +1 @@
+/home/dev/manual/linux.d/versions/VERSION-5.10
```
nkraetzschmar, line 1: maybe use relative symbolic link, so it doesn't break if we ever change the build user or directory structure

packages/manual/linux.d/patches/5.10.78/kernel/fpga-ofs.patch
```diff
@@ -0,0 +1,12501 @@
+diff --git a/Documentation/ABI/testing/sysfs-bus-dfl-devices-emif b/Documentation/ABI/testing/sysfs-bus-dfl-devices-emif
```
nkraetzschmar, line 1: just symlink to `../../5.10.70/kernel/fpga-ofs.patch`?

packages/manual/VERSION
Vincinator: True, this should not be checked in. 

The symbolic link is created via the Makefile for each target, and uses `$(realpath manual)` to get the absolute path to the manual folder relativ to the makefile.
1. https://github.com/gardenlinux/gardenlinux/blob/18c9703be4c0bf9b9e2df0879fd429db3c797bf0/packages/Makefile#L2
2. https://github.com/gardenlinux/gardenlinux/blob/18c9703be4c0bf9b9e2df0879fd429db3c797bf0/packages/Makefile#L51

What do you think?

packages/manual/linux.d/patches/5.10.78/kernel/fpga-ofs.patch
Vincinator: Agreed. If we have the same patch for multiple versions, we should symlink to the first version of the patch file.

5.10.78 was work in progress - removed it.

Vincinator:
Note: this PR will be updated. We do not want to replace the whole config, we should patch the config instead. (kudos to @gehoern )

also other improvements will be added before this PR is ready.

gehoern (approved):

PR #538, opened 2021-12-01
Reproducible disk images without loopback devices

nkraetzschmar:

/kind enhancement 
/area os
/os garden-linux

- Makes rootfs.raw disk images reproducible by setting all timestamps to the garden linux epoch and deterministically computing all UUIDS etc.
- Include files in ext4 and vfat partition file images directly (using `mke2fs` for ext4 and `mcopy` for vfat), without mounting any partitions &rArr; no longer requiring loopback devices.

**Notes**:

In theory we should now also be able to build in a container without `--privileged` flag, but this somehow breaks test `tiger`, so leaving `--privileged` on for now.


waldiTM:
Last time I looked at this two or three years ago, `mke2fs -d` was not able to create all file metadata. For example extended attributes where missing, so no longer any capabilities for `ping`.

Does this work now?

gehoern (approved):

PR #563, opened 2021-12-15
Disable selinux for gardener

5kt:

/kind fix
/area os
/os garden-linux

We need to disable selinux for gardener. 
Fixing the packages-musthave test in order to properly handle excluded packages.

Fixes #

**Release note**:

### DIFF
--- a/features/gardener/pkg.exclude
+++ b/features/gardener/pkg.exclude
@@ -1,0 +1,2 @@
+selinux-basics
+selinux-policy-default
--- a/features/server/file.include/etc/kernel/cmdline.d/90-lsm.cfg
+++ b/features/server/file.include/etc/kernel/cmdline.d/90-lsm.cfg
@@ -1,0 +1,1 @@
+CMDLINE_LINUX="$CMDLINE_LINUX selinux=1 apparmor=0"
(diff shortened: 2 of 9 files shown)


nkraetzschmar (approved):

gehoern (approved):

features/gardener/pkg.exclude
```diff
@@ -0,0 +1,2 @@
+selinux-basics
+selinux-policy-default
```
waldiTM, line 2: Why do you want to exclude those packages?

features/server/file.include/etc/kernel/cmdline.d/90-lsm.cfg
```diff
@@ -0,0 +1 @@
+CMDLINE_LINUX="$CMDLINE_LINUX selinux=1 apparmor=0"
```
waldiTM, line 1: Hmm, this does not qualify as "remove", as someone can expect from the commit message.

5kt: True, the wording is unfortunate.

5kt:
There have been some misunderstandings regarding these changes, closing.

features/gardener/pkg.exclude
5kt: They should not be removed, this was a misunderstanding.

PR #567, opened 2021-12-21
Ci notify fix

jensh007:

/kind enhancement
/area os
/os garden-linux

CI: fix a problem that notifications are no longer sent if an error in certain tasks happens

Fixes #

**Release note**:

### DIFF
--- a/ci/steps/attach_logs.py
+++ b/ci/steps/attach_logs.py
@@ -253,13 +253,14 @@ def upload_logs(
     )
     if ok:
         tasks = build_tasks.split(',')
-        build_dict = dict()
-        for task_name in tasks:
-            build_dict[task_name] = logs.get_task_result(
-                task_runs_dict=pipeline_run,
-                task_ref_name=task_name,
-                result_name='build_result',
-            )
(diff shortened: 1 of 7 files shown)


ci/steps/attach_logs.py
```diff
@@ -253,9 +252,24 @@ def upload_logs(
         only_failed=False,
     )
     if ok:
+        tasks = build_tasks.split(',')
+        build_dict = dict()
```
ccwienk, line 256: stylistic nit - how about using a dict-comprehension:
```
build_dict = {
  task_name: logs.get_task_result(
    task_runs_dict=pipeline_run,
    task_ref_name=task_name,
    result_name='build_result'
  ) for task_name in tasks
}
```

also, maybe rather name it `task_logs` or sth. similar (reflecting the contents)?

ccwienk (approved):
see inline nit

ci/steps/attach_logs.py
jensh007: Agreed.

PR #568, opened 2021-12-22
Add Additional Resources to Component-Descriptor

AndreasBurger:
Adds references to the rootfs-archives to the component-descriptor in addition to the existing virtual image file references. Also adds a reference to the release-manifest-set, if one is available.

### DIFF
--- a/ci/steps/component_descriptor.py
+++ b/ci/steps/component_descriptor.py
@@ -1,5 +1,6 @@
 import dataclasses
 import logging
+import os
 import pprint
 import sys
 import typing
@@ -46,6 +47,11 @@ def _calculate_effective_version(
         return f'{version}-{committish}'
 
 
+def _is_finalized_version(version):
+    parsed = version_util.parse_to_semver(version)
+    return parsed.finalize_version()
+
+
 def build_component_descriptor(
     version: str,
     committish: str,
@@ -66,7 +72,9 @@ def build_component_descriptor(
         sys.exit(0)
 
     cicd_cfg = glci.util.cicd_cfg(cfg_name=cicd_cfg_name)
-    flavour_set = glci.util.flavour_set(flavour_set_name='all')
+
+    flavour_set_name = 'all'
+    flavour_set = glci.util.flavour_set(flavour_set_name=flavour_set_name)
 
     find_releases = glci.util.preconfigured(
         func=glci.util.find_releases,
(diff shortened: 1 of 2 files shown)


ci/steps/component_descriptor.py
```diff
@@ -225,15 +288,78 @@ def virtual_machine_image_resource(
...
+    resource_access = cm.S3Access(
+        type=cm.AccessType.S3,
+        bucketName=bucket_name,
+        objectKey=rootfs_file_path.s3_key,
+    )
+
+    resource_type = 'application/tar+vm-image-rootfs'
+
+    return cm.Resource(
+        name='rootfs_tar_xz',
```
ccwienk, line 319: nit: how about naming this just `rootfs`?

ccwienk:
naming-nit, otherwise lgtm

ci/steps/component_descriptor.py
```diff
@@ -46,12 +47,20 @@ def _calculate_effective_version(
         return f'{version}-{committish}'
 
 
+def _is_finalized_version(version):
+    parsed = version_util.parse_to_semver(version)
+    if parsed.finalize_version() == parsed:
+        return True
+    return False
+
```
jensh007, line 55: How about?
return parsed.finalize_version() == parsed

ci/steps/component_descriptor.py
```diff
@@ -66,7 +75,10 @@ def build_component_descriptor(
         sys.exit(0)
 
     cicd_cfg = glci.util.cicd_cfg(cfg_name=cicd_cfg_name)
-    flavour_set = glci.util.flavour_set(flavour_set_name='all')
+
+    flavour_set_name = flavourset
+
+    flavour_set = glci.util.flavour_set(flavour_set_name=flavour_set_name)
```
jensh007, line 81: We have this pattern at many places and it may be arguable, but I consider it bad practice to overwrite parameters and use them as variable.

jensh007 (approved):
Added some minor comments.

ci/steps/component_descriptor.py
AndreasBurger: done

ci/steps/component_descriptor.py
AndreasBurger: good catch. Done

ci/steps/component_descriptor.py
AndreasBurger: I think the proper fix here would be to rename the `flavourset` param to `flavour_set_name` (which is what it _actually_ is).

Both the name and the actual flavour set would still be required though, since the two manifest-finding methods differ in their expected parameters in this regard, but at least the renaming done here could be avoided.

I also think that this renaming is out of the scope of this PR, but I'll prepare another branch for it immediately.

PR #570, opened 2021-12-22
Ci: Rename "flavourset" parameter

AndreasBurger:
This renames the `flavourset` parameter to `flavour_set_name` since it is **not** the actual flavour-set that is passed around, but rather its name.


jensh007 (changes requested):
I think you need a rebase. In general I agree to the renaming and appreciate it.

ccwienk:
I do not quite get this change. before, it was named `flavourset_name`, now you inserted another underscore to get `flavour_set_name`. from the PR's description, I would have assumed it was called `flavourset` before (which in fact would be misleading)

jensh007:
@ccwienk: The meaningful change is in ci/params.py from `flavourset`to `flavourset_name`. There are some more renamings in other files where it was named differently. I personally do not care too much about `flavourset_name` compared to 'flavour_set_name`. But it would indeed save a few files to be touched.

jensh007 (approved):

PR #574, opened 2022-01-06
use local packages for dev / testing

nkraetzschmar:

/kind enhancement
/area os
/os garden-linux

- adds option to provide local packages, overloading the ones from the repository
  - this allows to more easily test packages before deploying them to the repository

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,12 @@ AWS_DISTRIBUTE=
 BUILDDIR=.build
 MAINTAINER_EMAIL="contact@gardenlinux.io"
 
+ifneq ($(wildcard .packages),)
+LOCAL_PKGS=.packages
+endif
+
 ifdef LOCAL_PKGS
+$(info using local packages from $(LOCAL_PKGS))
 override BUILD_OPTS += --local-pkgs=$(LOCAL_PKGS)
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,8 @@ AWS_DISTRIBUTE=
 BUILDDIR=.build
 MAINTAINER_EMAIL="contact@gardenlinux.io"
 
-ifneq ($(wildcard .packages),)
-LOCAL_PKGS=.packages
+ifneq ($(wildcard local_packages),)
+LOCAL_PKGS=local_packages
 endif
 
 ifdef LOCAL_PKGS
--- a/Makefile
+++ b/Makefile
@@ -5,12 +5,16 @@ AWS_DISTRIBUTE=
 BUILDDIR=.build
 MAINTAINER_EMAIL="contact@gardenlinux.io"
 
+ifdef LOCAL_PKGS
+override BUILD_OPTS += --local-pkgs=$(LOCAL_PKGS)
+endif
+
 .PHONY:all
 all: all_dev all_prod
 
 cert/sign.pub:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -5,12 +5,16 @@ AWS_DISTRIBUTE=
 BUILDDIR=.build
 MAINTAINER_EMAIL="contact@gardenlinux.io"
 
+ifdef LOCAL_PKGS
+override BUILD_OPTS += --local-pkgs=$(LOCAL_PKGS)
```
waldiTM, line 9: I'm sceptical about this "get a directory from an environment variable". Wouldn't it be more useful if there would exist an directory in the build tree where a proper README can reside and where you just need to put the file?

waldiTM (changes requested):

gehoern (approved):

bin/garden-init
```diff
@@ -146,9 +146,19 @@ if ! "$debootstrap" "${debootstrapArgs[@]}"; then
 fi
 echo "$epoch" > "$targetDir/garden-epoch"
 
+if [ -v PKG_DIR ]; then
```
waldiTM, line 149: `[ -v` is a bash-ism. Please use at least `[[ -v` to make it explicit or stick to `[ "$var" ]`.

waldiTM (changes requested):
Please also squash related commits. Like if you need to fix something (like 95292a153a90ff641cd748e0ea717d544ab9aa69), it belongs in the commit that broke it in the first place (f5ac9b5883b9c3469ffb58394a9a1f5d24f6da84)

Makefile
```diff
@@ -5,12 +5,21 @@ AWS_DISTRIBUTE=
 BUILDDIR=.build
 MAINTAINER_EMAIL="contact@gardenlinux.io"
 
+ifneq ($(wildcard .packages),)
+LOCAL_PKGS=.packages
```
waldiTM, line 9: Please don't use a "hidden" directory. People might not see it and it's contents and will be surprised.

Also, please create that directory and add a README.

waldiTM (changes requested):

Makefile
mbanck: Can't we just re-use `./packages` for that?

waldiTM: I would use `./local_packages`.

waldiTM: For comparison: the Debian tools use `./localdeb`.

Vincinator: I think this discussion is unrelated to this change.
The output of the local packages builds is currently```.packages```.
[see](https://github.com/gardenlinux/gardenlinux/blob/48abbeddee267cb829b29802a5b9c0ee83866c41/packages/Makefile#L1)

However, I second @waldiTM argument not to use a hidden directory..

mbanck: personally, I find a top-level directory for this a bit polluting, so would vote for `./packages/localdeb`, but I'll refrain from further bikeshedding here.

I agree with @waldiTM that a hidden directory is probably not a good idea

bin/garden-init
nkraetzschmar: `[ "$var" ]` won't work because we have `set -u`
and the `-v` syntax seems clearer / more legible than some `[ -n "${var+x}" ]` parameter expansion magic
but yeah, using `[[ ... ]]` is probably a good idea :+1:

waldiTM (approved):

PR #581, opened 2022-01-17
Use systemd-repart instead of growroot

5kt:
/kind enhancement
/area os
/os garden-linux

Remove cloud-guest-utils (and growroot) as we switch to systemd-repart for the automatic resize of the root partition.

**Release note**:


gehoern (approved):

marwinski:
Of course I am all in for removing the cloud-guest-utils but it would be great to have this tested with the integration tests that check that growpart/growroot work correctly.

PR #597, opened 2022-01-20
Adjusted documentation/README fixes #593

### LINKED ISSUE #593: Improve the build requirements chapter in the README file
We noticed that the `Build Requirements` chapter in the README (see: https://github.com/gardenlinux/gardenlinux/blob/6d1223a03031684b0bd6ed05d3f07921ee393e46/README.md?plain=1#L51) may need some minor adjustments which would help to mitigate misunderstandings.

For example, the memory requirement of +2 GiB only applies to builds that uses the `--lessram` parameter. This should be clarified.
In addition, the README should mention that once a developer wants to install tekton via pip, he must ensure to install pip before by executing `apt install python-pip`

**Why is this needed**:
Improve documentation

florianpaulhoberg:
**Overview**
  - Adjusted sources that are accessed to obtain needed bins/pkgs/imgs etc.
  - Adjusted dep-pkgs for installing 'tekton'
  - Added build options table overview
  - Some smaller adjustments

/kind enhancement
/area os
/os garden-linux

Adjusting the primary documentation / README.md. This gives new users an easier way to test and build *Garden Linux*.

Fixes #593 

**Release note**:

```feature user
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,12 +45,27 @@ Garden Linux is a [Debian](https://debian.org) derivate that aims to provide a s
 
 ## Build Requirements
 
-The entire build runs in a docker container (well a privileged one with extended capabilities - since we need loop back support)
-We can run on any system supporting Docker and having loopback support and has
+The entire build runs in a docker container (well a privileged one with extended capabilities - since we need loop back support).
+We can run on any system supporting Docker and having loopback support and has:
 
-- 2+ GiB (use RAM-disk; use fs with sparse-file support)
+- 2+ GiB (use '--lessram' to lower memory usage)
 - 10+ GiB free disk space
-- Internet connection to access snapshot.debian.org and repo.gardenlinux.io
+- Internet connection (snapshot.debian.org, deb.debian.org, repo.gardenlinux.io, docker.io, golang.org, gopkg.in, github.com)
+
+## Build Options
+
+| Option | Description  |
+|---|---|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,12 +45,27 @@ Garden Linux is a [Debian](https://debian.org) derivate that aims to provide a s
...
+The entire build runs in a docker container (well a privileged one with extended capabilities - since we need loop back support).
+We can run on any system supporting Docker and having loopback support and has:
 
-- 2+ GiB (use RAM-disk; use fs with sparse-file support)
+- 2+ GiB (use '--lessram' to lower memory usage)
 - 10+ GiB free disk space
-- Internet connection to access snapshot.debian.org and repo.gardenlinux.io
+- Internet connection (snapshot.debian.org, deb.debian.org, repo.gardenlinux.io, docker.io, golang.org, gopkg.in, github.com)
+
+## Build Options
```
nanory, line 55: Can we move the `Build Options` chapter right before the `Quick Start` chapter? I think this makes more sense since it is now placed directly within the Build Requirements chapter, splitting it into two parts (see `Required packages for a convenient build` for example).

nanory (changes requested):

README.md
florianpaulhoberg: Makes sense. Done

nanory (approved):
Looks good for me now.

PR #600, opened 2022-01-20
Drop packages build

waldiTM:
Packages are now built in there own projects, so we don't need it in the central pipeline anymore.

/kind cleanup
/area os
/os garden-linux


Vincinator (changes requested):
We also don't need `docker/build-deb` anymore

Vincinator (approved):
Looks good to me

gehoern:
@Vincinator in theory you are right, that some docker containers can drop, but since we are planing to offer some container images soon i would just need the docker folder untouched .... there needs to be a bigger rework.
@waldiTM can you leave the docker folder off? thanks

Vincinator:
@gehoern Why do we need the docker/build-deb for any base container? 

It just installs the devscripts package. 

I think build-deb is only used by the package builds, which is removed by this PR. 

With regards to the base images, we can also discuss here: 

https://github.com/gardenlinux/gardenlinux/pull/587

PR #632, opened 2022-01-31
Fix some random verity problems

waldiTM:
Make verity mode more useful.


features/_fstab_verity/fstab
```diff
@@ -0,0 +1,4 @@
+# <file system>    <dir>        <type>    <options>     <makeimg agrs>
+LABEL=EFI          /boot/efi    vfat      umask=0077    type=uefi
+LABEL=ROOT         /            ext4      ro            verity
```
nkraetzschmar, line 3: changing mount options of verity devices to `ro` makes sense, but why do we need to move the fstab to a separate feature?

waldiTM: Why do you want to have hundred identical files for no purpose?

waldiTM: But after I looked too deep into `makepart`, maybe it would be better to just have scripts instead of this (incomplete) abstraction.

PR #645, opened 2022-02-10
Set selinux context for /etc/vertiytab

### LINKED ISSUE #641: Missing SELinux label for /etc/veritytab
**What happened**:
The file /etc/veritytab is labeled with the selinux label unlabeled_t.
The file /etc/veritytab is labeled with etc_t.

When selinux is installed and enabled in the build system, the file is labeled with etc_runtime_t.
**Environment**:

chrinorse:

/kind bug
/area os
/os garden-linux

Set proper selinux context for /etc/veritytab.

Fixes #641 

**Release note**:

```feature user

```


nanory:
Hi @chrinorse,
we talked about this PR in our meeting yesterday and added @waldiTM as a reviewer, since he had some feedback for the implementation.

bin/makepart
```diff
@@ -89,6 +89,7 @@ sed 's/#.*//;/^[[:blank:]]*$/d' \
 		fi
 		column -t "$fstab" > "$rootfs_work/etc/fstab"
 		column -t "$veritytab" > "$rootfs_work/etc/veritytab"
+		chcon system_u:object_r:etc_t:s0 "$rootfs_work/etc/veritytab"
```
waldiTM, line 92: While this works, I don't think this is a proper solution. It does not use the file table with default contexts, which might even be changed by other changes later.

Maybe the following will work?

* Create an empty `/etc/veritytab` in the proper `file.include`.
* Just append new entries in `makepart`.

waldiTM (changes requested):

gehoern (approved):

PR #652, opened 2022-02-11
git-ignore log files left behind from tests

MrBatschner:

/kind cleanup
/area os
/os garden-linux

Integration tests might leave log files behind which should not make their way into the repository.

**Release note**:

```feature developer
chore: git-ignore logs files from tests
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,7 @@
 .idea
 __pycache__
 /dev
+/test.log
+/test_results.json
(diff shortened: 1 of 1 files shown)


waldiTM:
How can I reproduce this? Because a `make dev_all` does not leave anything behind.

Vincinator (approved):
lgtm

PR #660, opened 2022-02-17
Disable rp_filter due to incompatibility with Cilium

marwinski:

/kind bug
/area os
/os garden-linux

This PR disables the kernel reverse path filtering (rp_filter) which clashes with Cilium.

Fixes #

**Release note**:

```feature bugfix
Disable reverse path filtering (rp_filter) which clashes with Cilium.
```


ScheererJ:
Addresses https://github.com/gardener/gardener-extension-networking-cilium/issues/51.

gehoern (approved):

Issue #613, opened 2022-01-24
Definition of kvm image kind of useless now


waldiTM:
**What happened**:
The `kvm` and `kvm-dev` images are now defined with
- hard immutability using `verity` and
- no container run-time.

So there is no way to install anything.

**What you expected to happen**:
Install `docker` or something like that.

nanory:
Hi @waldiTM ,

you're right. `verity` must be moved to _readonly.

gyptazy:
Hi @waldiTM @nanory,

I just spoke with @waldiTM - we need to discuss what we really want to achieve. I don't want to say the current *kvm-dev* packages are useless since you still may create an overlay fs for further tests but I agree that this should be as an expected default for the *kvm-dev* image. Therefore, I think we have several options:

 * We can create an overlay for this (like */var* etc. for other mnts)
 * We can move verity to feature *_readonly* and include this for needed image types
 * We can just provide docker/containerd binaries and provide a solution for executing images instead
 * Maybe other options?

IMHO, (re)moving verity from *kvm-dev* (to *_readonly*) could be an easy way to solve this. But I'd prefer a solution by using overlays. What do you think?

nanory:
Hi @gyptazy,

as discussed in the last meeting, we will revert the fstab of the `kvm` feature, so that it doesn't contain the dm-verity functionality and ensure that the root partition is writable again.

However, since we need dm-verity, we will move it to the [_readonly](https://github.com/gardenlinux/gardenlinux/tree/main/features/_readonly) feature by doing the following steps:
* Revert to a generic fstab in the `kvm` feature again.
* Add an executable called `fstab_mod` to each modifier feature which modifies a given fstab of a plattform feature (e.g. `kvm`)  and extends it with proper fstab entries needed by the corresponding modifier.
  * In our case, this script would be part of the modifier feature `_readonly` for example.
* Integrate the script to the build pipeline, so that it is called during the creation of the rootfs.

As discussed, I'm assigning this ticket to @gehoern , which will take care of the `fstab_mod` script.

nanory:
Hi @gehoern,

as discussed in our last meeting, a short reminder that we want to adjust the `fstab_mod` script in that way, that it ensures to set the read-write partition at the end of the fstab before it hands it over to the `makepart` script.

Thus, `fstab_mod` ensures that the read write partition can be resized properly.

[...]

PR #673, opened 2022-02-23
look for rw partition and make sure it is last for makedisk input

### LINKED ISSUE #631: /overlay is not resizeable
**What happened**:
In verity mode, the partition for `/overlay` is not last, so can't be resized.

Filesystem for `/overlay` should be last on disk.

`makepart` sort entries according to their mount point.

chrinorse:

/kind bug
/area os
/os garden-linux

Find the rw partition in fstab and move it to be last partition for makedisk input. 
Fixes #631 

A little hackish, but it solves the problem without re-writing the makepart script.
**Release note**:

```feature user

```


nanory:
Hi @chrinorse,

we may modify the partition order in #613. This would make this change obsolet then.

nanory (changes requested):
Hi @chrinorse,

we may modify the partition order in #613. This would make this change obsolet then.

gehoern (approved):

PR #675, opened 2022-02-23
fix: remove cni demo binaries from garden feature

MrBatschner:

/kind cleanup
/area os
/os garden-linux

The package `containernetworking-plugins` just contains "Some reference and example networking plugins" according to their [GitHub project page](https://github.com/containernetworking/plugins). Gardener does not need these, as we are deploying Calico or Cilium as CNI.

**Release note**:

```feature user
fix: remove cni demo binaries from garden feature
```

### DIFF
--- a/features/gardener/pkg.include
+++ b/features/gardener/pkg.include
@@ -1,6 +1,5 @@
 apparmor
 containerd
-containernetworking-plugins
 docker.io
 ethtool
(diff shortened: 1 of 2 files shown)


Vincinator (approved):
sounds reasonable to me.

gehoern (approved):

Issue #648, opened 2022-02-10
Complete FedRAMP feature


nanory:
**What would you like to be added**:
As of now, there is already a [FedRAMP feature](https://github.com/gardenlinux/gardenlinux/tree/main/features/fedramp) but it hasn't been completed yet. It already contains a test, that we would like to integrate in the Garden Linux build.

However, it is important that this test script is executed in the target image (rootfs) instead of executing it on the build system (container). Thereby, the script should be modified as little as possible.

Moreover, the Garden Linux image must probably be adjusted in order to fulfill the FedRAMP test. Therefore, this feature also includes adjustments made to the Garden Linux image, that enables it to fulfill the test.

**Why is this needed**:
This is needed to complete the roadmap goal "Testcases"

zobelhelas:
When trying to complete this feature, one question arises for me:

One check is trying to ensure that we have "PermitEmptyPasswords" explicitly set to "No" in `/etc/ssh/sshd_config`. This value is default, so setting it woul theoretically not neccessary. But to make this check pass, we need to set it explicitly.

Instead of shipping an own sshd_config in the fedramp feature, i wonder if I may add this config value to the existing sshd_config file in the `server` feature, which is included in all other features anyway.

@gehoern, maybe you could comment on this?

zobelhelas:
- [ ] clamav: no default cronjob for clamav from Debian packages
- [ ] limits: default libpam-modules does not have a maxlogin set by default; fedramp specific setting not known
- [ ] aide: no faillog in default aide configuration
- [ ] aide: cronjob via cron.daily/aide
- [ ] chrony: Referenced NTP Server IP is default internal AWS server, not available on all plattforms

gyptazy:
Hey @gehoern,

how shall we proceed with the last two action points (under the aspect that the test script may not be touched):

**krb5_principal**
Validating for a string beginning with `krb5_principal` in `/etc/audit/auditd.conf`. This is the principal for this server. The default is "auditd". Given this default, the server will look for a key named like `auditd/hostname@EXAMPLE.COM` which isn't used in Garden Linux at all.
Check: https://github.com/gardenlinux/gardenlinux/blob/main/features/fedramp/test/test.sh#L80
Doc: https://linux.die.net/man/5/auditd.conf

**chrony NTP Server**
[...]

PR #684, opened 2022-03-01
Split (unit)tests and add running OS support

### LINKED ISSUE #682: Support (unit) tests that may need a running OS
We need support for (unit) tests that may need to test and execute commands within a running OS.
This means that we need to run the final artifact in a VM and run the tests there. Therefore, we need to adjust the test handling to support this.

**Why is this needed**:
With #647 (CIS integration) and #648 (FedRAMP integration) we need to execute these scripts within a running OS (not only a chroot).

### LINKED ISSUE #683: Split (unit) tests from garden-build
Currently the whole logic for (unit) tests is built into *garden-build* which makes it very complex and difficult to handle tests that should be called from a final artifact. Therefore, we should split this and create a new *garden-test* script that contains all the needed tasks. Next to this, this should be called after the artifact/image got created and perform the tests. 

**Why is this needed**:
We will need further tests that may need a final artifact, as well as a executing tests on a running OS.

gyptazy:
This PR contains three changes for:
 * #683: Split (unit) tests from garden-build
 * #682: Support (unit) tests that may need a running OS
 * #647: Example integration for upcoming features (example: cis)

/kind TODO
/area os
/os garden-linux

All changes are needed to solve #648 and #647

Fixes #683 
Fixes #682 
Relates #647

**Release note**:
Still in draft!

- Creates a new *garden-test* script
- *garden-test* will be executed after the final artifact is created
- Takes the rootfs from artifact and modifies it for special use cases (e.g. tests that may need a running OS)
- Watches for *.startvm* files in features; these will be used as a trigger
- For each feature with a *startvm* script, the rootfs will be modified with the needed content and recreated as a image
- Within the docker instance the image will be booted via QEMU
- Remote SSH will perform and validate the executed (unit) tests

ToDo:
- Since I was unable to use the already present *start-vm* script, I directly call *qemu-system-x86_64*. Therefore, I still need to integrate an $arch switch.
- Too many static items
- [...]

```feature user

```

### DIFF
--- a/bin/garden-test
+++ b/bin/garden-test
@@ -243,14 +243,14 @@ for t in $(find $featureDir/*/test/ -maxdepth 1 -type f -executable -exec basena
 				${startvmUser}@localhost -p 2223 	\
 				-o StrictHostKeyChecking=no 		\
 				/home/${startvmUser}/tests/${test}.startvm
-						if [ $? -ne 0 ]; then
-								echo -e "\e[31mfailed\e[39m"
-								echo
-								let "failcounter=failcounter+1"
-						else
-								echo -e "\e[32mpassed\e[39m"
-								echo
-						fi
+			if [ $? -ne 0 ]; then
+				echo -e "\e[31mfailed\e[39m"
+				echo
+				let "failcounter=failcounter+1"
+			else
+				echo -e "\e[32mpassed\e[39m"
+				echo
+			fi
 			# Stop QEMU
 			pkill -F /tmp/qemu.pid
 
--- a/bin/garden-test
+++ b/bin/garden-test
@@ -3,9 +3,9 @@ set -Eeuo pipefail
 
 thisDir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 source "$thisDir/.constants.sh" \
-	--flags 'no-build,debug,skip-tests,suite:,gardenversion:,timestamp:' \
-	--flags 'ports,arch:,qemu,features:,disable-features:,disable-images:,commitid:,userid:,usergid:' \
-	--flags 'suffix:,prefix:' \
+	--flags 'no-build,debug,gardenversion:,timestamp:' \
+	--flags 'arch:features:,disable-features:,disable-images:,commitid:,userid:,usergid:' \
+	--flags 'prefix:' \
 	--
 
 export PATH="${thisDir}:${PATH}"
@@ -93,14 +93,13 @@ sshwatch() {
     echo "Waiting for SSH on port $1 to become ready..."
     while ! ssh-keyscan -p $1 localhost ; do sleep 2 ; done
     echo "SSH on port $1 is ready."
-
 }
 
 # Extract and use pre built
 # rootfs from production
-mkdir -p /rootfs/
-tar xf /output/${prefix}.tar.xz -C /rootfs/
-echo $epoch > /rootfs/garden-epoch
+mkdir -p ${rootfs}/
+tar xf /output/${prefix}.tar.xz -C ${rootfs}/
+echo $epoch > ${rootfs}/garden-epoch
 
 # Start unit tests
 echo "#### tests"
@@ -110,8 +109,8 @@ testcounter=0
 failcounter=0
 skipcounter=0
 
-mkdir -p rootfs/etc/ssl
-mount --bind /etc/ssl rootfs/etc/ssl
+mkdir -p ${rootfs}/etc/ssl
+mount --bind /etc/ssl ${rootfs}/etc/ssl
 garden-apt-get $rootfs update
 
 # Define APT settings and install
@@ -125,11 +124,6 @@ apt-get -y install procps iproute2 rsync openssh-client qemu-system-arm qemu-sys
 for t in $(find $featureDir/*/test/ -maxdepth 1 -type f -executable -exec basename {} \; | grep -v .disable | sort | uniq); do
         let "testcounter=testcounter+1"
         test=$(basename $t | cut -d. -f 1)
-        if [ "${tests:-1}" = "0" ]; then
-                echo "test ${test} is being skipped, --skip-tests has been used"
-                let "skipcounter=skipcounter+1"
-                continue
-        fi
         # go over features and build the enabled/disabled lists
         # a test with .disabled in a specific feature disables the test globally
         # a test that is not executable is not enabled for the specific feature
@@ -185,7 +179,7 @@ for t in $(find $featureDir/*/test/ -maxdepth 1 -type f -executable -exec basena
                 # move the actual tests from one of the features that enables it
                 actualTest="${featureDir}/$(echo ${enabledBy} | awk '{ print $1 }')/test/${t}"
 
-		# Test: chroot 
+                # Test: chroot
                 if [ ${t##*.} == "chroot" ]; then
                         cp -L ${actualTest} "${rootfs}/tmp/${test}"
                         if garden-chroot "${rootfs}" ./tmp/${test}; then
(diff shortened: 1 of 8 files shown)


gyptazy:
Ready for review.

bin/garden-test
```diff
@@ -0,0 +1,280 @@
...
+
+		# Test: startvm (running in qemu)
+                elif [ ${t##*.} == "startvm" ]; then
+                        cp -L ${actualTest} "${rootfs}/tmp/${test}"
+			# Strip trailing whitespace
+			startvmUser=$(echo ${enabledBy} | head -c-1)
+			# Create a temporary and only for unit tests valid
+			# ssh key for each user and feature
+			garden-chroot "${rootfs}" useradd -m ${startvmUser}
+			garden-chroot "${rootfs}" adduser ${startvmUser} wheel
```
nanory, line 208: Please use this instead:
```
useradd -m -G wheel ${startvmUser}
```

bin/garden-test
```diff
@@ -0,0 +1,280 @@
...
+			garden-chroot "${rootfs}" mkdir -p /home/${startvmUser}/tests/
+			garden-chroot "${rootfs}" mkdir -p /home/${startvmUser}/.ssh/
+			garden-chroot "${rootfs}" chmod 700 /home/${startvmUser}/.ssh/
+			/usr/bin/ssh-keygen -q -t rsa -N '' -f /rootfs/home/${startvmUser}/.ssh/id_rsa -m pem
+			cp /rootfs/home/${startvmUser}/.ssh/id_rsa.pub /rootfs/home/${startvmUser}/.ssh/authorized_keys
+			garden-chroot "${rootfs}" chmod 600 /home/${startvmUser}/.ssh/authorized_keys
+			garden-chroot "${rootfs}" chown -R ${startvmUser}:${startvmUser} /home/${startvmUser}/.ssh/
```
nanory, line 215: All of these commands do not require to execute them via `garden-chroot`. You can also use the test-image context for this.

bin/garden-test
```diff
@@ -0,0 +1,280 @@
...
+                        cp -L ${actualTest} "${rootfs}/tmp/${test}"
+			# Strip trailing whitespace
+			startvmUser=$(echo ${enabledBy} | head -c-1)
+			# Create a temporary and only for unit tests valid
+			# ssh key for each user and feature
+			garden-chroot "${rootfs}" useradd -m ${startvmUser}
+			garden-chroot "${rootfs}" adduser ${startvmUser} wheel
+			garden-chroot "${rootfs}" mkdir -p /home/${startvmUser}/tests/
+			garden-chroot "${rootfs}" mkdir -p /home/${startvmUser}/.ssh/
+			garden-chroot "${rootfs}" chmod 700 /home/${startvmUser}/.ssh/
```
nanory, line 211: Please use this instead:
```
mkdir -p ${rootfs}/home/${startvmUser}/.ssh -m 0700
```

bin/garden-test
```diff
@@ -0,0 +1,280 @@
...
+			/usr/bin/ssh-keygen -q -t rsa -N '' -f /rootfs/home/${startvmUser}/.ssh/id_rsa -m pem
+			cp /rootfs/home/${startvmUser}/.ssh/id_rsa.pub /rootfs/home/${startvmUser}/.ssh/authorized_keys
+			garden-chroot "${rootfs}" chmod 600 /home/${startvmUser}/.ssh/authorized_keys
+			garden-chroot "${rootfs}" chown -R ${startvmUser}:${startvmUser} /home/${startvmUser}/.ssh/
+			# Copy whole test directory to features home directory
+			rsync -aP "${featureDir}/${startvmUser}/test/" /rootfs/home/${startvmUser}/tests/
```
nanory, line 217: Why do we copy the tests into the rootfs again, if they are already copied into the rootfs some lines above?
The tests will be underneath the `/tmp/<test>/<test>.d` directory already.

bin/garden-test
```diff
@@ -0,0 +1,280 @@
...
+                                echo "Starting VM for ${arch}"
+                                qemu-system-aarch64 -display none -daemonize -cpu cortex-a72 -machine virt -pidfile /tmp/qemu.pid -bios /usr/share/qemu-efi-aarch64/QEMU_EFI.fd -m 1024M -device virtio-net-pci,netdev=net0,mac=02:9f:ec:22:f8:89 -netdev user,id=net0,hostfwd=tcp::2223-:22,hostname=garden /tmp/test.raw
+                        fi
+
+			# Wait for SSH in VM to be present
+			# Default: tcp/2223
+			echo "Starting VM - this will take some time"
+			sshwatch 2223
+			# Finally perform test
+			ssh -i /rootfs/home/${startvmUser}/.ssh/id_rsa ${startvmUser}@localhost -p 2223 -o StrictHostKeyChecking=no /home/${startvmUser}/tests/${test}.startvm
```
nanory, line 245: Can we split this line into multiple ones?

features/cis/exec.config
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+echo "dummy cis run"
+exit 0
```
nanory, line 5: Is this file really needed right now?

nanory (changes requested):
Hi @gyptazy, I have added some comments to this MR. If something is unclear, we can have a small discussion.

bin/garden-test
gyptazy: Ok, but still within the chroot.

bin/garden-test
gyptazy: Basically it does. *$startvmUser* resolves to a user that is only known within the chroot (created within l207). Changing the user to a not known one will result in a failure. However, I think it's still a good indicator to clarify that this commands are executed in a chroot instead the test-image.

bin/garden-test
gyptazy: Yep, it's */tmp* and the content isn't present anymore when the VM boots up in QEMU. Maybe we should move the overall tests to a new persistent directory since the new test image is dropped in any case. But this should be done in a new PR.

bin/garden-test
nanory: Yes, sound good.

bin/garden-test
```diff
@@ -0,0 +1,280 @@
...
+                        	fi
+                	done
+			# Start new generated image for unit tests
+                        if [ "$arch" = amd64 ]; then
+                                echo "Starting VM for ${arch}"
+                                qemu-system-x86_64 -display none -daemonize -pidfile /tmp/qemu.pid -m 1024M -device virtio-net-pci,netdev=net0,mac=02:9f:ec:22:f8:89 -netdev user,id=net0,hostfwd=tcp::2223-:22,hostname=garden /tmp/test.raw
+                        elif [ "$arch" = arm64 ]; then
+                                echo "Starting VM for ${arch}"
+                                qemu-system-aarch64 -display none -daemonize -cpu cortex-a72 -machine virt -pidfile /tmp/qemu.pid -bios /usr/share/qemu-efi-aarch64/QEMU_EFI.fd -m 1024M -device virtio-net-pci,netdev=net0,mac=02:9f:ec:22:f8:89 -netdev user,id=net0,hostfwd=tcp::2223-:22,hostname=garden /tmp/test.raw
+                        fi
```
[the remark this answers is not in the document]
gyptazy: I think we may create another issue for this to make *start-vm* compatible for this use case. For the first iteration we should go with this implementation.

bin/garden-test
```diff
@@ -0,0 +1,280 @@
...
+mkdir -p rootfs/etc/ssl
+mount --bind /etc/ssl rootfs/etc/ssl
+garden-apt-get $rootfs update
+
+# Define APT settings and install
+# needed dependencies
+sed -i 's/testing/bookworm/g' /etc/apt/sources.list
+rm /etc/apt/apt.conf.d/default-testing
+apt-get update
+apt-get -y install procps iproute2 rsync openssh-client qemu-system-arm qemu-system-x86
```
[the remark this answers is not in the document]
gyptazy: Ok, this speeds it up. But instead of creating a new image I decided to integrate these few packages into the build image.

nanory (approved):

gehoern (approved):

PR #693, opened 2022-03-08
Add SSH Hardening documentation

### LINKED ISSUE #516: SSH Hardening
We need to ensure that the SSH server is configured according to SAP corporate standards

1. Direct root login must be disabled
2. TCP forwarding must be disabled
3. X11 forwarding must be disabled
4. Use public key authentication
5. Use a verbose log level
6. Disable host authentication
7. Use secure cipher suites, kex algorithms and macs

nanory:
/kind enhancement
/area os
/os garden-linux

This PR contains a documentation about the SSH Hardening actions done in Garden Linux. It also explains what SSH configuration parameters have been set (and where) and what the reason were.

Fixes #516

As of now, the current [sshd_config](https://github.com/gardenlinux/gardenlinux/blob/516_add_ssh_hardening_docs/features/server/file.include/etc/ssh/sshd_config#L26-L28) still sets Cipher Suites, Kex Algorithms and MACs explicitly. However, the documentation recommends to use the default algorithms that the OpenSSH implementation provides.

There may be a change in future which affects either these configuration parameters or the corresponding section in the documentation.

### DIFF
--- a/docs/configuration/ssh-hardening.md
+++ b/docs/configuration/ssh-hardening.md
@@ -1,0 +1,90 @@
+---
+title: SSH Hardening
+weight: 10
+disableToc: false
+---
+
+This document highlights which SSH configuration parameters are set in Garden Linux in order to harden it. Thereby, it also explains why specific values have been set and what were the reasons for them.
+
+## Configuration parameters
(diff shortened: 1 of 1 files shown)


gyptazy:
Hey @nanory ,

thanks for this great documentation. LGTM; we just need to make sure the new documentation matches the current configuration.

Regards,
gyptazy

gyptazy (approved):

PR #698, opened 2022-03-11
Refactor ISO feature

5kt:
Removing the gardenlinux-live dracut module and use the already existent dmsquash-live module.
Small fixes/adjustements for the install script.


Vincinator (approved):
tested on physical machine, works as expected
/lgtm

PR #702, opened 2022-03-16
Use podman instead of docker

### LINKED ISSUE #464: No more docker please.
Since the strange license endeavours it is time to let docker go:

Very important only in the build process of Garden Linux. if there is a client (feature) that still needs docker, that is no problem but we do not want to use the functionality ourselfes.

So everything we do we should now do with `containerd` that is fully open. `Containerd` offers all functionality docker has in regards of execution and image pulling. The technology missing is `docker build`. To replace this there are various alternatives: kaniko, buildah, buildkit, podman ...

Kaniko seems to be here most attractive since it is only a container to run. But careful: since we want to avoid containers in containers we need to orchestrate from external via make. Best place to start is to rename the "docker" container to "container" :-D good moment to fix #231

chrinorse:

/kind enhancement
/area os
/os garden-linux

replace docker with podman
Fixes #464

**Release note**:

```feature user

```


tests/integration/test_gardenlinux.py
```diff
@@ -285,7 +285,7 @@ def test_docker(client):
     if exit_code != 0:
         (journal_rc, output, error) = client.execute_command("sudo journactl --no-pager -xu docker.service")
     assert exit_code == 0, f"no {error=} expected"
-    (exit_code, output, error) = client.execute_command("sudo docker run --rm  eu.gcr.io/gardenlinux/gardenlinux:184.0 sh -c 'echo from container'")
+    (exit_code, output, error) = client.execute_command("sudo podman run --rm  eu.gcr.io/gardenlinux/gardenlinux:184.0 sh -c 'echo from container'")
```
MrBatschner, line 288: Do not replace Docker with Podman here, these tests run inside the finished image which still contain Docker and no PodMan.

MrBatschner (changes requested):

docker/Makefile
```diff
@@ -14,21 +14,21 @@ needslim:
 .PHONY: build-image
```
nanory, line 14: We should probably rename this directory, too.
Maybe `container/` would be more suitable?

nanory (changes requested):

nkraetzschmar:
We should also setup `registries.conf` to use a pull through mirror for docker hub, else our self hosted runners will get rate limited again 😅 

There is one running and reachable at http://10.0.2.2:5000 from all runners, but I'm not sure if it works with podman or if it's somehow docker specific

nanory:
Hi @chrinorse,

this change had made me thinking that we should probably add a small hint to the README explaining that one must provide a `registries.conf` pointing to `docker.io` or a mirror of it since podman behaves a little bit different compared to docker in this regards.

Kind regards,
Michael

nanory (approved):

MrBatschner (approved):
/lgtm
And please remember to
/squash

Issue #676, opened 2022-02-24
Use pytest for feature test


nanory:
**What would you like to be added**:
As of now, features may provide tests that are executed at the end of the Garden Linux build, as the following lines show:
* https://github.com/gardenlinux/gardenlinux/blob/main/bin/garden-build#L223L310

What test will be executed is defined by the features enabled, what tests these features provide and what should be done during the tests (e.g. chroot).

However, there is no proper test framework that manages the execution of these tests. If the test script crashes for example, it is hard to determine what the issue was. Therefore, there is a plan to migrate one of the feature tests to `pytest` in order to create a Proof of Concept which shows how the integration of `pytest` could look like.

If it's successful, the other tests could be migrated the same way, but there would be upcoming issues to deal with them. 
In general, `pytest` is already used in the packages build, as one can see here:
* https://gitlab.com/gardenlinux/gardenlinux-package-build

**Why is this needed**:
* Make unit tests more verbose and robust in case of an error.
* Make it easier to identify what causes the test to fail.
* Simplify the creation of more complex tests.

PR #712, opened 2022-03-23
feat(test): add chroot as integration test platform (#709)

### LINKED ISSUE #709: Create new Provider 'chroot' for Integration Tests
A new provider `chroot` should be created for integration tests. This makes it possible for everyone to execute unit tests as integration tests based on the final artifact without any further cloud provider access.

This feature is the future base for the transition (https://github.com/gardenlinux/gardenlinux/issues/676) from unit tests to integration tests. These tests will be executed by the `chroot` platform as unit tests.

**Goal:** All tests via pytest

**Why is this needed**:
 * Tests are executed with pytest
 * This ensures that everyone can execute integration
 * No charges for cloud provider usage
 * Shared tests between unit- & integration tests (based on chroot) that may also be used by other platforms
 * Solves dependency from: https://github.com/gardenlinux/gardenlinux/issues/676

gyptazy:
Add 'chroot' as a new integration platform to perform tests.
  * New 'chroot' integration platform
  * Adjusted 'RemoteClient' to support 'chroot'
  * Adjusted 'test_gardenlinux.py' to support in/exclude for 'chroot' tests

/kind enhancement
/area os
/os garden-linux

This PR adds a new platform provider for integration tests. This allows us to perform tests inside a `chroot`. Next to the new `KVM` this allows users to run integration (and future unit) tests without any cloud platform like GCP, AWS,....  The new `chroot` platform is also the base for the migration (#676) from unit tests to a common Pytest base.

Fixes #709

The provided `test_config.yaml` can be used for testing this feature without any needed changes. Make sure to have a matching tar ball - this can be created by running:
```
make kvm-dev
```

`test_config.yaml` can be removed before merging.

**Release note**:

NONE
```feature user

```

### DIFF
--- a/tests/integration/chroot.py
+++ b/tests/integration/chroot.py
@@ -150,13 +150,9 @@ class CHROOT:
         """ Generate new SSH key for integration test """
         logger.info("Generating new SSH key for integration tests.")
         ssh_key_path = self.config["ssh"]["ssh_key_filepath"]
-        # Private key
-        key = paramiko.RSAKey.generate(2048)
-        key.write_private_key_file(ssh_key_path)
-        # Public key (as authorized_keys)
-        public_key = key.get_base64()
-        with open("/tmp/authorized_keys", "w") as f:
-            f.write("ssh-rsa " + public_key)
+        keyfp = RemoteClient.generate_key_pair(
+            filename = ssh_key_path,
+        )
         logger.info("SSH key for integration tests generated.")
 
 
@@ -219,12 +215,13 @@ class CHROOT:
             logger.error("Could not copy sshd_config_integration_tests to {dir}".format(
               dir=chroot_sshd_cfg_dir))
         # Copy ssh / authorized_keys file for root user
+        local_ssh_key_path = self.config["ssh"]["ssh_key_filepath"] + ".pub"
         chroot_root_dir = tmp_dir + "/root/"
         chroot_ssh_authorized_keys = chroot_root_dir + ".ssh/authorized_keys"
         ssh_authorized_keys = "/tmp/authorized_keys"
         self._create_dir(chroot_root_dir+".ssh", 0o600)
         try:
-            shutil.copyfile("/tmp/authorized_keys", chroot_ssh_authorized_keys)
+            shutil.copyfile(local_ssh_key_path, chroot_ssh_authorized_keys)
             logger.info("Copied authorized_keys to: {dir}".format(
               dir=chroot_ssh_authorized_keys))
         except OSError:
--- a/tests/integration/chroot.py
+++ b/tests/integration/chroot.py
@@ -197,7 +197,7 @@ class CHROOT:
         # Generate chroot sshd directory for run/pid
         chroot_sshd_run_dir = rootfs + "/run/sshd"
         try:
-            os.mkdir(rootfs + "/run/sshd")
+            os.mkdir(chroot_sshd_run_dir)
             logger.info("Created directory: {dir}".format(
               dir=chroot_sshd_run_dir))
         except OSError:
@@ -207,7 +207,7 @@ class CHROOT:
         sshd_config_src_file = "integration/misc/sshd_config_integration_tests"
         chroot_sshd_cfg_dir = rootfs + "/etc/ssh/"
         try:
-            shutil.copyfile(sshd_config_src_file, chroot_sshd_cfg_dir+"sshd_config_integration_tests")
+            shutil.copy(sshd_config_src_file, chroot_sshd_cfg_dir)
             logger.info("Copied sshd_config_integration_tests to: {dir}".format(
               dir=chroot_sshd_cfg_dir))
         except OSError:
@@ -232,13 +232,12 @@ class CHROOT:
         """ Start sshd inside the chroot """
         # Define vars to have it more readable
         gl_chroot_bin = "/gardenlinux/bin/garden-chroot"
-        chroot = rootfs
         chroot_cmd = "/usr/sbin/sshd -D -f /etc/ssh/sshd_config_integration_tests"
         # Execute in Popen as background task
         # while we may perform our integration tests
         proc_exec = "{chroot_bin} {chroot_env} {chroot_cmd}".format(
           chroot_bin=gl_chroot_bin,
-          chroot_env=chroot,
+          chroot_env=rootfs,
           chroot_cmd=chroot_cmd)
         p = subprocess.Popen([proc_exec], shell=True)
         logger.info("Started SSHD in chroot environment.")
@@ -249,12 +248,12 @@ class CHROOT:
         port = self.config["port"]
         logger.info("Waiting for SSHD in chroot to be ready on tcp/{port}...".format(
           port=port))
-        cmd_chroot = "ssh-keyscan -p {port} localhost".format(
+        cmd = "ssh-keyscan -p {port} localhost".format(
           port=port)
 
         rc = 1
         while rc != 0:
-            p = subprocess.run([cmd_chroot], shell=True, stdout=subprocess.PIPE, \
+            p = subprocess.run([cmd], shell=True, stdout=subprocess.PIPE, \
               stderr=subprocess.STDOUT)
             rc = p.returncode
             logger.info(str(p.returncode) + " Waiting...")
--- a/tests/integration/chroot.py
+++ b/tests/integration/chroot.py
@@ -70,13 +70,13 @@ class CHROOT:
         # Perform basic validation
         self._validate()
         # Unarchive defined tar ball
-        tmp_dir = self._unarchive_image()
+        rootfs = self._unarchive_image()
         # Generate temporary keys for integration testing
         self._generate_ssh_key()
         # Adjust chroot to be able to connect
-        self._adjust_chroot(tmp_dir)
+        self._adjust_chroot(rootfs)
         # Start sshd inside the chroot
-        self._start_sshd_chroot(tmp_dir)
+        self._start_sshd_chroot(rootfs)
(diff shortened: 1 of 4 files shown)


tests/integration/chroot.py
```diff
@@ -0,0 +1,298 @@
...
+        except tarfile.CompressionError:
+            logger.error("Used compression is not supported.")
+        except tarfile.HeaderError:
+            logger.error("Malformed tar header information.")
+        logger.info("Unarchived image {image} to {tmp_dir}".format(
+          image=image, tmp_dir=tmp_dir))
+        return tmp_dir
+
+
+    def _generate_ssh_key(self):
```
MrBatschner, line 149: This functionality is needed everywhere. There is a function `RemoteClient.generate_key_pair()` can you reuse that one?

MrBatschner (changes requested):
See above.

Note that there are huge changes to the integration test framework in the branch `testpipeline`, but since you were first committing your changes, I will integrate them into the `testpipeline` branch.

tests/integration/chroot.py
```diff
@@ -0,0 +1,298 @@
...
+            logger.info("Copied sshd_config_integration_tests to: {dir}".format(
+              dir=chroot_sshd_cfg_dir))
+        except OSError:
+            logger.error("Could not copy sshd_config_integration_tests to {dir}".format(
+              dir=chroot_sshd_cfg_dir))
+        # Copy ssh / authorized_keys file for root user
+        chroot_root_dir = tmp_dir + "/root/"
+        chroot_ssh_authorized_keys = chroot_root_dir + ".ssh/authorized_keys"
+        ssh_authorized_keys = "/tmp/authorized_keys"
+        self._create_dir(chroot_root_dir+".ssh", 0o600)
```
nanory, line 225: I would use the following command here:
```
access = 0o600
os.mkdir(path, access)
```
Then, there is no need for this helper function

tests/integration/chroot.py
```diff
@@ -0,0 +1,298 @@
...
+        """ Wait for defined SSH port to become ready in chroot environment """
+        port = self.config["port"]
+        logger.info("Waiting for SSHD in chroot to be ready on tcp/{port}...".format(
+          port=port))
+        cmd_kvm = "ssh-keyscan -p {port} localhost".format(
+          port=port)
+
+        rc = 1
+        while rc != 0:
+            p = subprocess.Popen([cmd_kvm], shell=True, stdout=subprocess.PIPE, \
```
nanory, line 258: Why not `subprocess.run` ?

tests/integration/chroot.py
nanory, line 222: I would use the following command here:
```
access = 0o600
os.mkdir(path, access)
```
Then, there is no need for this helper function

tests/integration/chroot.py
```diff
@@ -0,0 +1,298 @@
...
+        p = subprocess.Popen([proc_exec], shell=True)
+        logger.info("Started SSHD in chroot environment.")
+
+
+    def _wait_sshd(self):
+        """ Wait for defined SSH port to become ready in chroot environment """
+        port = self.config["port"]
+        logger.info("Waiting for SSHD in chroot to be ready on tcp/{port}...".format(
+          port=port))
+        cmd_kvm = "ssh-keyscan -p {port} localhost".format(
```
nanory, line 253: I would rename this variable from `cmd_kvm` to `cmd`

tests/integration/chroot.py
```diff
@@ -0,0 +1,298 @@
...
+            logger.warning("'garden-epoch' already present. This will be set to 'dev', now.")
+        try:
+            with open(chroot_epoch, 'w') as f:
+                f.write("dev")
+            logger.info("Wrote garden-epoch to chroot in {tmp_dir}".format(
+              tmp_dir=tmp_dir))
+        except IOError:
+            logger.error("Could not write garden-epoch to chroot in {tmp_dir}".format(
+              tmp_dir=tmp_dir))
+        # Generate SSH hostkeys for chroot
```
nanory, line 175: For better readability, I would suggest to add a newline in front of each comment section. Similar to the `_validate` method

tests/integration/chroot.py
```diff
@@ -0,0 +1,298 @@
...
+        logger.info("SSH key for integration tests generated.")
+
+
+    def _adjust_chroot(self, tmp_dir):
+        """ Adjust chroot to make it usable """
+        logger.info("Adjusting chroot environment")
+        # Adjust garden-epoch to dev
+        chroot_epoch = tmp_dir + "/" + "garden-epoch"
+        chroot_epoch_bool = os.path.exists(chroot_epoch)
+        if chroot_epoch_bool:
```
nanory, line 165: `chroot_epoch_bool` is only used in this if statement. If there is no other need, I would probably define it like this:
```
if os.path.exists(chroot_epoch):
```

tests/integration/chroot.py
```diff
@@ -0,0 +1,298 @@
...
+        cmd_kvm = "ssh-keyscan -p {port} localhost".format(
+          port=port)
+
+        rc = 1
+        while rc != 0:
+            p = subprocess.Popen([cmd_kvm], shell=True, stdout=subprocess.PIPE, \
+              stderr=subprocess.STDOUT)
+            output = p.stdout.read()
+            output, error = p.communicate()
+            rc = p.returncode
```
nanory, line 262: We should probably harden this while loop. Instead of keeping this running endlessly, we should check the return code or at least the output for cases that are not related for an `sshd` that is still starting (e.g connection issues).

Alternatively, we should add a retry counter and fail if it reached a specific limit.

tests/integration/chroot.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+    def __init__(self, config):
+        """ Creating self config """
+        # Define self.config
+        self.config = config
+        # Perform basic validation
+        self._validate()
+        # Unarchive defined tar ball
+        tmp_dir = self._unarchive_image()
```
nanory, line 73: In order to make it easier to understand what the content of this tmp_dir is, I would suggest to rename this variable and the corresponding parameters from `tmp_dir` to `rootfs`. It makes it easier for someone who already implemented chroot tests in `garden-build` since the variable there is called `rootfs` as well.

nanory (changes requested):

tests/integration/chroot.py
```diff
@@ -0,0 +1,298 @@
...
+        # while we may perform our integration tests
+        proc_exec = "{chroot_bin} {chroot_env} {chroot_cmd}".format(
+          chroot_bin=gl_chroot_bin,
+          chroot_env=chroot,
+          chroot_cmd=chroot_cmd)
+        p = subprocess.Popen([proc_exec], shell=True)
+        logger.info("Started SSHD in chroot environment.")
+
+
+    def _wait_sshd(self):
```
nanory, line 248: This method is used by `kvm` and `chroot`, no?
It may makes sense to move this method to another module. Maybe `sshclient.py`. What to you think?

tests/integration/chroot.py
gyptazy: Okay, done. But I think `generate_key_pair` should be more fail safe. While playing around with it this function has overwritten my private key. Luckily not a big deal in my case but I think there must be a check to validate if the given file is already present and not to overwrite it by default... If some users want to mount their prod key, this can be very risky...

```
$> md5sum /tmp/ssh_priv_key >> /tmp/foo && pytest --iaas=chroot --configfile=test_config.yaml integration/ && md5sum /tmp/ssh_priv_key >> /tmp/foo && cat /tmp/foo
99a5747c2e338b2a94dc5c58b7a369f9  /tmp/ssh_priv_key
e5f57dd4e229e62e222e095053d7963d  /tmp/ssh_priv_key
```

For `chroot` this isn't a big deal (there's only auto generation available) but it may become one for other cloud platforms. Maybe we just raise a issue for this...

tests/integration/chroot.py
gyptazy: There is no need fore, but IMHO it looks much better. Furthermore, there'd be more lines for handling errors and umask.

tests/integration/chroot.py
gyptazy: yep, does not need to run as bg.

tests/integration/chroot.py
gyptazy: Probably more a question regarding the `Style Guide` - while looking at other files it is mostly used this way.

tests/integration/chroot.py
gyptazy: Makes sense.

tests/integration/chroot.py
gyptazy: ok

tests/integration/chroot.py
gyptazy: Ok, but I think we should raise a new issue for this - so we can adjust both ones directly. However, we may also adjust https://github.com/gardenlinux/gardenlinux/pull/712#discussion_r833204784.

tests/integration/chroot.py
gyptazy: Could be solves with https://github.com/gardenlinux/gardenlinux/pull/712#discussion_r833259181

tests/integration/chroot.py
nanory: As discussed in comment https://github.com/gardenlinux/gardenlinux/pull/712#discussion_r833259181, since we will move this method in its own module, I would say we need to adjust this cmd anyways, so if `cmd` is not feasible enough, I would rather call this command `keyscan_cmd` or `ssh_cmd`/`cmd_ssh`

nanory (approved):

PR #717, opened 2022-03-24
Fstabmod

gehoern:
there have been multiple fstab files but they have been all *final* that means no feature could modify the fstab for their needs. Now we switched to a new model:
  During the feature build, using the feature order based on dependencies, a main fstab is identified and all following features can modify that fstab. The original feature stays intact but now also fstab modifiers can be used.
  So this patch converts all static fstabs into one (see feature base) and adds a dynamic change possibility (e.g. feature cloud).
  Also if the fstab is rendered empty (see feature _pxe) the build of the raw file will be disabled and rendering the parameter --disable-images useless, because this can now be explicitly expressed.
  The feature element image has now - next to the fstab - the possibility to build an complete separate image artifact (e.g. 
feature _iso).
  The feature element convert has now been added to provide the possibility to convert the raw file into any kind of simple conversion (e.g. fearture openstack) or more complex (e.g. feature gcp)


Vincinator (commented):
Only small remarks.
Otherwise looks good to me

Vincinator (approved):
/lgtm 

any shellcheck warnings remaining can be cleaned up together with the warnings in other scripts. 
I do not want to block this further

PR #719, opened 2022-03-24
Draft: ci(GitHub): add shellcheck 

Vincinator:
Adding shellcheck to GitHub actions

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,35 @@
+name: shellcheck
+on: [ push, pull_request]  
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - run: |
+        echo "Install shellcheck"
+        echo "deb http://archive.ubuntu.com/ubuntu jammy main restricted universe multiverse" | \
+          sudo tee -a /etc/apt/sources.list \
+        sudo apt update
+        printf "%s\n" \
+          "Package: *" \
+          "Pin: release a=jammy" \
+          "Pin-Priority: 100" \
+        | sudo tee -a /etc/apt/preferences.d/shellcheck.perf
+
+        sudo apt install shellcheck
+    - run: shellcheck --version
+    - run: |
+        echo "Do shellcheck on all *.sh files"
+        status=0
+        set +e 
+        while IFS= read -r -d '' file
+        do
+        first_line=$(head -n 1 "$file")
+        if [[ "$first_line" == *"bash"* ]]; then
+            shellcheck --severity=warning "$file"
+            status=$((status +$?))
+        fi
+        done < <(find . -type f -print0)
+        if [ $status ]; then exit 1; fi
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -9,15 +9,6 @@ jobs:
 
     - run: |
         echo "Install shellcheck"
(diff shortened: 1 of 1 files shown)


.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,31 @@
+name: shellcheck
+on:
+  pull_request:
```
nkraetzschmar, line 3: maybe also on push? that way you can see something is wrong in your own branch before opening a PR

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,31 @@
...
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'     
+        required: true
+        default: 'warning'
+
+jobs:
+  shellcheck:
+    runs-on: gardenlinux-amd64
```
nkraetzschmar, line 13: we can probably just use the free github hosted runners (`runs-on: ubuntu-latest`) for this, shellcheck shouldn't need the additional performance of self hosted ones

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - run: |
+        echo "Install shellcheck"
+        echo "deb http://deb.debian.org/debian testing main contrib non-free" > /etc/apt/sources.list \
```
nkraetzschmar, line 18: on the github hosted runners you can then probably just do update & install shellcheck from default sources

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,31 @@
...
+    - run: shellcheck --version
+    - run: |
+        echo "Do shellcheck on all *.sh files"
+        status=0
+        set +e 
+        while IFS= read -r -d '' file
+        do
+            shellcheck --severity=warning "$file"  
+            status=$((status +$?))
+        done < <(find . -name '*.sh' -print0)
```
nkraetzschmar, line 30: won't this miss a lot of files? most of our scripts don't end in `.sh` so we'll probably need a more clever way to find all.... maybe find all files with executable bit set and then check the `#!` line 🤔

nkraetzschmar:
or you could just cheat and use some pre-made github action like https://github.com/marketplace/actions/shellcheck 😅

.github/workflows/shellcheck.yml
Vincinator: Unfortunately latest ubuntu image available for the GitHub runner is `20.04` and this has an older shellcheck.. but now that I am adding to sources.list anyway, I need to try again

.github/workflows/shellcheck.yml
Vincinator: installs spellcheck 0.7.0

https://manpages.ubuntu.com/manpages/focal/man1/shellcheck.1.html

.github/workflows/shellcheck.yml
Vincinator: good point

.github/workflows/shellcheck.yml
Vincinator: > or you could just cheat and use some pre-made github action like https://github.com/marketplace/actions/shellcheck 😅

Does not work, only GitHub official or GardenLinux owned actions are allowed.. was my first try though :D

Vincinator:
> or you could just cheat and use some pre-made github action like https://github.com/marketplace/actions/shellcheck 😅

Does not work, only GitHub official or GardenLinux owned actions are allowed.. was my first try though :D

Vincinator:
Closing PR in favour for #725

PR #730, opened 2022-03-29
Adopt feature tests to pytest

### LINKED ISSUE #676: Use pytest for feature test

chrinorse:

/kind enhancement
/area os
/os garden-linux

Fixes #676 

**Release note**:

```feature user

```


tests/integration/chroot.py
```diff
@@ -147,10 +147,13 @@ def _generate_ssh_key(self):
         """ Generate new SSH key for integration test """
         logger.info("Generating new SSH key for integration tests.")
         ssh_key_path = self.config["ssh"]["ssh_key_filepath"]
-        keyfp = RemoteClient.generate_key_pair(
-            filename = ssh_key_path,
-        )
-        logger.info("SSH key for integration tests generated.")
+        if not os.path.isfile(ssh_key_path):
```
nanory, line 150: In my opinion, this if  statement should be part of the `generate_key_pair` function itself. Maybe, we could add an additional parameter, that triggers an exception if the key already exists (e.g. `force`)

tests/integration/test_gardenlinux.py
```diff
@@ -10,15 +10,15 @@
...
-from .sshclient import RemoteClient
+from integration.aws import AWS
+from integration.gcp import GCP
+from integration.azure import AZURE
+from integration.openstackccee import OpenStackCCEE
+from integration.chroot import CHROOT
+from integration.kvm import KVM
+from integration.manual import Manual
+from integration.ali import ALI
+from integration.sshclient import RemoteClient
```
nanory, line 21: Would it make sense to move `integration.sshclient` to `helper.sshclient`?
This way, we could probably call `integration` `platforms` but we should discuss this with the team first.

nanory (changes requested):

tests/helper/tests/blacklisted_packages.py
```diff
@@ -0,0 +1,56 @@
...
+                blklst = utils.read_test_config(enabledfeatures, 'blacklisted-packages')
+
+                blacklisted = []
+                for line in blklst:
+                    if line in pkgslist:
+                        blacklisted.append(line)
+
+                if not len(blacklisted) == 0:
+                    cls.failed_before = True
+                    raise TestFailed("%s are blacklisted, but installed" % (', '.join(blacklisted)))
```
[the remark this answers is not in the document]
gyptazy: I think we should stick around with f-strings (even while I personally prefer `str.format()`) since they're already mostly used within this project.

tests/integration/test_gardenlinux.py
nanory: As discussed, we would create a dedicacted issue for this since it isn't directly linked to the issue this PR is resolving.

nanory (changes requested):

nanory (approved):

PR #731, opened 2022-03-29
Add lynis CIS Pytest for Integration Tests - Discussion

gyptazy:

/kind test
/area os
/os garden-linux

This PR adds a basic integration of Cisofy's lynis check within Pytest.
To make sure this one can be executed on all platforms it will be downloaded and executed via the `client` object from `RemoteClient` in `sshclient`. Therefore, we are limited in Python commands on the target environment without installing further apps/libs which would result in a changed artifact. As a result, this minimal test is used.

However, this is like the current ones but I'm not quite sure if this is the way we want to go. Please, let's discuss...
 
**Which issue(s) this PR relates**:
Relates #647

DO NOT MERGE YET

**Release note**:
NONE

```feature user

```

### DIFF
--- a/tests/integration/test_cis.py
+++ b/tests/integration/test_cis.py
@@ -1,0 +1,23 @@
+import logging
+import os
+import pytest
+import yaml
+
+from .sshclient import RemoteClient
+
+LYNIS_VER = "3.0.7"
+
+logger = logging.getLogger(__name__)
+
+def test_download_lynis(client):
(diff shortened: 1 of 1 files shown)


gyptazy:
Hey @Vincinator,
> Would it make sense to organise the tests in modules? If a platform fixture is required, one can import it?
> I noticed that the added tests are located in the integration folder - for now I don't see a big issue, but we may have contradictory tests in the future (e.g. checking login banner). I would prefer to organise via modules, but a naming scheme may also suffice.
> Does it make sense?

Definitely! Thanks for creating the issue #739. I'll add a PR for this within the next days.

gyptazy:
Closing (DUP): https://github.com/gardenlinux/gardenlinux/pull/760

PR #738, opened 2022-03-30
Tests/shellcheck

### LINKED ISSUE #725: Add a shellcheck pytest
An optional [shellcheck](https://github.com/koalaman/shellcheck) pytest that runs `shellcheck` on bash scripts included in this repository. 

- [x] shellcheck pytest
- [x] does run only if explicitly requested
- [x] Check all bash scripts (e.g. all files containing the word "bash" in the first line, or a more sophisticated parsing of `#!...`)
- [x] Configure list of warnings to ignore ( e.g via `SHELLCHECK_OPTS="-e <id> -e <id> ..."`) 
- [x] Configure severity level (e.g. `shellcheck --severity=warning`) 
- [x]  (nice to have)   Option to limit shellcheck test on a set of folders

**Why is this needed**:
- To run it locally and (if enabled) also in pipelines
- Catch common typos and mistakes in bash scripts

Vincinator:
- Runs `shellcheck` on all bash scripts in this repository, except the files defined in `tests/shellcheck/file.ignore`
- shellcheck errors listed in `tests/shellcheck/error.ignore` will be ignored
- If a bash script contains a `#skip-shellcheck` comment in line 2, after the `#!/...` line, it will also be ignored by our shellcheck pytest
- The `severity` flag can be specified for pytest, and is passed on to shellcheck tests
    - `pytest shellcheck --severity=error` 
- [Documentation](https://github.com/gardenlinux/gardenlinux/tree/tests/shellcheck/tests/shellcheck/README.md)

Fixes #725

### DIFF
--- a/tests/shellcheck/test_shellcheck_gl_source.py
+++ b/tests/shellcheck/test_shellcheck_gl_source.py
@@ -14,6 +14,7 @@ from itertools import filterfalse
 logger = logging.getLogger(__name__)
 
 SKIP_COMMENT = "skip-shellcheck"
+SHEBANG_REGEX = r"^#! */[^ ]*/(env *)?[abk]*sh"
 
 
 def is_bash_script(filepath):
@@ -25,7 +26,7 @@ def is_bash_script(filepath):
     except IOError:
         logger.warn('File {filepath} not found')
         return False
-    if 'bash' in head:
+    if re.search(SHEBANG_REGEX, str(head), re.IGNORECASE):
         return True
     return False
 
--- a/tests/shellcheck/test_shellcheck_gl_source.py
+++ b/tests/shellcheck/test_shellcheck_gl_source.py
@@ -1,0 +1,37 @@
+import logging
+import glob
+import os
+import pytest
+from pathlib import Path
+from subprocess import PIPE, run
+
+logger = logging.getLogger(__name__)
+
+
+def is_bash_script(filepath):
+    try:
+        with open(filepath) as file:
+          head = file.readline()
+    except UnicodeDecodeError:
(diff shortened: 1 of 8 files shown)


gyptazy:
Hi @Vincinator,

great PR! Just two notes...

 * Garden Linux runs everything in Docker. Therefore, I suggest we should call this test within the `gardenlinux/integration-test` Container or create a new one that fulfills the prerequisites (Pytest, Shellcheck). Probably it's just adding the `shellcheck` package to the current Container. On the other hand we may create a new Container without all the dependencies of cloud platforms. I think the question is: Slim vs AIO Container and depends on the further usage.
 * I'm not quite sure if this if the right place in order of the upcoming changes (rel. https://github.com/gardenlinux/gardenlinux/tree/main/tests#12-unit-tests). I think there wasn't any case like this yet. But a merge wouldn't break anything. Just to take note that an implementation/usage of this could be adjusted later.

tests/shellcheck/test_shellcheck_gl_source.py
```diff
@@ -0,0 +1,116 @@
...
+SKIP_COMMENT = "skip-shellcheck"
+
+
+def is_bash_script(filepath):
+    try:
+        with open(filepath) as file:
+            head = file.readline()
+    except UnicodeDecodeError:
+        return False  # Ignore binary data
+    if 'bash' in head:
```
gyptazy, line 25: I'm not quite sure if we never would have 'sh' within the shebang line?

maybe
```
if 'bash' or 'sh' in head:
```

gyptazy (approved):

PR #742, opened 2022-03-31
Documentation: FAQ.md and Repository Readme

dnan0s:

/kind enhancement
/area documentation
/os garden-linux

- FAQ created
- Repository README.md

### DIFF
--- a/docs/Repository/README.md
+++ b/docs/Repository/README.md
@@ -21,7 +21,7 @@ While `dev` contains always the latest package versions, the dated-distros conta
 <details>
   <summary>More Info</summary>
   
-Try running [bin/garden-version](bin/garden-version), to get the $days_since_garden_linux_release value:
+Try running [bin/garden-version](../../bin/garden-version), to get the $days_since_garden_linux_release value:
   ```
 # ./bin/garden-version
     dev
--- a/docs/Repository/README.md
+++ b/docs/Repository/README.md
@@ -44,7 +44,7 @@ The Source Line for this example should look like:
 
 ## Misc
 
-- currently we keep old packages in `dev` dist, and I think we are keeping it that way for now.
+- currently we keep old packages in `dev` dist.
 - the repository contains own packages, and a selected set of packages mirrored from `debian testing`
 - garden packages are being built in our gitlab pipelines
 
--- a/docs/Repository/README.md
+++ b/docs/Repository/README.md
@@ -9,14 +9,19 @@ The location of sources are at `/etc/apt/sources.list` respectively in the folde
 
 #### Generally:
 
-	deb http://repo.gardenlinux.io/gardenlinux $GARDENLINUX_EPOCH main
+	deb http://repo.gardenlinux.io/gardenlinux $days_since_garden_linux_release main
 
-`$GARDENLINUX_EPOCH` is generally `dev`, but the Garden Linux Version number can be defined here. It is the number of days since first Garden Linux release.
+`$days_since_garden_linux_release` is the number of days since 31.03.2020.
+
+For each day since we switched to an own package repository, we offer a respective distribution in our repo.gardenlinux.io apt repository.
+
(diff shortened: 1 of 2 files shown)


dnan0s:
Let's see if this works:

/squash

docs/Repository/README.md
```diff
@@ -0,0 +1,46 @@
...
+#### Generally:
+
+	deb http://repo.gardenlinux.io/gardenlinux $GARDENLINUX_EPOCH main
+
+`$GARDENLINUX_EPOCH` is generally `dev`, but the Garden Linux Version number can be defined here. It is the number of days since first Garden Linux release.
+
+<details>
+  <summary>More Info</summary>
+  
+Try running [bin/garden-version](bin/garden-version), to get the $GARDENLINUX_EPOCH value:
```
nkraetzschmar, line 19: relative link, so points to `docs/Repository/bin/garden-version`. maybe change to:
```
[bin/garden-version](../../bin/garden-version)
```

docs/Repository/README.md
```diff
@@ -0,0 +1,46 @@
+## Garden Linux Repositories
+
+Garden Linux is a Debian-based Operating System. Therefore it uses apt.
+
+The location of sources are at `/etc/apt/sources.list` respectively in the folder `/etc/apt/sources.list.d/`
+
+#### Default Repo:
+	deb http://repo.gardenlinux.io/gardenlinux dev main
```
Vincinator, line 8: In a sense you are right, `dev` is the default repo if you want to have always the latest packages. 

However, for long term releases this is a little bit more complicated. The `dev` repo can contain major package updates, but a long term release does not automatically update major version. 

An example is openssl1.1.1 vs openssl3, and the respective dependency to libc versions. 


long story short; maybe we should not use the word "default" here.

docs/Repository/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+#### Generally:
+
+	deb http://repo.gardenlinux.io/gardenlinux $GARDENLINUX_EPOCH main
+
+`$GARDENLINUX_EPOCH` is generally `dev`, but the Garden Linux Version number can be defined here. It is the number of days since first Garden Linux release.
```
Vincinator, line 14: I think we need more explanation here
here is my input:

---

`$days_since_garden_linux_release` is the number of days since 31.03.2020. 

For each day since we switched to an own package repository, we offer a respective distribution in our repo.gardenlinux.io apt repository. 

While `dev` contains always the latest package versions, the dated-distros contain the packages from that day. 

---

docs/Repository/README.md
```diff
@@ -0,0 +1,46 @@
...
+The Source Line for this example should look like:
+
+    deb http://repo.gardenlinux.io/gardenlinux 730.0 main
+	
+
+</details>
+
+## Misc
+
+- currently we keep old packages in `dev` dist, and I think we are keeping it that way for now.
```
Vincinator, line 42: If the package name changes, we do not delete old package names. This is currently (afaik) only the case for our kernel related packages, that contain the full version string


>  "I think we are keeping it that way for now." 

In a documentation I would avoid to tell from a "me" perspective. 

The fact that we are keeping it for now, does not need to be explicitly written out. The documentation will be updated once we change this.

Vincinator (commented):
Hey, thanks again for improving the documentation. 
I have some small remarks, and look bigger than they are.

docs/Repository/README.md
dnan0s: that was actually the first version i had - i was unsure if it works, so i chose the full path. i changed this back to your suggestion.

docs/Repository/README.md
dnan0s: Good point, thank you. I will keep that in mind for future documentation!

Vincinator (approved):
/lgtm

thank you!

PR #746, opened 2022-04-05
Remove cloud-init hotplug

marwinski:

/kind bug
/area os
/os garden-linux

Remove the cloud-init hotplug mechanism for net and block devices. This causes very high CPU load on Kubernetes machines when many pods are restarted and causes volumes not to handled correctly by the CSI driver. This issue has also been resolved by a new version of cloud-init (see https://bugs.launchpad.net/cloud-init/+bug/1946003) but we would like to avoid adding a new cloud-init version to this branch.

Fixes #

**Release note**:

```feature bugfix
Disable cloud-init hotplug mechanism for net and block devices which causes very high CPU load on Kubernetes nodes when many pods are restarted.
```


5kt (approved):

MrBatschner (approved):
/lgtm

PR #748, opened 2022-04-05
clean: delete flavours.md

Vincinator:
The flavours.md is not required anymore. 

Not to confused with `flavours.yaml`, which is currently used by ci (tekton)

dump of grep for flavours.yaml: 
```
./ci/glci/model.py:    A declaration of a set of gardenlinux flavours. Deserialised from `flavours.yaml`.
./ci/paths.py:flavour_cfg_path = os.path.join(repo_root, 'flavours.yaml')
./ci/README.md:The build can handle various variants of build artifacts. These are configured by a flavour set. The flavours are defined in the file flavours.yaml in the root directory of th Git repository. By default there is one set in this file named "all". You can add more sets according to your needs.
./ci/README.md:Here is example to build only the AWS image. Append the following snippet to `flavours.yaml`:
./ci/render_pipelines_and_trigger_job:pipeline_cfg="${repo_root}/flavours.yaml"
./bin/.cicd-cli.py:        help='Flavour set, see: https://github.com/gardenlinux/gardenlinux/blob/main/flavours.yaml'
```


MrBatschner (approved):
/lgtm

PR #750, opened 2022-04-05
fix: remove error message for error string

MrBatschner:

/kind cleanup
/area os
/os garden-linux

Silences the go compilers complaint "Error strings should not be capitalized or end with punctuation" (the rest is gofmt doing its work).

**Release note**:

```feature user
fix: remove error message for error string
```


Vincinator (approved):
/lgtm

PR #752, opened 2022-04-06
add integration tests to build.sh

### LINKED ISSUE #720: Link Garden Linux build with the corresponding pytests
As of now, the existing pytests defined in [tests/](https://github.com/gardenlinux/gardenlinux/tree/main/tests) are separated from the actual Garden Linux build and need to be called separately.

Besides of this, there are simple feature tests which are triggered by the [bin/garden-test](https://github.com/gardenlinux/gardenlinux/blob/main/bin/garden-test) script right after a Garden Linux build. These feature tests typically contain unit tests that show the functionality of each feature of a previously created image.

Since we want to unify both tests to keep the maintenance simple, we must migrate each of the unit test to a corresponding pytest. Issue #676 deals with this and provides a blueprint, showing some of the migrated unit tests.

However, since only unit tests are called directly after a Garden Linux build as of now, we must ensure that the pytests are executed after a build, too.

For this reason, the `build.sh` script must be extended, so that it does not only call the simple unit tests via `garden-test` but also the pytests.

Two new pytest platforms has been introduced to locally run pytests without requiring access to plattforms like AWS, Azure, AliCloud, ...
* KVM #695  
* chroot #709 

These two platforms should be used to execute the migrated pytests right after the Garden Linux build. Thereby, a developer should still have the possibility to decided if tests should be executed or no (`--skip-test`). 

Additionally, it would be useful if a developer could not only choose if tests should be executed but also choose what tests should be executed:
* garden-test (will be removed once all unit tests have been migrated)
* pytest using the chroot platform (for tests that do not require a running image)
* pytest using the KVM platform (for tests, which require a running instance of Garden Linux)

**Why is this needed**:
In future, we want to get rid of the simple unit tests triggered by `garden-test` and only use tests defined in pytest.

chrinorse:

/kind enhancement
/area os
/os garden-linux

start integration and unit tests after the build using pytest in  the chroot and/or kvm environment.
Fixes #720 #772 #775

**Release note**:

```feature user

```


build.sh
```diff
@@ -134,12 +140,102 @@ else
...
 			/opt/gardenlinux/bin/garden-test &
 		wait %1
 	fi
+	if [ ${skip_tests} -eq 1 ] && [[ "${tests}" == *chroot* ]]; then
+		echo "Creating config file for chroot tests"
+		containerName=$(cat /proc/sys/kernel/random/uuid)
+		prefix="$(${thisDir}/bin/garden-feat --featureDir $featureDir --features "$features" --ignore "$disablefeatures" cname)-$arch-$version-$commitid"
+		mkdir -p ${thisDir}/config
+		cat > ${thisDir}/config/${containerName}.yaml << EOF
+chroot:
```
nanory, line 157: In order to keep `build.sh` clean, I would probably move the configuration creation in a dedicated `bin/garden-*` script.

build.sh
```diff
@@ -33,7 +34,10 @@ lessram=
 arch=$(${thisDir}/get_arch.sh)
 features=
 disablefeatures=
-tests=1
+commitid="${commitid:-local}"
+dpkgArch="${arch:-$(dpkg --print-architecture | awk -F- "{ print \$NF }")}"
+skip_tests=1
```
nanory, line 39: We should probably invert this value, because this way it seems that 1 means that tests are skipped and 0 means they are not.

The current implementation however, uses this variable the other way around.

nanory (changes requested):

build.sh
chrinorse: Not sure if that is a good, skip_tests is set to off by default, so it it should be false. In shell/bash true is represented by 0 and false (error) by a none 0 value.
```
true ; echo $?
0
false ; echo $?
1
```

build.sh
```diff
@@ -136,12 +142,38 @@ else
...
 			/opt/gardenlinux/bin/garden-test &
 		wait %1
 	fi
+	if [ ${skip_tests} -eq 0 ] && [[ "${tests}" == *"chroot"* ]]; then
+		echo "Creating config file for chroot tests"
+		containerName=$(cat /proc/sys/kernel/random/uuid)
+		prefix="$(${thisDir}/bin/garden-feat --featureDir $featureDir --features "$features" --ignore "$disablefeatures" cname)-$dpkgArch-$version-$commitid"
+		${thisDir}/bin/garden-integration-test-config chroot ${prefix} ${containerName} ${outputDir}
+		echo "Running pytests in chroot"
+		docker run --cap-add SYS_ADMIN --security-opt apparmor=unconfined \
```
chrinorse, line 159: This should also be executed with podman not docker.

build.sh
```diff
@@ -136,12 +142,38 @@ else
...
+		wait %1
+		rm config/${containerName}.yaml
+	fi
+	if [ ${skip_tests} -eq 0 ] && [[ "${tests}" == *"kvm"* ]]; then
+		echo "Creating config file for KVM tests"
+		containerName=$(cat /proc/sys/kernel/random/uuid)
+		prefix="$(${thisDir}/bin/garden-feat --featureDir $featureDir --features "$features" --ignore "$disablefeatures" cname)-$dpkgArch-$version-$commitid"
+		${thisDir}/bin/garden-integration-test-config kvm ${prefix} ${containerName} ${outputDir}
+		echo "Running pytests in KVM"
+		docker run --name $containerName --rm -v /boot/:/boot \
```
chrinorse, line 172: This should also be executed with podman not docker.

gehoern (approved):

build.sh
```diff
@@ -136,12 +142,38 @@ else
...
+	fi
+	if [ ${skip_tests} -eq 0 ] && [[ "${tests}" == *"kvm"* ]]; then
+		echo "Creating config file for KVM tests"
+		containerName=$(cat /proc/sys/kernel/random/uuid)
+		prefix="$(${thisDir}/bin/garden-feat --featureDir $featureDir --features "$features" --ignore "$disablefeatures" cname)-$dpkgArch-$version-$commitid"
+		${thisDir}/bin/garden-integration-test-config kvm ${prefix} ${containerName} ${outputDir}
+		echo "Running pytests in KVM"
+		sudo podman run --name $containerName --rm -v /boot/:/boot \
+			-v /lib/modules:/lib/modules -v `pwd`:/gardenlinux  \
+			gardenlinux/integration-test:dev \
```
nanory, line 174: We should probably create a dedictaded image for this instead of using `gardenlinux/integration-test:dev` since it contains much more package than we need to perform our tests.

In addition, we should create a `container-test` target in the Makefile to ensure that this new test container (we may call it `gardenlinux/base-test`) already exists once we call build.sh

nanory (changes requested):

gehoern (approved):

nanory (approved):
The change looks good to me now

PR #756, opened 2022-04-08
fix(tests): do not attempt to read empty ssh keys

MrBatschner:

/kind bug
/area os
/os garden-linux

1. With the changes introduced in PR #730, the ssh-key generator for the test framework will run into an error if the ssh-key file exists but is of size zero (which is the case for the platform tests). Therefore, extending the check if a keyfile is already present by additionally checking on its file size too.

1. The `SubscriptionClient` for Azure which we use to translate a subscription name to its id moved packages which required a small change to the test code.

**Release note**:

```feature user
fix(tests): do not attempt to read empty ssh keys
fix(test): honor move of Azure SubscriptionClient
```

### DIFF
--- a/tests/integration/azure.py
+++ b/tests/integration/azure.py
@@ -15,11 +15,13 @@ from azure.core.exceptions import (
 from urllib.request import urlopen
 from urllib.parse import urlparse
 
-from azure.mgmt.subscription import SubscriptionClient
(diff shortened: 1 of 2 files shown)


chrinorse:
lgtm

gehoern (approved):

5kt (approved):

PR #761, opened 2022-04-11
skip tests that do not pass on chroot or kvm

### LINKED ISSUE #757: Check integration tests, some are constantly failing
**What happened**:
Some integration tests in tests/integration/test_gardenlinux.py are constantly failing.
```
FAILED integration/test_gardenlinux.py::test_ntp - AssertionError: no error="...
FAILED integration/test_gardenlinux.py::test_metadata_connection_non_az_non_ali
FAILED integration/test_gardenlinux.py::test_loadavg - AssertionError: Expect...
FAILED integration/test_gardenlinux.py::test_systemctl_no_failed_units - json...
FAILED integration/test_gardenlinux.py::test_startup_time - AssertionError: n...
FAILED integration/test_gardenlinux.py::test_docker - AssertionError: no erro...
FAILED integration/test_gardenlinux.py::test_apparmor - AssertionError: no er...
```
That the tests pass or are skipped.
Run the integration tests.

**Environment**:
Edit: Happens in the chroot and kvm integration test environment.

chrinorse:

/kind bug
/area os
/os garden-linux

Fixes #757 

**Release note**:

```feature user

```


tests/integration/kvm.py
```diff
@@ -185,34 +170,34 @@ def _adjust_kvm(self, ssh_inject):
...
-            cmd_kvm_adj.append("guestfish -a {image} -i chown 0 0 /root/.ssh".format(
-              image=image))
-            cmd_kvm_adj.append("guestfish -a {image} -i chown 0 0 /root/.ssh/authorized_keys".format(
-              image=image))
-            cmd_kvm_adj.append("guestfish -a {image} -i chmod 0700 /root/.ssh".format(
-              image=image))
-            cmd_kvm_adj.append("guestfish -a {image} -i chmod 0600 /root/.ssh/authorized_keys".format(
-              image=image))
+        # Create a snapshot image and inject SSH key
+        cmd_kvm_adj.append("qemu-img create -f qcow2 -F raw -b {image} {image}.snapshot.img".format(
```
nanory, line 174: I would probably store the snapshot to /tmp so that it does not leave the test container. This way, we do not face problems with permissions once we would store it next to the original image.

nanory (changes requested):

nanory (approved):

tests/integration/test_gardenlinux.py
```diff
@@ -264,10 +280,10 @@ def test_growpart(client, openstack, openstack_flavor):
     sgb = int(lines[1].strip()[:-1])
     assert sgb == expected_disk_size, f"partition size expected to be ~{expected_disk_size} GB but is {sgb}"
 
-def test_docker(client):
```
MrBatschner, line 267: Why is the docker test failing on kvm?

tests/integration/test_gardenlinux.py
```diff
@@ -378,7 +394,7 @@ def test_aws_ena_driver(client, aws):
     assert exit_code == 0, f"no {error=} expected"
     assert output.rstrip() == "ena", "Expected network interface to run with ena driver"
 
-def test_apparmor(client):
+def test_apparmor(client, non_chroot, non_kvm):
```
MrBatschner, line 397: Please rebase to latest master as the change introduced in PR#766 should make this test succeed even on KVM.

MrBatschner (changes requested):
see inline and when finished, remeber to
/squash

tests/integration/test_gardenlinux.py
chrinorse: I tested the kvm test with the kvm image, the kvm image is build without the gardener feature. The gardener feature installs docker. 
Today I tried to test the gcp-gardener image with kvm, docker is installed, but the image has only about 8MB of space left, the docker test tries to pull an image that is >25MB, the test fails with no space left on device.
I guess we have two options,

1. Not running the test in kvm or
2. check if the feature gardener was used for building the image and resize the image for the test.

tests/integration/test_gardenlinux.py
chrinorse: I tested the kvm test with the kvm image, the kvm image is build without the gardener, chost and khost feature. The gardener, chost and khost feature install apparmor .
I tested with the gcp-gardener image with the kvm image and the apparmor test passes. Since the test is connected to features that are not enabled in all images we need a test to check if a feature that installs apparmor is enabled.
Or maybe we can check in the apparmor test first if apparmor is installed, if not the test passes, too.

MrBatschner: Ok, I believe that the test fails because `aa-enabled` is not installed in the image. Maybe we should change the test to work like the SELinux test by looking into `/sys/kernel/security/lsm` instead. If the image runs with SELinux, this test should not fail because of a missing `aa-enabled` but actually succeed.

tests/integration/test_gardenlinux.py
MrBatschner: Option two sounds good. I believe we should have these tests for the Gardener feature already run in KVM. Right now, the Gardener feature is the most important one for us as we are still primarily releasing Garden Linux for use with Gardener.

MrBatschner (approved):
/lgtm

tests/integration/test_gardenlinux.py
chrinorse: changed the test and the kvm environment to make the test work in kvm as suggested in option 2.

MrBatschner:
@chrinorse You need to rebase this PR to include the changes of PR#790.

PR #764, opened 2022-04-12
Masking String <website> to extract Documentation to Garden Linux Website

dnan0s:

/kind enhancement,discussion
/area documentation
/os garden-linux

I am currently trying to extract relevant Documentation from this repository to the future Garden Linux Website. 

To keep only one source of truth, no extra folder for website shall be created.
Further, relevant parts of existing Documentation-Files must be correctly identified in order to be extracted.
I suggest using `<website>` as a masking string for website relevant parts. A python Script then searches for this `<website>`-paragraph to integrate it into the website.

If there is no `<website>`, the whole file gets crafted into the website.

For now i expect each readme to only contain one of those <website> constructs - if there will be a need for more, i'll implement another extra function.

---

### DIFF
--- a/features/server/test/sshd.d/ssh_expected
+++ b/features/server/test/sshd.d/ssh_expected
@@ -3,8 +3,8 @@ HostKey /etc/ssh/ssh_host_ed25519_key
 HostKey /etc/ssh/ssh_host_rsa_key
 # HostKey /etc/ssh/ssh_host_ecdsa_key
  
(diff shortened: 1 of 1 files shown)


gyptazy:
Great idea, maybe this could be switched to more common closing tag like `</website>`.

This would result in:
```
<website>
content
</website>
```

instead of:
```
<website>
content
<website>
```

which could be more problematic in understanding when multiple parts should be added to a website from a single long file and would probably handle the parsing easier. Therefore, the handling for multiple website anchors could be easily done, which I think will be needed later when the whole documentation gets extended more and more.

dnan0s:
Good point - i also think, there could be more tags - maybe using <website:subsitename>, e.g. <website:build> might also be a good idea, if different pages need text from one file.

dnan0s:
well, i think i accidentally closed this PR - i'll put in more changes and reopen it soon, then.

PR #767, opened 2022-04-14
fix: metal/cloud exec.confing

5kt:
/kind bug
/area os
/os garden-linux

exec.config from the metal/cloud features are missing the shebang, plus the build process should be aborted if there are any issues.


Vincinator (approved):
/lgtm

PR #774, opened 2022-04-19
Migrate sshd test to pytest

### LINKED ISSUE #773: migrate sshd test to pytest
migrate the existing sshd.chroot to pytest

**Why is this needed**:

chrinorse:

/kind enhancement
/area os
/os garden-linux

Add the sshd test as a pytest
Fixes #773 

**Release note**:

```feature user

```


tests/helper/tests/sshd.py
```diff
@@ -0,0 +1,139 @@
...
+                            " is set but not expected")
+                    if where == cls.DIFFERENT:
+                        cls.failed_before = True
+                        raise TestFailed(f"For {key} " +
+                            f"'{', '.join(expected_value)}' is expected, " +
+                            f"'{', '.join(sshd_config_value)}' is set")
+
+        return cls.instance
+
+    def _create_dict(self, input):
```
nanory, line 94: If we define these methods as functions of this python module, we don't need to call them with `cls.instance` each time. Since they do not process the class object or any attribute of it, we don't need to pass this information.

tests/helper/tests/sshd.py
```diff
@@ -0,0 +1,139 @@
...
+                                                                quiet=True)
+            assert exit_code == 0, f"no {error=} expected"
+
+            sshd_config = cls._create_dict(cls.instance, output)
+
+            expected = utils.read_test_config(
+                enabled_features, 'sshd', '_expected')
+            sshd_expected = cls._create_dict(cls.instance, expected)
+
+            for key, value in sshd_config.items():
```
nanory, line 70: We should probably call `value` `values` instead because it always contain a list of values.

nanory (changes requested):

nanory (approved):

Issue #513, opened 2021-11-17
Ensure no sensitive content is present in images


5kt:
Unencrypted password safes, like .netrc, must not be in the image.

We currently test for this with tiger https://github.com/gardenlinux/gardenlinux/blob/1c78507b03d6cab9ec075878f07aa11c0c229068/features/base/test/tiger.d/tigerrc#L9 but we need to make sure everything is properly covered with this test.

florianpaulhoberg:
Hi @5kt ,
IMHO this sounds like a valid task but the subject is quite misleading in this case. As you already mentioned before, this is already done for *.netrc* files by *Tiger*. While *Tiger* already validates this by looping through all home directories and checking every present .netrc file for credentials, we must ensure an equal process for all other possibilities where passwords are used.

I think this results in:
 * Validating where passwords/secrets are used (relates all sensitive content)
 * Avoiding the usage of them (maybe there're other approaches possible)
 * Ensuring to have them absent
 * Writing *Tiger* test files

Maybe I'm missing something more - feel free to add additional points. Let me know your opinion.

Thanks,
Florian

marwinski:
Well, adding to this we should look for files that contain strings/data with high entropy. Sure, we will have to mask some false positives but this should provide another safety net in case the other tests don't catch it.

florianpaulhoberg:
Ok, so are you all fine with this DoD?

 * Inspecting where sensitive content is used/integrated
 * Looking for solutions to avoid the further usage of them (if possible)
 * Ensuring to have them absent in images
 * Creating *Tiger* files for pipeline integration tests

Maybe we should rename this issue to **Ensure no sensitive content is present in images**?

Are you fine with this @5kt @marwinski?

marwinski:
Fine with me to rename to **Ensure no sensitive content is present in images**

PR #777, opened 2022-04-20
Remove authorized_key from _dev feature

gyptazy:
* Remove authorized_key from _dev feature
 * Adjust permission for home directory to 750

/kind enhancement
/area os
/os garden-linux

We should not ship the `_dev` feature with an `authorized_keys` file. It should be in the operators decision to add something like this if needed. Furthermore, we should directly define correct permissions for the home directory to fit the `CIS` requirements later.

Fixes #

**Release note**:

```feature user
NONE
```


Vincinator:
Yes, I think we should remove the hard coded public key. And maybe add a line to a `.gitignore`. 

From another point of view, the `_dev` feature/modifier is for developer convenience, and I guess that is why we have the `devkey.sh` to generate a fresh key.

gyptazy:
Hey @Vincinator,

that were one of my thoughts, too. However, IMHO Garden Linux should
 * never ship a static key file (including foreign keys)
 * `devkey.sh` is currently unused (and should be adjusted from its current state)
 * not ship images with sensitive content (#513 & #647)

I think we should provide a tool (e.g. `inject_ssh_keys.sh`) in `bin/` which modifies the final artifact afterwards. This could be done in an equal way like we already do this for unit testing on `KVM` platform. Therefore, I think we could create a new issue for creating such a script and make sure to have sensitive absent from any image wether it's built with `_dev` feature.

Vincinator (approved):
agreed

PR #797, opened 2022-04-25
Add libssl1.1 to blacklisted packages

### LINKED ISSUE #796: Check for libssl1.1 and bail out
After we have finished #658, we should explicitly check for the `libssl1.1` package and bail out to prove that all Garden Linux packages have been built against OpenSSL3. Otherwise, the corresponding unit test should fail.

We could use the `blacklisted_packages` test for this.

**Why is this needed**:
This way, we can verify that Garden Linux only uses OpenSSL 3.

nanory:

/kind enhancement
/area os
/os garden-linux

This PR adds `libssl1.1` to the blacklisted packages. Thereby, this change should verify if there are still packages that depend on `libssl1.1`. All Garden Linux packages should depend on `libssl3` instead.

Fixes #796


gyptazy (approved):
/lgtm

PR #798, opened 2022-04-25
Fix kaniko build

AndreasBurger:
This adds another workaround to the current kaniko build: The pre-existing python-binaries will be moved out of reach of the build running in the container (similar to `ùsr/lib` and `cc-utils`) during its run-time.

This will unblock the current issues related to Python 3.10 in the build-deb image.


AndreasBurger:
I will try to free up some time to look into removing Kaniko alltogether, there are far too many hacks necessary here.

MrBatschner (approved):
/lgtm

PR #800, opened 2022-04-27
feat: switchable build container runtime

MrBatschner:

/kind enhancement
/area os
/os garden-linux

With PR #702 we changed the container runtime environment for all build and test containers to Podman to exclusively use free and open-source licenses. However, some users might not be able to use Podman (e.g. because they are not sudoer but members of a `docker` group) but a different container runtime which is compatible to Docker/Podman.

With this PR, by setting the environment variable `GARDENLINUX_BUILD_CRE` to e.g. `docker`, the container runtime used for building Garden Linux can be configured.

**Release note**:

```feature developer
- feat: build container runtime can be switched
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,8 @@ $(info using local packages from $(LOCAL_PKGS))
 override BUILD_OPTS += --local-pkgs=$(LOCAL_PKGS)
 endif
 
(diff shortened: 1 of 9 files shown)


Vincinator (approved):
/lgtm 
tested with metal build

PR #834, opened 2022-04-28
harden kvm test environment

### LINKED ISSUE #803: ssh connection into KVM test environment fails
**What happened**:
when running the pytest in kvm the connection into the running kvm fails, therefore no test are executed.
that the pytest is able to connect into the running image

**Environment**:

chrinorse:

/kind bug
/area os
/os garden-linux

harden/improve the kvm test environment
Fixes #803

**Release note**:

```feature user

```


gyptazy (changes requested):
Probably the related CIS test (`99.3.3.2_hosts_allow`) should be whitelisted while it will fail, now.

chrinorse:
The hosts.allow check only test if the file exists, the content is not checked

gyptazy (approved):

gyptazy:
ah, right. 
lgtm

PR #836, opened 2022-04-28
Fix Kaniko workaround.

AndreasBurger:
This PR fixes a bug introduced with the recent workaround to Kaniko. 
After the first successful base-image-build, the new base-image only contains Python 3.10. Thus, the attempt to copy the binaries to workaround the Kaniko-issues fails. I've added some additional safeguards to the code.


MrBatschner (approved):
/lgtm

PR #840, opened 2022-04-29
fix: Gitter chat badge links to gitter channel

### LINKED ISSUE #158: Use of public Slack / Matrix / ...
Currently a Slack channel on a SAP internal Slack server on gardenlinux.io is linked. A login there is only possible with SAP account or an existing guest account. A registration does not seem to be possible. It would be nice if the Slack from Kubernetes could be used like in the Gardener project. A channel #gardenlinux could be created there. Then the channels for Gardener and Gardenlinux would be in the same place, which makes sense in my opinion.

Vincinator:
Clicking on the badge redirects to the Garden Linux Gitter channel

Fixes #158


MrBatschner (approved):
/lgtm

PR #850, opened 2022-05-03
Migrate legacy test `min-users` to PyTest

### LINKED ISSUE #807: Pytest: Migrate feature test `min-users`
* [features/base/test/min-users](https://github.com/gardenlinux/gardenlinux/blob/main/features/base/test/min-users)

**Why is this needed**:

gyptazy:

/kind test
/area os
/os garden-linux

Migrate legacy test `min-users` to PyTest

Fixes #807

I made the following changes:
 * Naming (`min-users` has no relation to the test. New name: `test_duplicate_uids`)
 * The legacy test just tested for duplicated UID `0` for root. However, there should never be any collision. This will be validated for all users, now.

**Release note**:

```feature user
NONE
```

**Tests**:
Fail:
```
ERROR    helper.tests.find_dup_uids:find_dup_uids.py:81 UID 0 is used multiple times in /etc/passwd
FAILED base/test/test_duplicate_uids.py::test_duplicate_uids - helper.exception.TestFailed: UID 0 is used multiple times in /etc/...
```

OK:
```
PASSED base/test/test_duplicate_uids.py::test_duplicate_uids
```


chrinorse (approved):
lgtm

PR #865, opened 2022-05-05
Enable chroot pytests by default

### LINKED ISSUE #860: Enable chroot pytests by default
As of now, we are currently only executing the old bash unit tests (feature tests) by default once we call `build.sh` or the corresponding `Makefile`. 

The `chroot` or `kvm` pytests however are not executed. This should be adjusted.

**Why is this needed**:
Since some feature test have been migrated to corresponding pytests and therefore have been removed, we should enable the `chroot` tests to ensure that these tests are still executed as expected.

chrinorse:

/kind enhancement
/area os
/os garden-linux

additionally run chroot test by default, remove legacy test that have been ported to pytest 
Fixes #860 

**Release note**:

```feature user

```


tests/integration/test_gardenlinux.py
```diff
@@ -48,7 +48,7 @@ def test_ntp(client, non_azure, non_chroot):
     assert ntp_ok, "NTP not activated"
     assert ntp_synchronised_ok, "NTP not synchronized"
 
-def test_ls(client):
+def test_ls(client, non_arm64):
     (exit_code, output, error) = client.execute_command("ls /")
```
nanory, line 52: We should probably let this test run on ARM64 images as well but we should modify the assertion, that does not work on ARM machines instead.

This way, we still let this test run but adjust it slightly for ARM machines..

nanory (changes requested):

nanory (approved):

PR #887, opened 2022-05-11
Set digest attribute for resources in Component Descriptor

AndreasBurger:
We plan to sign Component Descriptors in the near future. This PR sets the necessary values in the CDs published for Gardenlinux that tell the signing code to skip including the resources in our s3 bucket for the time being.

Also fixes some under-indented lines.


MrBatschner (approved):
/lgtm

PR #890, opened 2022-05-11
Adjust podman documentation in `README.md`

nanory:

/kind enhancement
/area os
/os garden-linux

Adjust the podman documentation in the `README.md`. The example of adding `docker.io` to the `registries.conf` did not work for on system on which the `registries.conf` is only writable for root.

The adjustment takes this into consideration now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ By default, Garden Linux uses [Podman](https://podman.io/) as container runtime 
 
 **Debian/Ubuntu:**
 ```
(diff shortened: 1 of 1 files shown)


gyptazy:
In this case, all other steps should also be `sudo` prefixed. A non-root user would already run in issues earlier (e.g. installing packages via `apt` or `yum`).

gyptazy (approved):
lgtm

PR #892, opened 2022-05-12
feat(pytest): Migrate legacy `autologin` test to `PyTest`

### LINKED ISSUE #822: Pytest: Migrate feature test `autologin`
* [features/server/test/autologin](https://github.com/gardenlinux/gardenlinux/blob/main/features/server/test/autologin)

**Why is this needed**:

gyptazy:
feat(pytest): Migrate legacy `autologin` test to `PyTest`
 * Migrate `autologin` test to `PyTest`
 * Enhance `key_val_in_file` function to invert pattern and ignore files

/kind test
/area os
/os garden-linux

Fixes #822

The enhanced `key_val_in_file` function is also used by other tests like `umask`.
To make it more flexible, two new options have been added:
 * `invert` (Inverting the search pattern result)
 * `ignore_missing` (ignores files that are defined by the decorator but not present)

**Release note**:

```feature user
NONE
```

**Tests**:
Failed:
```
PASSED server/test/test_autologin.py::test_autologin[/system/getty@tty1.service.d/autologin.conf-dict1]
FAILED server/test/test_autologin.py::test_autologin[/etc/systemd/system/serial-getty@.service.d/autologin.conf-dict0] - AssertionError: Found {'ExecStart': 'autologin'} in /e...
```

Passed:
```
PASSED server/test/test_autologin.py::test_autologin[/system/getty@tty1.service.d/autologin.conf-dict1]
PASSED server/test/test_autologin.py::test_autologin[/etc/systemd/system/serial-getty@.service.d/autologin.conf-dict0]
```


features/server/test/test_autologin.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+# Run the test unit to perform the
+# final tests by the given artifact.
+# Testing:
+#  - The key/value search lookup is inverted. As a
+#    resulty any found pattern is an error.
+#  - Files that can not be found are ignored
+#    and will not fail the test.
+def test_autologin(client, file, dict):
```
nanory, line 25: `dict` is problematic because it also a key / type word in Python. As far as I know, the style guide does not allow to use it this way although it may work.

tests/helper/tests/key_val_in_file.py
```diff
@@ -15,7 +22,10 @@ def _get_content_remote_file(client, fname):
 
 def _get_key_val_from_content(content, dict, fname):
     """ Validates if key/val are present in content """
+    present = False
     for k in dict:
         for line in content.split('\n'):
             if line.startswith(k):
```
nanory, line 28: Just checking if a line starts with a given key may not be robust, especially thinking about the use case you are using this function with. 

For example, there are unit directives like `ExecStart`, `ExecStartPre` or `ExecStartPost`, where is if condition would also work with. This creates false positives.

tests/helper/tests/key_val_in_file.py
```diff
@@ -15,7 +22,10 @@ def _get_content_remote_file(client, fname):
...
-                assert dict[k] in line, f"Could not find {dict} in {fname}."
+                if dict[k] in line:
```
nanory, line 29: In my opinion, this if condition is quiet dangerous. It either checks an exact match nor does it ignore the key of a specifc line.

nanory (changes requested):

features/server/test/test_autologin.py
gyptazy: true, thanks.

tests/helper/tests/key_val_in_file.py
gyptazy: Please find the last commit...

tests/helper/tests/key_val_in_file.py
gyptazy: That's what this function should do; only validate if the value is somewhere present. While the exact match for the keys makes sense, only a general presence of the value somewhere by the given line should be done here.

tests/helper/tests/key_val_in_file.py
gyptazy: for the `str.split` to get the first word based on `index 0` and to cut everything after a pattern of `=` or `:` (to have it more generic). But I see your problem, now. I'll rewrite this to only have a single regex pattern.

gyptazy: Ok, adjusted.

nanory (approved):

PR #915, opened 2022-05-18
907: dpkg origin gardenlinux

### LINKED ISSUE #907: Make gardenlinux a Debian derivative - dpkg
**What would you like to be added**
- [x] Add file `/etc/dpkg/origins/gardenlinux`
- [x] Change symlink `/etc/dpkg/origins/default` to point to the gardenlinux file

**Why is this needed**
Gardenlinux is a Debian derivative, and should behave as such. Calling `dpkg-vendor` currently shows Gardenlinux as Debian, but should identify itself as Gardenlinux, a Debian derivative.

zobelhelas:

/kind enhancement
/area os
/os garden-linux

This PR changes the default vendor from Debian to GardenLinux. GardenLinux is not Debian, but a Debian Derivative.

Fixes #907 

**Release note**:
NONE

### DIFF
--- a/features/base/exec.config
+++ b/features/base/exec.config
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
+# set Garden Linux as default for dpkg
+ln -sf /etc/dpkg/origins/gardenlinux /etc/dpkg/origins/default
+
 chmod u-s /bin/umount /bin/mount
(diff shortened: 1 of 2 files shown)


zobelhelas:
> There's probably no further need for `/etc/dpkg/origins/debian` after linking to `/etc/dpkg/origins/gardenlinux`. Therefore, we should remove the Debian related one.

I am currently checking with other deriviatives. I currently tend to leave  `/etc/dpkg/origins/debian` untouched in the images, as Debian is listed as the parent distribution.

gyptazy:
> > There's probably no further need for `/etc/dpkg/origins/debian` after linking to `/etc/dpkg/origins/gardenlinux`. Therefore, we should remove the Debian related one.
> 
> I am currently checking with other deriviatives. I currently tend to leave `/etc/dpkg/origins/debian` untouched in the images, as Debian is listed as the parent distribution.

Looks like the most ones still include the Debian related one. So, nvm :)

waldiTM (approved):

PR #916, opened 2022-05-18
908: Set distribution to gardenlinux in os-release

### LINKED ISSUE #908: Make gardenlinux a Debian derivative - os-release
**What would you like to be added**
- [ ] Change content of `/usr/lib/os-release`

**Why is this needed**
Gardenlinux is a Debian derivative, and should behave as such.

zobelhelas:
/kind enhancement
/area os
/os garden-linux

Make GardenLinux a Debian Derivative

Fixes #908 

**Release note**:
NONE

### DIFF
--- a/features/base/file.include/usr/lib/os-release
+++ b/features/base/file.include/usr/lib/os-release
@@ -1,0 +1,7 @@
+PRETTY_NAME="Garden Linux"
+NAME="GardenLinux"
+ID=gardenlinux
+ID_LIKE=debian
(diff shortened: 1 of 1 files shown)


zobelhelas:
During a recent weekly we spoke about outsourcing that file into a 'base-files' package. I would recommend not doing it in base-files directly, as this would override the existing and constantly patching it. We maybe should think about a base-files-gardenlinux package.

PR #919, opened 2022-05-19
feat(cis): Obtain `CIS` related packages from the Garden Linux repository

### LINKED ISSUE #912: Rebuild `syslog-ng-core` with OpenSSL3 support
**What happened**:
The current package of `yslog-ng-core` is taken from Debian and supports only legacy OpenSSL. We should ensure that this package is compatible with `OpenSSL3` (Garden Linux default).

Should be build with OpenSSL3 support.

### LINKED ISSUE #913: Fix for `CIS` related packages that are taken from Debian instead of GL
**What happened**:
Currently, the `CIS` feature adds a Debian source to obtain further packages. We need to fix this by mirroring these packages to Garden Linux. Afterwards, we can switch to the features `pkg.include` option and can remove the currently used `exec.config` solution. This affects all packages in `https://github.com/gardenlinux/gardenlinux/blob/main/features/cis_packages/exec.config`.

To solve this, we should first rebuild `syslog-ng-core` with `OpenSSL3` as requested in #912.

All used packages should be obtained from the Garden Linux repository.

### LINKED ISSUE #918: Move package `git` from `exec.config` to `pkg.include` in `CIS`
**What happened**:
Currently, the `git` package will be installed via `exec.config` in `cis_packages`. This was done as a workaround to get rid of a non interactive mode. This package should be moved from `exec.config` to `pkg.include` in `cis_packages` but not before #917 is done.

### LINKED ISSUE #927: Add package `sgml-base` to Garden Linux
**What happened**:
Building a Garden Linux image with `CIS` feature results in an issues regarding the new package `syslog-ng`. However, it requires a dependency to `sgml-base` which prevents it from being built. To be honest, I don't understand why this dependency is needed since this is whether required by the package itself (see also https://packages.debian.org/bookworm/syslog-ng), nor by any other dependency. If there would be a direct dependency, this would already be mirrored since we mirror the whole packages including all dependencies (https://gitlab.com/gardenlinux/gardenlinux-package-build/-/blob/main/packages/mirror/debian-include#L88).

All dependencies should be present.

Build GL with feature `CIS`

This is immediately fixed when `sgml-base` is present. This can be validated with local packages:
```
gyptazy@gl-buildbox:~/gardenlinux$ tree local_packages/
local_packages/
└── sgml-base_1.30_all.deb
```

gyptazy:

/kind enhancement
/area os
/os garden-linux

feat(cis): Obtain `CIS` related packages from the Garden Linux repository
 * Use Garden Linux repository for packages instead of injecting Debian repository
 * Use OpenSSL3 rebuild of `syslog-ng` for `CIS`

Fixes #912
Fixes #913
Fixes #918
Fixes #927

**Release note**:

```feature user
- category: bugfix
- target_group: user
```


chrinorse (approved):
lgtm

PR #924, opened 2022-05-19
fix(pkg): Fix pkg install in debootstrap not using `noninteractive` mode

### LINKED ISSUE #917: Package installation with `pkg.include` should use `noninteractive` mode
**What happened**:
When installing packages with `pkg.include` function this may fail when packages require a non interactive mode. Therefore, we should fix this and set the env var `DEBIAN_FRONTEND=noninteractive` to avoid this.

All packages from the Garden Linux repository should be installable in a non interactive way.

Install the package `git` via `png.include`

gyptazy:
fix(pkg): Fix pkg install in debootstrap not using `noninteractive` mode

/kind TODO
/area os
/os garden-linux

This PR fixes that the debootstrap process doesn't use a `noninteractive` mode for installing packages from repository that may regularly need a further dialog.

Fixes #917

Can be tested with the package `git`

**Release note**:

```feature user
bugfix
```

### DIFF
--- a/bin/garden-init
+++ b/bin/garden-init
@@ -189,13 +189,13 @@ fi
 [ "${include_aptfile}" != "" ] &&
 "$thisDir/garden-chroot" "$targetDir" bash -c '
 	cd '${tmp#$targetDir}'
-	INITRD="No" apt-get install -y --allow-downgrades --no-install-recommends -f $1
+	INITRD="No" DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-downgrades --no-install-recommends -f $1
 ' -- "$include_aptfile"
 rm -rf $tmp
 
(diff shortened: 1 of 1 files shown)


Vincinator (approved):
/lgtm 

tested like this:
- added `git` to `features/metal/pkg.include` 
- `make metal` 

**Success** with this PR
Fail without this PR
- without the PR, it failed for me at `Configuring kexec-tools`

gyptazy:
Hey @Vincinator,

thanks for testing and approving this PR. That was the correct way to try this; currently we were working around this by using custom `apt-get install` commands via `exec.config` which should be highly avoided. Therefore, I already fixed this in pending PRs, as well this has to be done within the last one in #918. This'll be fixed todays.

Thanks,
gyptazy

PR #931, opened 2022-05-20
doc: make it clear that integration_tests is only used by tekton

### LINKED ISSUE #713: Requirements.txt for integration_tests

a requirements.txt for pytest inside `/integration_tests`
**Why is this needed**:

as a new user to pytest i installed pytest by `pip3 install pytest`, which brought me a non functioning pytest. I suppose, other users might also run into this issue. I think, Installing the right working versions is a good idea, so a requirements.txt with versions might benefit this issue.

Vincinator:
This PR makes it clear that the integration_tests/ folder is only used by tekton pipelines.
A new developer looking for tests may find the integration_tests/ folder.

Fixes #713


gyptazy (approved):
lgtm

PR #958, opened 2022-05-27
feat(PyTest): Replace legacy `wireguard.chroot` test by PyTest

### LINKED ISSUE #821: Pytest: Migrate feature test `wireguard.chroot`
* [features/cloud/test/wireguard.chroot](https://github.com/gardenlinux/gardenlinux/blob/main/features/cloud/test/wireguard.chroot)

**Why is this needed**:

gyptazy:
feat(PyTest): Replace legacy `wireguard.chroot` test by PyTest
Check for a present `wireguard.ko` Kernel module within
the running kernel. This avoids false positives on bare-metal
systems where multiple Kernel could be present.

/kind enhancement
/area os
/os garden-linux

This PR performs further PyTest tests to validate that `wireguard.ko` Kernel modules are present within the running Kernel.

Fixes #821

**Release note**:

```feature user
NONE
```


tests/helper/tests/wireguard.py
```diff
@@ -0,0 +1,19 @@
...
+
+def wireguard(client):
+    """Validate if the wirefuard kernel module is present"""
+    # Get running kernel version
+    kernel_ver = utils.get_kernel_version(client)
+
+    # Create path to wireguard kernel module
+    kernel_mod_path = f"/lib/modules/{kernel_ver}/kernel/drivers/net/wireguard/"
+    kernel_mod_file = "wireguard.ko"
+    kernel_wireguard_path = f"{kernel_mod_path}{kernel_mod_file}"
```
nanory, line 14: I would suggest to parameterize this tests, so that we can re-use it for other modules as well.

nanory (changes requested):

nanory (approved):

tests/helper/tests/wireguard.py
gyptazy: Goot point... done

Issue #874, opened 2022-05-10
simplify existing pytest


chrinorse:
**What would you like to be added**:
make the structure of the pytest easier and conform.
**Why is this needed**:

PR #965, opened 2022-05-30
simplify existing pytest kernel config

chrinorse:

/kind enhancement
/area os
/os garden-linux

simplify existing pytest kernel config
Fixes #874 

**Release note**:

```feature user

```


gyptazy:
As a result `base/test/test_kernel_config.py` could look like:

```
@pytest.mark.parametrize(
     "file,args",
     [
         ("/boot/config-", {"KEY_TO_CHECK": "VALUE_TO_CHECK"}),
     ]
 )


 def test_kernel_config(client, file, args, non_chroot):
     kernel_ver = get_kernel_version(client)
     file = f"{file}{kernel_ver}"
     key_val_in_file(client, file, args)
```

features/base/test/test_kernel_config.py
```diff
@@ -1,10 +1,25 @@
...
+            }
+        )
+    ]
+)
+
+def test_kernel_config(client, args):
+    """compare kernel config options from the parametrize section with the
+    kernel config options in the /boot/config-* file. The values 'is not set'
+    and 'n' are treated as the same."""
+    file = "/boot/config-*"
```
gyptazy, line 24: Instead of `/boot/config-*` I'd rather take an explicit config file. This could be done by evaluating the boot config in real time. However, this wouldn't work on `chroot` platform. Another solution would be to determinate which config is present in this directory. But I think we may skip this in this test since only freshly generated images will be tested. But this could become a problem by an adjusted image that could be tested by `Pytest`.

gyptazy (approved):
lgtm

PR #977, opened 2022-05-31
feat(pytest): Migrate legacy `dmesg` unit test to Pytest

### LINKED ISSUE #824: Pytest: Migrate feature test `dmesg`
* [features/server/test/dmesg](https://github.com/gardenlinux/gardenlinux/blob/main/features/server/test/dmesg)

**Why is this needed**:

gyptazy:
feat(pytest): Migrate legacy `dmesg` unit test to Pytest
 * Migrate legacy `dmesg` unit test to Pytest
 * Adjust regex in `file_content.py` to match dots in first word group

/kind test
/area os
/os garden-linux

Migrate legacy `dmesg` unit test to Pytest
 * Migrate legacy `dmesg` unit test to Pytest
 * Adjust regex in `file_content.py` to match dots in first word group
 * 
Fixes #824

**Release note**:

```feature user
NONE
```

@5kt, while the current legacy test is broken (see the now deleted file (https://github.com/gardenlinux/gardenlinux/blob/main/features/server/file.include/etc/sysctl.d/allow-nonroot-dmesg.conf) that explicitly allowed this), this is inverted, now. Furthermore, this check expects to enforce this - I think this is what is was supposed to be? Can you confirm?


chrinorse (approved):
lgtm

PR #985, opened 2022-06-02
feat(pytest): Migrate legacy `groups` unit test to Pytest

### LINKED ISSUE #806: Pytest: Migrate feature test `groups.chroot`
* [features/base/test/groups.chroot](https://github.com/gardenlinux/gardenlinux/blob/main/features/base/test/groups.chroot)

**Why is this needed**:

gyptazy:
feat(pytest): Migrate legacy `groups` unit test to Pytest

/kind test
/area os
/os garden-linux

Migrate legacy `groups` unit test to Pytest

Fixes #806

**Release note**:

```feature user
NONE
```


tests/helper/tests/users.py
```diff
@@ -51,3 +55,24 @@ def users(client):
...
+
+
+def _has_user_sudo_cmd(client, user):
+    """ Check if user has any sudo permissions """
+    # Execute command on remote platform
+    cmd = f"sudo -l -U {user}"
+    out = utils.execute_remote_command(client, cmd)
+
+    # Write each line as output in list
+    output_lines = []
```
nanory, line 67: why not simply calling it like this?
```
output_lines = out.split("\n")
```

tests/helper/tests/users.py
```diff
@@ -51,3 +55,24 @@ def users(client):
...
+    output_lines = [] 
+    for line in out.split("\n"):
+       output_lines.append(line)
+
+    # Check if there is enough content in our list
+    # and validate if there are related sudo commands
+    sudo_cmd = False
+    if len(output_lines) > 3:
+        if "may run the following commands on" in output_lines[-2]:
+           sudo_cmd = output_lines[-1]
```
nanory, line 76: I don't know if this is robust enough. What if a user has the privileges to execute multiple commands via sudo. The rule set can be longer than just having one entry, which this implementation assumes. In this case `[-2]` and `[-1]` does not work properly.

tests/helper/tests/users.py
gyptazy: at this point, yes. But the loop will be needed later. tbh, we can reverse the order (as we expect the content within the last lines).

tests/helper/tests/users.py
gyptazy: true.

chrinorse (approved):
lgtm

PR #1017, opened 2022-06-20
fix: use $version instead of hardcoded dev for test container

nkraetzschmar:

/kind bug
/area os
/os garden-linux

`build.sh` was using hardcoded `dev` version of the `base-test` container.
This breaks when building a specific version in a clean build environment where no prior dev builds ran.


MrBatschner (approved):
/lgtm

PR #1018, opened 2022-06-20
fix: use .endswith to check for allowed image extensions

nkraetzschmar:

/kind bug
/area os
/os garden-linux

Currently tests fail with
```
ERROR    integration.chroot:chroot.py:123 0-local.tar.xz is not supported for this platform test type.
```
when using any version other then `dev` since in that case everything after the dot in the version number gets treated as part of the image extension.

This fixes it by using the string `endswith` method instead of manually parsing out the extension from the filename.

### DIFF
--- a/tests/integration/chroot.py
+++ b/tests/integration/chroot.py
@@ -105,20 +105,16 @@ class CHROOT:
               path=self.config["image"]))
 
         # Validate if image extension is defined corretly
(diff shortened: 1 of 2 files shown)


gyptazy:
lgtm

gyptazy (approved):

PR #1023, opened 2022-06-21
Ci: also add digest to rootfs-resources

AndreasBurger:
With this PR, rootfs-resources will also get digest-information added to the component-descriptor.


MrBatschner (approved):
/lgtm

PR #1029, opened 2022-06-23
remove out commented cloud-guest-utils from metal and cloud feature

### LINKED ISSUE #102: replace growpart with systemd repart
A basic feature of a cloud image is: it grows to the disk size it is located on.

Actually this is done via an updated (corrected and more versatile) growpart script in initram. I think the more future proof and even more versatile approach would be to use repart from systemd.

The problem here is repart is not yet part of the debian systemd (as of version 245)
All initial tests have to be done with a recompiled systemd for now

chrinorse:

/kind enhancement
/area os
/os garden-linux

remove out commented cloud-guest-utils from metal and cloud feature
Fixes #102

**Release note**:

```feature user

```

### DIFF
--- a/features/cloud/pkg.include
+++ b/features/cloud/pkg.include
@@ -2,7 +2,6 @@ fdisk
 [${arch}=amd64] syslinux
 [${arch}=amd64] syslinux-common
 #[${arch}=arm64] grub-cloud-arm64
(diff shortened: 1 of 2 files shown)


gyptazy (approved):
lgtm

PR #1058, opened 2022-06-28
feat(startvm): General enhancements for `startvm`

### LINKED ISSUE #1052: ARM64: Misc -> Toolsets around GL -> `startvm` -> Add Support for `HVF` Acceleration
Meta Issue: #1028 

## Toolsets around GL `startvm`
`startvm` is unable to use `HVF` acceleration and currently only support `kvm`. We should also include `HVF` which is used on macOS for a better Apple Silicon support. This is also already used by the `PyTest` Pipeline (see also: #1042).

## Bug
- Misc
  - Toolsets around GL
    - `startvm`
      - Add Support for `HVF` Acceleration

### LINKED ISSUE #1055: ARM64: Misc -> Toolsets around GL -> `startvm` -> Add `Homebrew` Path for Apple Silicon
Meta Issue: #1028 

## Toolsets around GL `startvm`
`startvm` is defined to use `/usr/local/bin/qemu-system-aarch64` and should evaluate if there's another path like the also used `/opt/homebrew/bin/qemu-system-aarch64` path.

## Bug
- Misc
  - Toolsets around GL
    - `startvm`
      - Add `Homebrew` Path for Apple Silicon

### LINKED ISSUE #1056: ARM64: Misc -> Toolsets around GL -> `startvm` -> Handle `arm64` from `uname -m` same as `aarch64`
Meta Issue: #1028 

## Toolsets around GL `startvm`
We should ensure to handle `arm64` outputs from macOS (running Apple Silicon chips) in the sam way, as `aarch64`. While mostly all tools use `aarch64`, we should reset the var to `aarch64` if `arm64` is detected. Example outputs of `uname -m` on macOS and Linux...

**macOS**:
```
~ % uname -a ; uname -m
Darwin 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:29 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T8101 arm64
arm64
```

**Linux**:
```
x$ uname -a ; uname -m
Linux ip-172-31-6-186 5.10.0-15-cloud-arm64 #1 SMP Debian 5.10.120-1 (2022-06-09) aarch64 GNU/Linux
aarch64
```

## Bug
- Misc
  - Toolsets around GL
    - `startvm`
      - Handle `arm64` from `uname -m` same as `aarch64`

### LINKED ISSUE #1079: ARM64: Misc -> Toolsets around GL -> `startvm` -> Get OS related `UEFI` code/vars file
Meta Issue: #1028 

## Get OS related `UEFI` code/vars file
Currently, we have no switch case for `homebrew` path. Therefore, we should evaluate the `PREFIX`. Next to it, we should evaluate the installed `qemu` version to obtain the `UEFI Code` and `UEFI Vars` files.

## Bug
- Misc
  - Toolsets around GL
    - `startvm`
      - Get OS related `uefi` code/vars file

gyptazy:
feat(startvm): General enhancements for `startvm`
* Translate `arm64` to `aarch64` on macOS (Fixes: https://github.com/gardenlinux/gardenlinux/issues/1056)
 * Fix macOS paths for `qemu-system-$arch` binary (Fixes: https://github.com/gardenlinux/gardenlinux/issues/1055)
 * Add `HVF` acceleration on macOS on `AARCH64` systems (Fixes: https://github.com/gardenlinux/gardenlinux/issues/1052)
 * Add eval for HomeBrew path (Fixes: https://github.com/gardenlinux/gardenlinux/issues/1079)
 * Add eval for QEMU version
 * Add eval to obtain correct UEFI code/vars
 * Rewrite `-cpu`, `-host`, `-accel` parts

/kind enhancement
/area os
/os garden-linux

Fixes #1056
Fixes #1055
Fixes #1052
Fixes #1079 

Validated on:
- Linux Debian AMD64
- Linux Debian ARM64
- macOS AMD64
- macOS ARM64

**Release note**:

```feature user
NONE
```

### DIFF
--- a/bin/start-vm
+++ b/bin/start-vm
@@ -16,13 +16,20 @@ set -Eeuo pipefail
 # Define default VARs
 gnu_head="head"
 gnu_stat="stat"
+build_os="$(uname -s)"
+arch="$(uname -m)"
 
 # OS Validation: macOS
-build_os="$(uname -s)"
-if [ "Darwin" == $build_os ]; then
+if [ "Darwin" = $build_os ]; then
         running_os="macos"
         gnu_head="ghead"
         gnu_stat="gstat"
+
+        # Translate ARM64 to AARCH64
+        # on macOS
+        if [ "arm64" = $arch ]; then
+            arch="aarch64"
+        fi
 fi
 
 # OS Validation: CentOS
@@ -40,7 +47,6 @@ thisDir=$(dirname "$(readlink -f "$BASH_SOURCE")")
 targetDir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 tmpDir=${TMPDIR:-/tmp}
 cpu=2
-arch="$(uname -m)"
 memory=2Gi
 monitor=1
 bridge=
@@ -136,25 +142,50 @@ while true; do
 	esac
 done
 
+# Add UEFI options
 if [ "$arch" = x86_64 ]; then
-	uefiCode="${uefiCode:-/usr/share/OVMF/OVMF_CODE.fd}"
-	uefiVars="${uefiVars:-/usr/share/OVMF/OVMF_VARS.fd}"
+
+        if [ "$running_os" = macos ]; then
+                # Test for HomeBrew path
+                homeBrewPrefix=$(brew info qemu | awk 'NR==4{print $1}')
+                uefiCode="${uefiCode:-$homeBrewPrefix/share/qemu/edk2-x86_64-code.fd}"
+                uefiVars="${uefiVars:-$homeBrewPrefix/share/qemu/edk2-i386-vars.fd}"
+        else
+                uefiCode="${uefiCode:-/usr/share/OVMF/OVMF_CODE.fd}"
+                uefiVars="${uefiVars:-/usr/share/OVMF/OVMF_VARS.fd}"
+        fi
+
 elif [ "$arch" = aarch64 ]; then
-	uefi=1
-	uefiCode="${uefiCode:-/usr/share/AAVMF/AAVMF_CODE.fd}"
-	uefiVars="${uefiVars:-/usr/share/AAVMF/AAVMF_VARS.fd}"
+
+        # Enable uefi mode
+        uefi=1
+
+        # Obtain uefiCode and uefiVars from
+        # Homebrew path on macOS
+        if [ "$running_os" = macos ]; then
+                # Test for HomeBrew path
+                homeBrewPrefix=$(brew info qemu | awk 'NR==4{print $1}')
+                uefiCode="${uefiCode:-$homeBrewPrefix/share/qemu/edk2-aarch64-code.fd}"
+                uefiVars="${uefiVars:-$homeBrewPrefix/share/qemu/edk2-arm-vars.fd}"
+        else
+                uefiCode="${uefiCode:-/usr/share/AAVMF/AAVMF_CODE.fd}"
+                uefiVars="${uefiVars:-/usr/share/AAVMF/AAVMF_VARS.fd}"
+        fi
+
 fi
+
 
 if [ "$pxe" ] && [ "$uefi" ]; then
 	pxe_binary="$pxe"
 	pxe=
 fi
 
+
 # Adding the disks (using parameters without - or --)
 inflatelist=( )
 diskcount=0
 
-if [ "$arch" == "x86_64" ]; then
+if [ "$arch" = "x86_64" ]; then
         [ "$#" == "" ] || virtOpts+=(     "-device virtio-scsi-pci,id=scsi0" )
 fi
 
@@ -181,7 +212,7 @@ while (( "$#" )); do
 
 		# if there is a bigger size, we need to inflate
 		[ $($gnu_stat --printf="%s" $imagefile) -lt ${imagesizeBytes} ] && inflatelist+=( "${imagefile},${imagesizeBytes}" )
-                if [ "$arch" == "x86_64" ]; then
+                if [ "$arch" = x86_64 ]; then
 		        virtOpts+=(	"-device scsi-hd,drive=drive${diskcount},bus=scsi0.0"
 			        	"-drive format=${imageext},if=none,discard=unmap,${imagedirect}id=drive${diskcount},file=${imagefile}" )
                 else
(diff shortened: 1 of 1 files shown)


bin/start-vm
```diff
@@ -239,29 +255,40 @@ macfull="$(sed 's/../&:/g;s/:$//' <<< $mac)"
 uuid="12345678-0000-0000-0000-${mac}"
 virtOpts+=(     "-uuid $uuid" )
 
-if [ "macos" != $running_os ]; then
-    if [ "$arch" = "$(uname -m)" ]; then
```
nanory, line 243: Keep in mind that a developer could call `start-vm` for other architectures than the architecture used to execute this script. For example, a developer could run a x86_64 Garden Linux image while being on a ARM64 machine and vice versa. This is achieved by using the `--arch` parameter of `start-vm`, so that `$arch` can be different than `$(uname -m)` in some cases.

However, by removing the implementation in this PR and replacing it with two separate if statements that are only checking if arch is set to x86_64 or aarch64, this functionality would be lost.

nanory (changes requested):

bin/start-vm
gyptazy: Why I can understand this statement, I do not see why this shouldn't work. You can still overwrite the arch. For example (here breakpoint set to just print the output):
```
% uname -m ; bin/start-vm --arch x86_64 .build/kvm_dev-amd64-dev-local.raw
arm64
x86_64
```

Passing the `--arch` as `x86_64` on an `AARCH64` based system.

nanory: The old implementation has set the `-cpu` parameter to `-cpu Broadwell` for x86_64 (while being on arm64) and `-cpu cortex-a72 -machine virt` for arm64 (while being on x86_64). 

As far as I can tell, this functionality isn't given anymore with this PR, no?

bin/start-vm
```diff
@@ -16,13 +16,20 @@ set -Eeuo pipefail
 # Define default VARs
 gnu_head="head"
 gnu_stat="stat"
+build_os="$(uname -s)"
+arch="$(uname -m)"
```
nanory, line 20: Maybe we should use [get_arch.sh](https://github.com/gardenlinux/gardenlinux/blob/main/get_arch.sh) for this. This script is also called in the `Makefile` and `build.sh`, so we would align `start-vm` with both other implementations.

In addition, it also normalizes the architecture, so it always returns arm64, no matter if the host system identifies itself as `arm64` or `aarch64`.

What do you think?

bin/start-vm
gyptazy: You're right, this isn't given anymore because it's useless. Furthermore, this doesn't fail. Why? Let's take a deeper look at this with some reproducible commands. 

## Functionality

### Running on `x86_64` an `aarch64` Guest
First, let's give it a try if it works. We start an `aarch64` based guest on a `x86_64` host.
```
$ uname -m ; bin/start-vm --arch aarch64 test/linux_gardenlinux_kvm-dev_arm64.utm/Images/kvm_dev-arm64-dev-local.raw ; ps aux | grep qemu
x86_64
vagrant   109635 74.4  2.9 3779020 242532 pts/1  Sl+  13:43   0:14 qemu-system-aarch64 -drive if=virtio,format=raw,file=test/linux_gardenlinux_kvm-dev_arm64.utm/Images/kvm_dev-arm64-dev-local.raw -smp 2 -m 2048 -watchdog i6300esb -nodefaults -uuid 12345678-0000-0000-0000-020bb2fb3a63 -cpu max -machine virt -monitor unix:/tmp/020bb2fb3a63.monitor,server,nowait -nographic -serial mon:stdio -drive if=pflash,format=raw,unit=0,file=/usr/share/AAVMF/AAVMF_CODE.fd,readonly=on -drive if=pflash,format=raw,unit=1,file=/home/vagrant/gardenlinux/bin/020bb2fb3a63.vars -device virtio-net-pci,netdev=net0,mac=02:0b:b2:fb:3a:63 -netdev user,id=net0,hostfwd=tcp::2223-:22,hostname=garden
```
Looks pretty good. Up & running...

### Running on `aarch64` a `x86_64` Guest
Next, let's give it a try if it works. Therefore, we start a `x86_64 ` based guest on an `aarch64` host.
```
$ uname -m ; bin/start-vm --arch x86_64 .build/kvm_dev-amd64-dev-local.raw ; ps aux | grep qemu
aarch64
admin     819457  100  0.5 3605400 1461640 pts/0 Sl+  06:43   2:05 qemu-system-x86_64 -device virtio-scsi-pci,id=scsi0 -device scsi-hd,drive=drive0,bus=scsi0.0 -drive format=raw,if=none,discard=unmap,aio=native,cache.direct=on,id=drive0,file=.build/kvm_dev-amd64-dev-local.raw -smp 2 -m 2048 -watchdog i6300esb -nodefaults -device virtio-balloon -device virtio-rng-pci,rng=rng0 -object rng-random,id=rng0,filename=/dev/random -device ipmi-bmc-sim,id=bmc0 -device isa-ipmi-kcs,bmc=bmc0,ioport=0xca2 -chardev socket,path=/tmp/02:c6:37:bf:4e:c7.guest,server=on,wait=off,id=qga0 -device virtio-serial -device virtserialport,chardev=qga0,name=org.qemu.guest_agent.0 -uuid 12345678-0000-0000-0000-02c637bf4ec7 -monitor unix:/tmp/02c637bf4ec7.monitor,server,nowait -nographic -serial mon:stdio -device virtio-net-pci,netdev=net0,mac=02:c6:37:bf:4e:c7 -netdev user,id=net0,hostfwd=tcp::2223-:22,hostname=garden
```
This good, too. Up & running... Why does it work? Because `QEMU` uses the default CPU type on `x86_64` platforms.
 
### CPU Type
You're right, we do not have `-cpu cortex-a72 -machine virt` anymore in that way. But why you want to get rid of this enhancement? Based on a defined `--arch aarch64` arg, we may encounter the following cases:
 * Native Support
 * Emulation

#### Native Support
So, if `--arch aarch64` is se and we have native support, we `do not` want to emulate a specific CPU model/type and slow down for emulation. Therefore, we want to proceed with `-cpu host` in addition of an accelerator like `kvm` oder `hvf`.

#### Emulation
If we're running `--arch aarch64` in emulation on a `x86_64` system, we may `-cpu cortex-a72`. It works, but why not just set it to `-cpu max`? In this case `QEMU` will takeover for the feature probing.

nanory: Ok, looks good to me.

nanory (approved):

nanory (approved):

PR #1060, opened 2022-06-29
add nftables as replacement for iptables

### LINKED ISSUE #972: Garden Linux Firewall: Replace `iptables` by a more common solution
Currently, Garden Linux uses `iptables` for the basic firewall services. While `iptables` works for this it may not fit our way of implementations for further rule sets. Therefore, we should discuss to switch to a more common and comfortable solution like:
 * ferm
 * ufw
 * [...]

**Why is this needed**:
I'd prefer `ferm` with further `ferm.d` (drop-in) support. This would also allow us to inject rules for unit tests via Pytest later instead of all time present static rules.

Opinions here?

chrinorse:

/kind enhancement
/area os
/os garden-linux

Replace ```iptables```with ```nftables```

Fixes #972 

**Release note**:

```feature user

```


features/server/exec.config
```diff
@@ -6,6 +6,7 @@ systemctl enable systemd-resolved
 systemctl enable ssh-keygen
 systemctl enable tmp.mount
 #systemctl enable ssh-moduli
+#systemctl enable nftables
```
gyptazy, line 9: Probably `nftables` should be enabled? (it's commented)

gyptazy (changes requested):

gyptazy (changes requested):
I think we should keep the original IDs like `3.5.1.1` and not increment them by ourselves. We should overwrite the original ones with our content or deactivate them and add them with the same id but with a different name.

gyptazy:
We should also update the related docs like:
* https://github.com/gardenlinux/gardenlinux/blob/main/features/firewall/README.md
* https://github.com/gardenlinux/gardenlinux/blob/main/features/cis/README.md
* https://github.com/gardenlinux/gardenlinux/blob/main/features/fedramp/README.md

features/firewall/README.md
```diff
@@ -1,19 +1,13 @@
 # Feature: cis_firewall
```
nanory, line 1: Since the firewall feature isn't directly related to CIS anymore, we should adjust the header here and change it to `firewall` then.

nanory (changes requested):

features/firewall/README.md
chrinorse: Thanks for pointing this out, I updated the header.

gyptazy (approved):
thanks for the changes @chrinorse. 
lgtm

nanory (approved):

PR #1061, opened 2022-06-29
feature(pytest): Add PyTest for validating the `machine-id`

### LINKED ISSUE #1051: `startvm` - May fail to Access/Initialize `kvm` for Acceleration
**What happened**:
User may encounter issues accessing `kvm` acceleration when general `kvm` support is detected but lacks in permissions:
```
Could not access KVM kernel module: Permission denied
qemu-system-aarch64: -accel kvm: failed to initialize kvm: Permission denied
```

We should catch this error and provide helpful instructions.

Could be fixable in several ways, but IMHO we should not do it ourself while it's in the users decision. Therefore, we should just check and e.g.
 * Fail with message
 * Print warning and proceed w/o acceleration

Possible solutions (just to note them here):
 * `sudo chmod +666 /dev/kvm`
 * `sudo usermod -G -a kvm "username" ; sudo usermod -G -a libvirtd "username"`
 * `udev` rule

gyptazy:
feature(pytest): Add PyTest for validating the `machine-id`
This PyTest validates the systemd unit for creating a machine-id
as well as the content of the created machine-id

/kind enhancement
/area os
/os garden-linux

This PR adds another PyTest for validating the `machine-id`.

Fixes #1051

**Release note**:

```feature user
NONE
```

### DIFF
--- a/features/server/test/test_machine_id.py
+++ b/features/server/test/test_machine_id.py
@@ -1,1 +1,8 @@
-from helper.tests.machine_id import machine_id as test_machine_id
+from helper.tests.machine_id import machine_id
+from helper.tests.machine_id import machine_id_powered_on
+
+def test_machine_id(client, chroot):
+     machine_id(client)
+
+def test_machine_id_powered_on(client, non_chroot):
+     machine_id_powered_on(client)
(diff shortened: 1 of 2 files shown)


chrinorse (approved):
lgtm

PR #1075, opened 2022-06-29
fix: use dynamic version for the base image of integration-test

### LINKED ISSUE #1074: Use correct base image version for `integration-test` container
**What happened**:
As of now, the build of the `integration-test` container fails due to the change of #1059 since the default version changed from `dev` to `today`.

The integration-test container uses a base image called `base-test` that is built once the integration-test container is built. What version is used, is defined in the Dockerfile of the `integration-test` container. As of now, it always uses the `dev` version of the `base-image` since the version has been hard coded.

Due to #1059 , the build fails in main now but in fact this change would also fail in a new release once it would be created, so it must be fixed anyways. 

The `integration-test` container should use the same base image version like the version used to build the `integration-test` container itself. Based on the Makefile, both versions are kept in sync during a build.

Execute the following command from main:
```
make --directory=container build-integration-test
```

nanory:

/kind bug
/area os
/os garden-linux

This PR ensures that the build pipeline uses a dynamic version for referencing the base image of the the `integration-test` container in its Dockerfile. This way, we can ensure that the image version between the `integration-test` and `base-test` (base image of `integration-test`) are kept in sync during the build.

In the past, the base image version has been hard coded and did not work anymore due to a change from `dev` to `today`.

Fixes #1074

### DIFF
--- a/container/Makefile
+++ b/container/Makefile
@@ -40,7 +40,7 @@ build-base-test: needslim
 build-integration-test: build-base-test
 	mkdir -p integration-test/_pipfiles
 	cp ../tests/Pipfile* integration-test/_pipfiles
-	@$(GARDENLINUX_BUILD_CRE) build -t gardenlinux/integration-test:$(VERSION) integration-test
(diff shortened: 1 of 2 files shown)


gyptazy (approved):
lgtm

chrinorse (approved):
lgtm

PR #1077, opened 2022-06-29
Integrate arch for KVM tests properly

nanory:

/kind bug
/area os
/os garden-linux

This PR ensures that the correct architecture is passed to the KVM tests during a build. Previously, the `garden-integration-test-config` returned a wrong configuration which provided a pytest configuration for `amd64` tests although the image has been built as an `arm64` image.


gyptazy (approved):
lgtm

PR #1080, opened 2022-06-30
feature(startvm): Check if `/dev/kvm` can be used by user

### LINKED ISSUE #1051: `startvm` - May fail to Access/Initialize `kvm` for Acceleration

gyptazy:
feature(startvm): Check if `/dev/kvm` can be used by user
 * Check if device is accessible and fallback to non accelerated usage
 * Add documentation how to solve a non accessible device

/kind enhancement
/area os
/os garden-linux

Fixes #1051

**Release note**:

```feature user
NONE
```

### DIFF
--- a/bin/README.md
+++ b/bin/README.md
@@ -70,10 +70,36 @@ This script allows to inject a SSH pubkey to a final Garden Linux image to ensur
 ### start-vm
 This script starts a given `.raw` or `.qcow2` image in a local QEMU/KVM VM and supports `amd64` and `arm64 builds`. Keep in mind, that running different architectures may be very slow. However, it may still be useful for validating and running unit tests. A spawned VM runs in `textmode` which a `hostfwd` (portforward) for SSH on `tcp/2222`. By the given options this allows the user to user copy/paste in the terminal, as well as connecting to the sshd. *(Hint: Custom SSH pubkeys can be injected with `inject-sshkey`.)*
 
+**Acceleration Support:**
+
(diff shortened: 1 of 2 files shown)


bin/README.md
```diff
@@ -70,10 +70,36 @@ This script allows to inject a SSH pubkey to a final Garden Linux image to ensur
...
+| --cpu | CPU type to emulate if an specific should be used |
+| --mem | Memory size to use for VM |
+| --uuefi | Run in `UEFI` mode instead of legacy `Bios`|
+| --ueficode | Path to custom UEFI Code file |
+| --uefivars | Path to custom UEFI Vars file |
+
 **Usage:**
 
-`startvm /path/to/$image_name`
+Running same arch: `startvm /path/to/$image_name`
```
nanory, line 100: It's called `start-vm`, no?

nanory (changes requested):

bin/README.md
gyptazy: Ah, shh... You're absolutely right. fixed && thanks

PR #1082, opened 2022-06-30
enhancement(buildpipeline): Move `get_arch.sh` to `bin/`

### LINKED ISSUE #1081: Build Pipeline: Move `get_arch.sh` to `bin/`
`get_arch.sh` should be moved to `bin/` where all other binaries are located.

gyptazy:

/kind cleanup
/area os
/os garden-linux

Move `get_arch.sh` to `bin/`.
Keep in mind, that PyTest evaluates this via:
`(exit_code, output, error) = client.execute_command("dpkg --print-architecture", quiet=True)`

Fixes #1081

**Release note**:

```feature user
NONE
```


nanory (changes requested):
Since we are moving `get_arch.sh` to `bin/`, we should also extend `bin/README.md` to keep it in sync.

gyptazy:
> Since we are moving `get_arch.sh` to `bin/`, we should also extend `bin/README.md` to keep it in sync.

Good catch…

nanory (approved):

PR #1083, opened 2022-06-30
mount /overlay and etc overlay in initramfs

### LINKED ISSUE #1053: machine-id not set properly in readonly feature
**What happened**:
I noticed that `systemd-journald` does not work properly by using a Garden Linux image build with the `readonly` feature. After some debugging, I've noticed that the `machine-id` is not set properly if one uses the `readonly` feature.

It seems that the `machine-id` is stuck in the initialization, so that `/etc/machine-id` returns `uninitialized` way after the image has been booted.

I assume it has something to do with `systemd-machine-id-commit.service` which returns the following information:
```
root@localhost:/etc# systemctl status systemd-machine-id-commit.service
○ systemd-machine-id-commit.service - Commit a transient machine-id on disk
     Loaded: loaded (/lib/systemd/system/systemd-machine-id-commit.service; sta>
     Active: inactive (dead)
  Condition: start condition failed at Mon 2022-06-27 13:50:17 UTC; 12min ago
             └─ ConditionPathIsMountPoint=/etc/machine-id was not met
       Docs: man:systemd-machine-id-commit.service(8)
```

Once the `machine-id` is set manually, everything seems to work normally.

The `machine-id` should be set properly and `systemd-journald` should be able to show system logs.

Build Garden Linux like this and then run it.
```
./build.sh --arch amd64 --skip-build --features server,cloud,kvm,_dev,_readonly .build
```
Once started, you shouldn't be able to see logs via `journalctl`

chrinorse:

/kind bug
/area os
/os garden-linux

set machine-id properly on read only filesystem
Fixes #1053 

**Release note**:

```feature user

```

### DIFF
--- a/bin/makepart
+++ b/bin/makepart
@@ -100,7 +100,7 @@ sed 's/#.*//;/^[[:blank:]]*$/d' \
 			echo "$source" "$target" "$fs" "$options" "$dump" "$pass" >> "$fstab"
 			rm -rf "$rootfs_work/overlay"
 		else
-			echo "overlay /var overlay defaults,lowerdir=/var,upperdir=/overlay/var/upper,workdir=/overlay/var/work,x-systemd.requires-mounts-for=/overlay 0 2" >> "$fstab"
+			echo "overlay /var overlay defaults,lowerdir=/sysroot/var,upperdir=/sysroot/overlay/var/upper,workdir=/sysroot/overlay/var/work,x-systemd.requires-mounts-for=/sysroot/overlay,x-initrd.mount 0 2" >> "$fstab"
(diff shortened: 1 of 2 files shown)


bin/makepart
```diff
@@ -101,7 +101,7 @@ sed 's/#.*//;/^[[:blank:]]*$/d' \
 			rm -rf "$rootfs_work/overlay"
 		else
 			echo "overlay /var overlay defaults,lowerdir=/var,upperdir=/overlay/var/upper,workdir=/overlay/var/work,x-systemd.requires-mounts-for=/overlay 0 2" >> "$fstab"
```
nanory, line 103: What about `/var`? Although we hadn't issues with something like the machine-id there, it may make sense to mount this directory as soon as possible, too. This way we keep the time for a read only /var short.

nanory (changes requested):

nanory (approved):

Issue #1102, opened 2022-07-07
The tiger tests let's other feature test fail


nanory:
**What happened:**
The `tiger` test let's other feature tests, that are executed afterwards fail. Because of this, the `tiger` test has been disabled temporarily in #1040. However, it must be analyzed and fixed, so that we can re-add `tiger` again.

**Debug Logs:**
Once the tiger test has been executed, commands like `timedatectl show` freezes and are canceled by `paramiko` resulting in an `exit_code=255` as one can see here:
```
______________ test_umask[/etc/profile.d/50-autologout.sh-dict0] _______________

client = <helper.sshclient.RemoteClient object at 0xffff8e5339d0>
file = '/etc/profile.d/50-autologout.sh'
dict = {'TMOUT': '600', 'export': 'TMOUT', 'readonly': 'TMOUT'}

    @pytest.mark.parametrize(
        "file,dict",
        [
            ("/etc/profile.d/50-autologout.sh", {
                "TMOUT": "600",
                "readonly": "TMOUT",
                "export": "TMOUT"
            })
        ]
    )
    
    
    # Run the test unit to perform the
    # final tests by the given artifact.
    def test_umask(client, file, dict):
>       file_content(client, file, dict, quiet=False)

client     = <helper.sshclient.RemoteClient object at 0xffff8e5339d0>
dict       = {'TMOUT': '600', 'export': 'TMOUT', 'readonly': 'TMOUT'}
file       = '/etc/profile.d/50-autologout.sh'

../features/cloud/test/test_autologout.py:24: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

client = <helper.sshclient.RemoteClient object at 0xffff8e5339d0>
fname = '/etc/profile.d/50-autologout.sh'
args = {'TMOUT': '600', 'export': 'TMOUT', 'readonly': 'TMOUT'}, invert = False
ignore_missing = False, only_line_match = False, ignore_comments = False
quiet = False

    def file_content(client, fname, args, invert=False, ignore_missing=False, only_line_match=False, ignore_comments=False, quiet=True):
        """ Performing unit test to find key/val in files """
        content = _get_content_remote_file(client, fname, quiet)
    
        if not content:
>           assert not content and ignore_missing, f"Content or file not found: {fname}."
[...]

PR #1103, opened 2022-07-07
feature: Make KVM work with Garden Linux ARM

### LINKED ISSUE #1040: ARM64: PyTest -> Integration Tests -> `KVM`
Meta Issue: #1028 

## PyTest
This issue covers the validation of `integration tests` of PyTest for the platform provider `KVM`  running on `ARM64` / `AARCH64` architectures.

## Tests
Within this issue the following components will be validated and tested:
- PyTest
  - Integration Tests
    - `KVM`

nanory:
/kind bug
/area os
/os garden-linux

This PR fixes the KVM platform and some tests which failed on the `kvm` Garden Linux image using the `kvm` pytests.
In addition, the `kvm` platform has been modified, so that it uses the `bin/start-vm` script to spawn a KVM machine instead of using its own Qemu command. However, the `start-vm` script has been adjusted as well in order to work with the `kvm` platform by introducing the new CLI parameter `--destport`.

Fixes #1040 

The `tiger` tests has been disabled by this PR since it causes issues with other tests that are executed afterwards. In order to fix `tiger`, the following issue has been created: #1102


bin/start-vm
```diff
@@ -449,7 +453,7 @@ printf "  starting VM(UUID:%s) with MAC:%s in %s\n" $uuid $macfull $targetDir
 [ $pxe ]	&& printf "  pxeboot: %s.ipxe\n" $mac
 [ $pxe ]	&& printf "  pxeboot: files served from %s\n" "$pxe"
 [ $bridge ]	&& printf "  interface: %s  bridged\n" $bridge
-[ $bridge ]	|| printf "  sshport: %s  unbridged\n" $port
+[ $bridge ]	|| printf "  sshport: %s (%s) unbridged\n" $port $port_dest
```
gyptazy, line 456: we should shortly describe the differences within the output, while only `support: 2223 (22) unbridged` could be a little bit confusing. Maybe:

`printf " sshport: VM tcp/%s mapped to host tcp/%s unbridged\n" $port_dest $port`
or
`printf " sshport: local %s  -> vm %s (unbridged)\n" $port $port_dest`

tests/integration/kvm.py
```diff
@@ -176,112 +178,102 @@ def _adjust_kvm(self):
...
+                image_name=image_name,
+                authorized_keys_file=authorized_keys_file,
+                sshd_systemd_file=sshd_systemd_file,
+                sshd_config_file=sshd_config_file
+            )
+        )
+        cmds.append(copy_cmd)
+
+        # Modify the snapshot via guestfish
+        guestfish_cmd = (
```
gyptazy, line 209: Probably, this is a good time to get rid of this old format of `"{foo}".format(foo=bar)` and switch to `f"{bar}"`

gyptazy (changes requested):

build.sh
```diff
@@ -184,7 +184,7 @@ else
 		fullfeatures="$(cat $outputDir/fullfeature.info)"
 		configDir=$(${thisDir}/bin/garden-integration-test-config kvm ${prefix} ${fullfeatures} ${outputDir} ${arch})
 		echo "Running pytests in KVM"
-		${gardenlinux_build_cre} run --name $containerName --rm -v /boot/:/boot \
+		${gardenlinux_build_cre} run --name $containerName --privileged --rm -v /boot/:/boot \
```
nkraetzschmar, line 187: why do we need `--privileged` now?

nkraetzschmar: shouldn't `--device=/dev/kvm` be sufficient?

chrinorse:
Please do not remove the tiger test, the test works as expected in the chroot test environment and should only be excluded from other test environments like kvm.

nanory:
> Please do not remove the tiger test, the test works as expected in the chroot test environment and should only be excluded from other test environments like kvm.
I re-added the tiger test since it is only executed on `chroot` from now on, so it shouldn't cause any further problems.

gyptazy (approved):
technically lgtm

While we could change all `"{foo}".format(foo=bar)` to `f"{bar}"` here for `kvm.py`, I'm also fine with creating a new issue to proceed on this because this could be out of scope here. What do you think?

chrinorse (approved):
lgtm

PR #1104, opened 2022-07-07
feature(start-vm): Refactor `start-vm` script

### LINKED ISSUE #1062: Refactor: `startvm` script
The whole `startvm` script should be completley refactored.

**Why is this needed**:
While `startvm` was initially only used on Debian with less options, `startvm` now supports more operating systems, more options and multiple architectures. This makes it hard to maintain with several different options, `if` statements, not using common scripts like `get_arch.sh` and so on. Next to it, it doesn't follow any style guide(s). While this is currently technically ok, we should take care for this with low priority.

gyptazy:
feature(start-vm): Refactor `start-vm` script

/kind cleanup
/area os
/os garden-linux

Refactor of `start-vm` to have shell check valid code base, uniformed var names, same style, etc.

Fixes #1062

**Release note**:

```feature user
NONE
```


chrinorse (approved):
lgtm

bin/start-vm
```diff
@@ -1,523 +1,728 @@
...
+        if [ "amd64" = $(uname -m) ]; then
+            test_arch="amd64"
+        fi
+
+        if [ "x86_64" = $(uname -m) ]; then
+            test_arch="amd64"
+        fi
+
+        # Validate for native or emulated arch
+        if [ "$arch" = "$test_arch" ]; then
```
nanory, line 493: By using `get_arch.sh` here, you don't need to define a temp variable like `test_arch` and could simply compare `$arch` with the result of `get_arch.sh`

bin/start-vm
```diff
@@ -1,523 +1,728 @@
...
+    QEMU_OPTS+=("-device virtio-rng-pci,rng=rng0"
+                "-object rng-random,id=rng0,filename=/dev/random")
 fi
 
-# Start QEMU with virtOpts
-if [ "centos" = $running_os ]; then
-        /usr/libexec/qemu-kvm ${virtOpts[@]}
+# Set QEMU start options depending on
+# daemonizing the VM
+if [ $daemonize ]; then
```
nanory, line 680: Indention is missing

bin/start-vm
```diff
@@ -1,523 +1,728 @@
...
+    # (currently unsupported on macOS)
+    if [ $os != macos ]; then
+        if [ $bridge ]; then
+            [ $(id -u) == 0 ] || eusage "Error: For bridging you must be root. More information can be found on: https://github.com/gardenlinux/gardenlinux/tree/main/bin#start-vm"
+            if [ ! -d /etc/qemu ]; then
+                mkdir -p /etc/qemu
+                chown root:kvm /etc/qemu
+            fi
+            if [ ! -e /etc/qemu/bridge.conf ]; then
+                    touch /etc/qemu/bridge.conf
```
nanory, line 637: Wrong indention

nanory (changes requested):

bin/start-vm
gyptazy: This wouldn't offer legacy support when someone defines `--arch` for emulating as `x86_64` or `aarch64` and would only work if the user types `amd64` or `arm64` which is uncommon. Therefore, we need to validate this before which is stored in a var called `test_arch`.

nanory: You already normalize the `arch` in `set_binaries`, so `arch` can only be `arm64` or `amd64` at this point. I would also suggest to create a new function called `normalize_arch` which takes the if statement of `set_binaries` and executes them right after `get_opts` (or make it a part of `get_opts`)

bin/start-vm
```diff
@@ -1,523 +1,714 @@
...
-        gnu_stat="gstat"
-
-        # Translate ARM64 to AARCH64
-        # on macOS
-        if [ "arm64" = $arch ]; then
-            arch="aarch64"
+
+# Constants
+readonly CURR_DIR="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+readonly TARGET_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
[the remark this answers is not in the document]
gyptazy: Good catch! TBH, I just remember that I ran into an issue related to this but I don't see any reasons for this anymore. If replaced `TARGET_DIR` with `CURR_DIR` and revalidated this without any issues.

Vincinator (approved):
also looks good for me

Issue #876, opened 2022-05-10
User build modules under secure boot


nkraetzschmar:
If during boot the kernel detects that it is booted with secure boot enabled, it automatically goes into lockdown mode.
In lockdown mode loading modules requires their signatures to be verified against a trusted certificate.
Trusted certificates are compiled into the kernel or loaded from the UEFI secure boot signature db.

This raises the question: how to allow users to load custom build kernel modules on a secure booted kernel?
The easy option would be to use user generated secureboot keys.
However, this requires the user to build their own gardenlinux images, rather than using images build by us.

Can we allow loading user build modules, while using pre-build images with the official gardenlinux secure boot keys enrolled?
One option for this might be machine owner keys.

## Machine Owner Keys (MOK)

At boot time, in addition to the UEFI db keys, the kernel also loads the mok keys from the EFI variable `MokListRT-605dab50-e046-4300-abb6-3dd810dd8b23` into the `.platform` keyring ([security/integrity/platform_certs/load_uefi.c](https://github.com/torvalds/linux/blob/v5.15/security/integrity/platform_certs/load_uefi.c)).
Depending on kernel configuration, these keys may also be trusted for module signature verification.

Important to note here is that debian applies a [patch](https://salsa.debian.org/kernel-team/linux/-/blob/bde70848a22c19fbb8c586aafd1c378cb83cffc7/debian/patches/features/all/db-mok-keyring/0004-MODSIGN-check-the-attributes-of-db-and-mok.patch) to this function which makes sure the kernel only loads mok keys from runtime EFI variables (i.e. variables without the `EFI_VARIABLE_NON_VOLATILE` bit set).
This requires the boot loader to mirror the non-volatile boot time access only EFI variable `MokList` into the volatile runtime variable `MokListRT`.
This is done automatically when using the shim bootloader.
The problem with using shim is, that it is strongly coupled with grub and does not play nicely with unified UEFI images.

Simply disabling the debian patch would work but opens up the system to a quite severe security problem:
Without this patch a mallicious root user could create a non-volatile `MokListRT` variable with any key they control and after the next reboot this key would be trusted to load modules; essentially bypassing all the security of lockdown mode.

## Potential solution

A potential solution would be to extend the systemd EFI stub used in our combined EFI image to also perform the mirroring of boot time variable `MokList` into volatile runtime var `MokListRT`.
[...]

PR #1109, opened 2022-07-08
feature(secureboot): Extend _secureboot feature with additional tools and config

### LINKED ISSUE #1107: Move `enroll-gardenlinux-secureboot-keys` to `secureboot` feature
We should move `enroll-gardenlinux-secureboot-keys` to `secureboot` feature, instead of placing it from different features. This ensures that only the `secureboot` needs to be included. (Rel. #1106)

nanory:

/kind enhancement
/area os
/os garden-linux

This PR moves scripts, configuration and tools needed to enable secure boot for Garden Linux to the `_secureboot` feature

Fixes #1107


features/cloud/info.yaml
```diff
@@ -3,6 +3,7 @@ type: element
 features:
   include:
     - server
+    - _secureboot
```
nkraetzschmar, line 6: secureboot should not be enabled by default as it removes legacy boot support and prevents loading user build kernel modules, or at least makes this significantly more challenging (see #876)
&xrArr; should not be included from cloud feature

features/metal/info.yaml
```diff
@@ -3,6 +3,7 @@ type: platform
 features:
   include: 
     - server
+    - _secureboot
```
nkraetzschmar, line 6: as for cloud secureboot probably also shouldn't be included by default for metal either

features/_secureboot/file.exclude
```diff
@@ -1 +1 @@
-/boot/efi/*
+/boot/efi/syslinux/*
```
nkraetzschmar, line 1: why limit this to syslinux files only? the idea behind the secure boot feature is to *only* support secure boot (to reduce image size by avoiding duplicated data for kernel and initrd as both standalone and inside the combined efi image) &rArr; only the combined efi image should be needed on the efi system partion, which gets build later on (`makepart` / `makesecureboot`), so all other files can safely be cleared from the efi partition entirely

nkraetzschmar (changes requested):

features/cloud/info.yaml
nanory: Ok, removed it.

features/metal/info.yaml
nanory: Removed it as well

features/_secureboot/file.exclude
nanory: I reverted this change back to `/boot/efi/*` but I had to adjust `makepart` since it tried to touch two files in the `/boot/efi/loader` directory which didn't exists because of the `file.exclude`.
* https://github.com/gardenlinux/gardenlinux/blob/feature/1107_add_secureboot_feature/bin/makepart#L28-L30

PR #1110, opened 2022-07-08
run debsums check only in chroot integration test environment

### LINKED ISSUE #1101: no ```apt-get update``` when fedramp is enabled
**What happened:**
In the fedramp feature *fips* is enabled in the kernel, as a result ```apt-get update``` is not working anymore.

**Debug Logs:**
```
root@localhost:~# apt-get update
Hit:1 https://repo.gardenlinux.io/gardenlinux today InRelease                                                                                                    
Hit:2 https://deb.debian.org/debian bookworm InRelease                                                                                                           
Hit:3 https://deb.debian.org/debian bookworm-updates InRelease
Hit:4 https://security.debian.org/debian-security bookworm-security InRelease
fatal error in libgcrypt, file ../../src/misc.c, line 92, function _gcry_fatal_error: requested algo not in md context

Fatal error: requested algo not in md context
Aborted (core dumped)
```
**How to reproduce it**:
Run ```apt-get update``` when the fedramp feature is enabled.
 
**Todo**
Add a fixture to the pytest integration tests to skip all tests that run ```apt-get update``` when the feature fedramp is enabled. Disabling *fips* is not an option since it is needed by fedramp.

chrinorse:

/kind bug
/area os
/os garden-linux

don't run debsums in an environment that prevents to run apt-get update and install additional packages.
Fixes #1101 

**Release note**:

```feature user

```

### DIFF
--- a/features/base/test/test_debsums.py
+++ b/features/base/test/test_debsums.py
@@ -12,5 +12,5 @@ import pytest
     ]
 )
 
-def test_debsums(client, exclude):
+def test_debsums(client, exclude chroot):
     debsums(client, exclude)
--- a/features/base/test/test_debsums.py
+++ b/features/base/test/test_debsums.py
@@ -12,5 +12,5 @@ import pytest
     ]
 )
 
-def test_debsums(client, exclude chroot):
+def test_debsums(client, exclude, chroot):
     debsums(client, exclude)


features/base/test/test_debsums.py
```diff
@@ -12,5 +12,5 @@
     ]
 )
 
-def test_debsums(client, exclude):
+def test_debsums(client, exclude, chroot):
```
gyptazy, line 15: Instead of pinning this only to `chroot`, we probably should create a new fixture (e.g. `non_fedramp` that only prevents it from running when `fedramp` feature is activated.

gyptazy (changes requested):

features/base/test/test_debsums.py
chrinorse: no, this test should really only run on chroot. Since /usr is read-only and it will not be possible to install anything except in the chroot environment.

gyptazy: Ok, I'm fine with it in this case (just struggled because of the branch name).

lgtm

gyptazy (approved):

PR #1113, opened 2022-07-11
fix(test): Make packages_musthave work with pkg.exclude properly

### LINKED ISSUE #1112: packages_musthave test not working properly with excludes
**What happened:**
During tests of #1107, I noticed that the `packages_musthave` test fails although all packages were installed. It seems that the test does not work properly with the `pkg.exclude` functionality. If a feature requires a package, that another feature is excluding, the tests fails and creates a false positive.

**How to reproduce it**:
Create a `kvm-dev` image with `_secureboot` enabled.
 
**Environment:**
| Name | Value |
| ------------------- | ----- |
| Garden Linux Branch | main |
| Container runtime (version)  | podman (3.0.1) | 
| Host OS of build machine  | Debian 11 | 

The `package_musthave` test already contains a routine to handle `pkg.exclude` but there seems to be a bug in it.

nanory:

/kind bug
/area os
/os garden-linux

This PR resolves an issue with the `packages_musthave` test which did not work properly with `pkg.exclude` files within affected features.

Fixes #1112

### DIFF
--- a/tests/helper/tests/packages_musthave.py
+++ b/tests/helper/tests/packages_musthave.py
@@ -6,7 +6,7 @@ import pytest
 
 def packages_musthave(client, testconfig):
     """"Test if the packages defined in pkg.include are installed"""
-    installed_pkgslist = get_package_list(client)
+    installed_package_list = get_package_list(client)
 
     current = (os.getenv('PYTEST_CURRENT_TEST')).split('/')[0]
     path = f"/gardenlinux/features/{current}/pkg.include"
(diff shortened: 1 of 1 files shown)


tests/helper/tests/packages_musthave.py
```diff
@@ -54,8 +54,9 @@ def packages_musthave(client, testconfig):
         # * the architecture as a variable in the package name
         elif package.endswith(r"-${arch}"):
             package = package.replace(r"${arch}", arch)
+
         # explicitly excluded packages are allowed to miss 
-        elif package in exclude:
+        if package in exclude:
             continue
```
gyptazy, line 60: I'm not quite sure if we should print an info or warning for the pytest output (e.g. `logger.info(f"Package {package} is excluded. Continuing")`). However, lgtm

gyptazy (approved):

PR #1126, opened 2022-07-21
fix: skip dev builds in github actions

nkraetzschmar:
we probably don't need to build `_dev` flavours all the time.
this should slightly reduce build times and significantly reduce aws costs for autoscaling runners.


MrBatschner (approved):
Automatically building dev-images is not reasonable and produces lots of unnecessary waste. Dev-images are meant to debug Garden Linux images and if that became necessary, one could always check out the offending commit and build a dev-image manually.

Therefore...
/lgtm
but
/needs second-opinion

PR #1151, opened 2022-08-02
Feature/460 gardenlinux build image is not updated on every release day

### LINKED ISSUE #460: gardenlinux/build-image is not updated on very release day
gardenlinux/build-image in docker / containerd repository should be updated at least once a day so we have at least the daily apt-get update

currently that is broken:

```
REPOSITORY                       TAG            IMAGE ID       CREATED        SIZE
gardenlinux/build-image          575.0          751d223ff52a   2 days ago     497MB
gardenlinux/build-image          577.0          751d223ff52a   2 days ago     497MB
```

both imageid's are identical

chrinorse:

/kind bug
/area os
/os garden-linux

Check if the major.minor build-image version exists, if not delete the build-image and rebuild it.

Fixes #460

**Release note**:

```feature user

```

### DIFF
--- a/container/Makefile
+++ b/container/Makefile
@@ -17,7 +17,10 @@ needslim:
 .PHONY: build-image
 build-image: needslim
 	cp ../bin/garden-feat.go build-image/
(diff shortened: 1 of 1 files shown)


Vincinator (commented):
Thank you! Looks good.

@gehoern will rework the `bin/garden-version`, 
so that garden-version will not output `today`, but the actual version of that day in the future.

So the tags `VERSION` and `VERSION_NUMBER` might be the same after @gehoern's changes.

Vincinator (approved):
/lgtm  

will eventually need to adapt that one line to only use `VERSION`

PR #1161, opened 2022-08-10
fix: move login.defs umask to base feature

### LINKED ISSUE #39: Set a strict global umask value
umask value of `077` or minimum `027` 

**Why is this needed**:
The permissive value of `022` should not be used if possible.

Vincinator:

- Move setting of login.defs umask 027 to base feature 
    - including test_umask

Fixes #39 

**Notes**
- Noticed during testing: when creating a user with adduser, files from `/etc/skel/*` are copied to the users new home directory. Permission of these files will not respect the umask value of login.defs.

### DIFF
--- a/features/base/exec.config
+++ b/features/base/exec.config
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
+# set default umask to a more conservative value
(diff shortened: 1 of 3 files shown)


gyptazy (approved):
lgtm

PR #1165, opened 2022-08-11
check if .git directory is present

### LINKED ISSUE #1149: Garden Linux cannot be built without `.git` directory
**What happened:**

Building Garden Linux with its source outside of a git repository fails. This makes downloading a .ZIP from GitHub and building from it impossible.

The error that eventually is encountered is this:
```
make[1]: Leaving directory '/home/mrbatschner/gl-test/gardenlinux-main/bin'
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
make: *** [Makefile:66: aws] Error 128
```

This is very likely caused (but not limited to) by the several calls to git in `build.sh` to determine the git commit-hash that should be placed inside of `/etc/os-release`.

**How to reproduce it**:
- Check out this repository either per git and `rm -rf .git` or
- download this repository as .zip file
- `make aws`
 
**Environment:**

| Name | Value |
| ------------------- | ----- |
| Garden Linux Branch | *main* |
| Container runtime (version)  | _irrelevant_ | 
| Host OS of build machine  | Linux | 

- tested with `main` at 99c8db0c6038ace3ad2bf6f443915f0d5d4bcb8c

chrinorse:

/kind bug
/area os
/os garden-linux

check if .git directory is present
Fixes #1149 

**Release note**:

```feature user

```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -131,7 +131,7 @@ if [ -n "$cert" ]; then
 	dockerArgs+=" --volume $(realpath "$cert"):/cert:ro"
 fi
 
-if [ -z "$(git status --porcelain)" ]; then
+if [ -d .git ] && [ -z "$(git status --porcelain)" ]; then
 	commitid_long="$(git rev-parse HEAD)"
 	commitid="$(git rev-parse --short HEAD)"
 	echo "clean working tree, using $commitid as commit id"


gyptazy (approved):
lgtm

PR #1169, opened 2022-08-15
fix(SELinux): Reenable `setfiles` for `SELinux` file contexts

### LINKED ISSUE #1014: Unable to build Garden Linux: /sbin/setfiles: Could not set context for /:  No such file or directory
**What happened**:
Currently, we're unable to build Garden Linux which fails in one of the last steps... 

```
#### building diskimage
  found new fstab in base
    modifying fstab with cloud
  building rawfile
/sbin/setfiles: Could not set context for /:  No such file or directory
/sbin/setfiles: Could not set context for /overlay:  No such file or directory
/sbin/setfiles: Could not set context for /overlay/var:  No such file or directory
/sbin/setfiles: Could not set context for /overlay/var/work:  No such file or directory
[...]
```

Fetch a fresh GIT copy and build with `make kvm-dev`.

Today, `libselinux1` got updated to upstream version `3.4-1` in Debian Testing (see also: https://metadata.ftp-master.debian.org/changelogs//main/libs/libselinux/libselinux_3.4-1_changelog)

gyptazy:
fix(SELinux): Reenable `setfiles` for `SELinux` file contexts

/kind bug
/area os
/os garden-linux

Reenable `setfiles` for `SELinux` file contexts

Fixes #1014

None

**Release note**:

```feature user
None
```

**Test(s)**:
```
root@localhost:~# sestatus -v
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             default
Current mode:                   permissive
Mode from config file:          permissive
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     actual (secure)
Max kernel policy version:      33

Process contexts:
Current context:                unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
Init context:                   system_u:system_r:init_t:s0
/usr/sbin/sshd                  system_u:system_r:sshd_t:s0

File contexts:
Controlling terminal:           unconfined_u:object_r:user_tty_device_t:s0
/etc/passwd                     system_u:object_r:etc_t:s0
/etc/shadow                     system_u:object_r:shadow_t:s0
/bin/bash                       system_u:object_r:shell_exec_t:s0
/bin/login                      system_u:object_r:login_exec_t:s0
/bin/sh                         system_u:object_r:bin_t:s0 -> system_u:object_r:shell_exec_t:s0
/sbin/agetty                    system_u:object_r:getty_exec_t:s0
/sbin/init                      system_u:object_r:bin_t:s0 -> system_u:object_r:init_exec_t:s0
/usr/sbin/sshd                  system_u:object_r:sshd_exec_t:s0


root@localhost:~# ls -alZ /bin/bash
-rwxr-xr-x. 1 root root system_u:object_r:shell_exec_t:s0 1230360 Aug 15 00:00 /bin/bash
```

### DIFF
--- a/bin/makepart
+++ b/bin/makepart
@@ -6,6 +6,7 @@ set -Eeuo pipefail
 exec 3>&1
 exec 1>&2
 
+thisDir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 rootfs="$1"
 arch="${2:-$arch}"
 timestamp=$(garden-version --epoch "$version")
@@ -16,10 +17,9 @@ mkdir -p "$rootfs_work/overlay"/{etc,var}/{upper,work}
 
 find "$rootfs_work/var/log/" -type f -delete
 
+# SELinux 'default' policy file contexts
 chcon -R system_u:object_r:unlabeled_t:s0 "$rootfs_work"
-# Temporary deactivated until this is fixed by upstream:
-# GitHub: https://github.com/gardenlinux/gardenlinux/issues/1014
-#chroot "$rootfs_work" /usr/bin/env -i /sbin/setfiles /etc/selinux/default/contexts/files/file_contexts /
+"$thisDir/garden-chroot" "$rootfs_work" /sbin/setfiles /etc/selinux/default/contexts/files/file_contexts /
 rm "$rootfs_work/.autorelabel"
 
 if ! grep _secureboot <<<"$features" > /dev/null; then
--- a/features/_pxe/image
+++ b/features/_pxe/image
@@ -2,6 +2,7 @@
 
 set -Eeuxo pipefail
 
+thisDir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 rootfs=$1
 targetBase=$2
 targetBaseDir=$(dirname "$targetBase")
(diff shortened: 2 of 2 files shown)


bin/makepart
```diff
@@ -16,10 +16,9 @@ mkdir -p "$rootfs_work/overlay"/{etc,var}/{upper,work}
 
 find "$rootfs_work/var/log/" -type f -delete
 
+# SELinux 'default' policy file contexts
 chcon -R system_u:object_r:unlabeled_t:s0 "$rootfs_work"
-# Temporary deactivated until this is fixed by upstream:
-# GitHub: https://github.com/gardenlinux/gardenlinux/issues/1014
-#chroot "$rootfs_work" /usr/bin/env -i /sbin/setfiles /etc/selinux/default/contexts/files/file_contexts /
+/opt/gardenlinux/bin/garden-chroot "$rootfs_work" /sbin/setfiles /etc/selinux/default/contexts/files/file_contexts /
```
Vincinator, line 21: A small design decision may be hiding here. I think this is fine, but want to hear your perspective before approving. A short reply is also fine for me.

Question: Can we safely assume that `/opt/gardenlinux/bin/` programs exist when `bin/makepart` is called?

In theory, it would sound nice to not introduce a hard coded path to its dependencies in /opt/gardenlinux.
 
However, this discussion would be bikeshedding - so I lean towards accepting that assumption. And just fix the path when we need to.

As long as [build.sh](https://github.com/gardenlinux/gardenlinux/blob/3f5bc0a235285ccbb3add8980730e1a29dbd5fff/build.sh#L121) does it that way, we should be fine.

nanory: I agree with @Vincinator, we should probably not hard code the path.
Why not using the variable `thisDir` like in other scripts within the `bin/` directory? This should be sufficient, no?
```
thisDir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
```

Then, we could use the variable it like this:
```
"$thisDir/garden-chroot" "$rootfs_work" /sbin/setfiles /etc/selinux/default/contexts/files/file_contexts /
``` 

What do you think?

gyptazy: Thank you both for the input. I think _thisDir_ is perfectly fine for this (and intended for this cases).

nanory (approved):

PR #1171, opened 2022-08-16
feat(test): Move LSM tests to dedicated gardener test folder

### LINKED ISSUE #1170: Move current SELinux and AppArmor tests to gardener feature
As of now, there are two tests in the `test_gardenlinux.py` pytest that check if AppArmor is the default MAC:
* test_selinux
* test_apparmor

We should move those tests to the gardener feature (feature tests).

**Why is this needed**:
Those tests ensure that AppArmor is the default MAC in the Garden Linux image. However, only `gardener` is setting AppArmor as the default. All other features still use SELinux. Therefore, we should move those tests to the gardener feature in order to mitigate false positives.

gyptazy:
feat(test): Move LSM tests to dedicated gardener test folder

/kind test
/area os
/os garden-linux

Move LSM tests to dedicated gardener test folder and use PyTest syntax.

Fixes #1170

**Release note**:

```feature user
NONE
```

**Test(s)**:
```
PASSED gardener/test/test_lsm.py::test_lsm[selinux-False]
PASSED gardener/test/test_lsm.py::test_lsm[apparmor-True]
```

### DIFF
--- a/features/gardener/test/test_lsm.py
+++ b/features/gardener/test/test_lsm.py
@@ -1,0 +1,20 @@
+import pytest
+from helper.utils import execute_remote_command
+
+
+@pytest.mark.parametrize(
+    "lsm,active",
+    [
+        ("selinux", False),
+        ("apparmor", True),
+    ]
+)
+
+
+def test_lsm(client, lsm, active, non_chroot):
+    out = execute_remote_command(client, "cat /sys/kernel/security/lsm")
+
+    if active:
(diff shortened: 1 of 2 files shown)


nanory (approved):

nanory:
Change lgtm @gyptazy 
What do you think about this change @MrBatschner?

Vincinator (approved):

PR #1172, opened 2022-08-16
feat:  kernel module build container

Vincinator:
Adds the kernel module build container.
This is required to include build container for the upcoming releases.

- Added tests that check if libdb is installed
    - test is executed for target `test` in the docker multi-stage setup of `container/kernelmodule-builder/Dockerfile`
    - building via makefile automatically executes tests
- Uses `garden-version` to determine garden linux repo suite to use
- Added `garden-version --kernel` flag to print the kernel version used in this build context
- Added `KERNEL_VERSIONS` file to specify which kernel version should be used 
    - `KERNEL_VERSION="latest"`  just installs the latest available patchlevel for the given kernel lts
    - `KERNEL_BASE=` defines which lts version to use (e.g. 5.10 or 5.15)

The libdb test will fail, the base image `debian:bookworm-slim` contains `libdb`. We need to base this kernelmodule-builder container on the garden linux base image, which is currently under development.

Build container with 
```
cd container
# for the amd64 headers  
make kernelmodule-builder-amd64
# for the arm64 headers
make kernelmodule-builder-arm64
```

To test, you could use the open nvidia kernel modules.. compile kernel module with:
```
sudo podman run -it --rm localhost/gardenlinux/kernelmodule-builder-amd64:today

git clone https://github.com/NVIDIA/open-gpu-kernel-modules.git
cd open-gpu-kernel-modules
make modules

cd kernel-open
modinfo nvidia.ko | grep vermagic
# On success, you see "gardenlinux"  in the version string. Current latest 5.15 kernel
```


Vincinator:
We need to remove libdb and add tests to check if libdb is not included in the container image.

PR #1178, opened 2022-08-17
Feature/345 add gardenlinux image flavour for hypervisor generation v2

### LINKED ISSUE #345: Add Garden Linux image flavour for hypervisor generation V2 [Azure]

As a user I want to create Garden Linux images for hypervisor generation V2. 

**Why is this needed**:

See https://docs.microsoft.com/en-us/azure/virtual-machines/generation-2

chrinorse:

/kind enhancement
/area os
/os garden-linux

add option to test configuration to choose which hyper v generation to use for testing a garden linux image on azure
Fixes #345 

**Release note**:

```feature user

```


tests/README.md
```diff
@@ -242,6 +242,8 @@ azure:
     #image_region: eu-central-1
     # already existing image in Azure to be used for testing (required/alternatively optional)
     image_name: <image name>
+    # hyperV generation to use. Possible values are 'V1' and 'V2'. Default is 'V1'
+    hyper_v_generation: V2
```
gyptazy, line 246: According to the default in `azure.py` we might set this to `V1`, too. This could avoid confusions while already mentioning the that `V1` and `V2` are accepted values.

tests/integration/azure.py
```diff
@@ -583,12 +583,17 @@ def upload_image(self, progress_function = None):
 
             image_uri = f"https://{self._storageaccount.name}.blob.core.windows.net/vhds/{image_name}.vhd"
 
+            if "hyper_v_generation" in self.config:
+                hyper_v_generation = self.config["hyper_v_generation"]
```
gyptazy, line 587: What happens with an unexpected value like `V3`? Maybe we should validate for allowed values?

gyptazy (changes requested):
Please take a short look on my inline notes.

tests/integration/azure.py
nanory: I agree with this. We should probably do something like this:
```
allowed_generations = ["V1", "V2"]
hyper_v_generation = self.config.get("hyper_v_generation", allowed_generations[0])
if hyper_v_generation not in allowed_generations:
   raise RuntimeError(f"Hypervisor generation '{hyper_v_generation}' not supported ({allowed_generations})") 
```

nanory (changes requested):

gehoern (approved):

nanory (approved):

gyptazy (approved):
lgtm

PR #1183, opened 2022-08-17
CI: publish images on Azure using shared image gallery

AndreasBurger:
Update the release manifest we generate so that we can later on determine the way the image was published (needed for gardener `cloudprofile` generation) and enable publishing via shared image gallery.


ccwienk (approved):

jensh007 (approved):
Maybe needs a rebase?

PR #1186, opened 2022-08-18
feat(_selinux): Add inital SELinux feature

### LINKED ISSUE #1184: `SELinux`: Create an additional feature `_selinux` as type `flag`
Currently, `SELinux` is sourced from many features and configured in slightly different ways. We should ensure that we have a common feature that gets included from other features that require `SELinux` and only modify it in the smallest way to fit the features requirements.

gyptazy:
feat(_selinux): Add inital SELinux feature

/kind enhancement
/area os
/os garden-linux

This PR adds `SELinux` as a part of the feature system (type `flag`). The primary goal is to have a common feature for SELinux that applies all basic stuff for Garden Linux (e.g. correct policies, correct modes, etc), that should only slightly be changed by other features (or in best case situations, no changes are needed). This could affect `CIS`, as well as `FedRAMP`.

Fixes #1184

We should discuss if we want to proceed with `enforcing` as default. Within this mode, one the one hand we may find faster and more missing policy rules, on the other one it may cause more trouble at the beginning.

Do NOT merge until `selinux-policy-targeted-fedora` is present in Garden Linux repositories. 

**Release note**:

```feature user
- category: feature
- target_group: user
```

### DIFF
--- a/features/_selinux/file.include/etc/selinux/config
+++ b/features/_selinux/file.include/etc/selinux/config
@@ -1,0 +1,14 @@
+# This file controls the state of SELinux on the system.
+# SELINUX= can take one of these three values:
+# enforcing - SELinux security policy is enforced.
+# permissive - SELinux prints warnings instead of enforcing.
+# disabled - No SELinux policy is loaded.
+SELINUX=enforcing
+# SELINUXTYPE= can take one of these two values:
+# default - equivalent to the old strict and targeted policies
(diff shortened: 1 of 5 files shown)


nanory (changes requested):

gyptazy:
Thanks for you comments. Pleas note that this PR is drafted and this isn't ready for review, yet.

PR #1216, opened 2022-08-24
Validate the /usr is read-only.

marwinski:

/kind test
/area os
/os garden-linux

Validates that the `/usr` directory is read-only.

Fixes #

The test has been moved to `features/base/test`. This means that 

```
    features:
      - "base"
```

must be included in the test configuration.

**Release note**:

```feature user

```

### DIFF
--- a/features/base/test/test_usr_ro.py
+++ b/features/base/test/test_usr_ro.py
@@ -1,0 +1,5 @@
+
+def test_usr_read_only(client, non_chroot):
+    (exit_code, output, error) = client.execute_command("/usr/bin/touch /usr/fail-test")
+    assert exit_code == 1, "Read-only file system error expected"
+    assert error.rstrip() == "/usr/bin/touch: cannot touch '/usr/fail-test': Read-only file system"


gyptazy:
We should probably avoid adding new tests in `test_gardenlinux.py` and add this within the new feature PyTest structure. This would fit in `features/base/tests/`.

MrBatschner (approved):
/lgtm

gyptazy:
Hi @marwinski,

we do not really distinguish between unit- and integration tests after switching to PyTest anymore. All tests are based on PyTest. To have a short overview for the process of testing in general (unit, integration, platform tests) you may find a handy chart here: https://github.com/gardenlinux/gardenlinux/tree/main/tests#chart 
As a result, it more or less is based on the used fixture if this is interpreted as a unit test (mostly `chroot` and `kvm`) or a platform test (`ali`, `azure`, `gcp`, ...) or gets called as an integration test where all of them may get applied if the fixtures are matching. For further information please check https://github.com/gardenlinux/gardenlinux/tree/main/tests#tests - if you're missing information feel free to raise an issue an assign this issue to me. I'll update the docs asap.

However, we still have the legacy part of `tests/integration/test_gardenlinux.py` and shouldn't add any new tests there even both are still sourced. Instead, we should use the new file structure where tests are placed in a feature's test folder (e.g. `base/test/test_mount_usr.py`). Next to this, I think this could become a more common test where different filesystems or mounts could get check in the near future. Therefore, we could make use of PyTest's parametrize feature and write something like this:

`base/test/test_mounts_read_only.py`
```
# Parametrize the test
@pytest.mark.parametrize(
     "mount, read_only",
     [
         ("/usr", True),
         ("/tmp", False)
     ]
 )

# Run the integration test
# Validate that mounts are readonly or not
def test_mounts_read_only(client, mount, read_only, non_chroot, non_kvm):
     do_magic_here(client, mount, read_only)
```

Maybe the movement from `def test_apparmor(client, non_chroot):` from the legacy one to the new layout could be helpful from this commit: https://github.com/gardenlinux/gardenlinux/commit/bcd2aa16f59c95cfd215f799b402f8759e523b4c

I hope this helps - feel free to task if anything is still unclear. I think we should shortly move the legacy content to the new style. Therefore, I just created #1222. We may discuss this in our today's weekly meeting.

Thanks,
gyptazy

marwinski:
pushed new version and moved test to `features/base/test`

gyptazy (approved):
lgtm

PR #1220, opened 2022-08-24
Change the way package versions are represented in the component-descriptor

AndreasBurger:
Instead of putting the package versions into the CD as part of the rather generic `build-metadata`-label, put them in their own label in a better-parseable format.

As an added benefit, this new label will be evaluated to auto-assign versions when our image-scan is unable to determine the correct version for a package that is part of the component-descriptor.


MrBatschner (approved):
/lgtm

PR #1217, opened 2022-08-24
Enable Azure accelerated networking

marwinski:
**How to categorize this PR?**

/kind enhancement
/area os
/os garden-linux

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #1198

**Special notes for your reviewer**:

The Azure feature is described here: https://docs.microsoft.com/en-us/azure/virtual-network/accelerated-networking-overview

I have checked the configuration on the VM to ensure it is enabled and working. I will add an integration test for this later on.

**Release note**:

```feature user
Garden Linux supports Azure accelerated networking.
```

PR #1228, opened 2022-08-26
test: make Azure vm_size selectable

MrBatschner:

/kind test
/area os
/os garden-linux

Platform tests on Azure should have a selectable `vm_size` and not one that is hardcoded to `Standard_D4_v4`.

**Release note**:

```feature user
test: make Azure vm_size selectable
```


gyptazy (approved):
lgtm

marwinski (commented):
this appears to be in conflict with #1217

MrBatschner:
> this appears to be in conflict with #1217

True, we both seem to have been working at similar stuff at the same time. I took the liberty and make some few changes to your code, I prefer to have parameter validation in the `validate_config()` method and not in the `create_vm()` method. 

Also, I changed the VM size to `Standard_D4_v4` as I do not believe that burstable machine types will give us reliable startup times.

PR #1229, opened 2022-08-26
feat: github actions: arm64 aws platform tests

nkraetzschmar:
**How to categorize this PR?**

/kind enhancement
/area os
/os garden-linux

**What this PR does / why we need it**:

enable aws platform tests for arm64

PR #1231, opened 2022-08-30
fix: github actions: gcp ignore --arch arguments

nkraetzschmar:

/kind bug
/area os
/os garden-linux

adding `--arch` argument to github action platform tests in #1229 broke gcp tests, this fixes the issue by using the last instead of the first argument to `gcp_tests.sh`


MrBatschner (approved):
/lgtm

Issue #1173, opened 2022-08-17
`Unit Tests`: Each `feature` should be represented by an unit test


gyptazy:
**What would you like to be added**:
Each feature that is available in `main` should be represented by a corresponding unit test to make sure it works as desired. This should include some basic tests like validating that expected packages are installed or some options have been set.

I just encountered this by the broken `vhost` feature that was recently fixed (but was still outdated in my PR) (https://github.com/gardenlinux/gardenlinux/commit/cf190c4882d78668c4281cd8815299ebd8ea944f#diff-58d034a6414ddf23c3fca49065cf98a7617721923b37b501b8863f202eec3159). However, we should ensure that each feature is represented by an unit test.

**Why is this needed**:
We should make sure that available features are really usable.

gyptazy:
I just took a look around our features that might need additional unit tests... However, many of them already provide unit tests.

No further unit tests are needed for:

**Covered**:
-  [x] _dev
-  [x] _iso
-  [x] _pxe
-  [x] ali
-  [x] aws
-  [x] azure
-  [x] base
-  [x] chost
-  [x] cis (including all cis_ sub features)
-  [x] cloud
-  [x] fedramp
-  [x] gardener
-  [x] gcp
-  [x] github_action_runner
-  [x] metal
-  [x] server

However, there're still some that provide basic ones but should be extended.

**To extend**:
- [ ] chost
- [ ] example (with docs)
- [ ] fedramp
- [ ] khost
- [ ] openstack
- [ ] vhost
- [ ] vmware

And at least we have the ones that're fully lacking of unit tests:

**Open**:
- [ ] _readonly
- [ ] _secureboot
- [ ] _slim
- [ ] aide
- [ ] clamav
- [ ] firecracker
- [ ] firewall
- [ ] oci

gyptazy:
tests have been implemented for all currently used features.
=> closed

PR #1238, opened 2022-08-31
tests(clamav): Add unit tests for feature `clamav`

gyptazy:
Add unit tests for feature `clamav`

/kind test
/area os
/os garden-linux

Add unit tests for feature `clamav`

Fixes #1173

**Release note**:

```feature user
NONE
```

**Failed**:
```
FAILED clamav/test/test_clamav_cron.py::test_clamav_cron - AssertionError: no error='cat: /var/spool/cron/crontabs/root: No such file or directory\n' expected
FAILED clamav/test/test_packages_musthave.py::test_packages_musthave - AssertionError: clamav should be installed, but are missing
```

**Passed**:
```
PASSED clamav/test/test_clamav_cron.py::test_clamav_cron
PASSED clamav/test/test_packages_musthave.py::test_packages_musthave
```

### DIFF
--- a/features/clamav/test/test_clamav_cron.py
+++ b/features/clamav/test/test_clamav_cron.py
@@ -1,0 +1,9 @@
+from helper.utils import execute_remote_command
+
+
+def test_clamav_cron(client):
+    # This will already fail if file is absent
+    rc, out = execute_remote_command(client, "cat /var/spool/cron/crontabs/root", skip_error=True)
+    # Validate that clamscan gets called (cron times may vary)
+    err_msg = "Cron configuration for ClamAV expected but could not be verified."
+    assert "clamscan" in out and rc == 0, f"{err_msg} Error: {out}"
--- a/features/clamav/test/test_packages_musthave.py
+++ b/features/clamav/test/test_packages_musthave.py
@@ -1,0 +1,1 @@
+from helper.tests.packages_musthave import packages_musthave as test_packages_musthave
--- a/tests/helper/utils.py
+++ b/tests/helper/utils.py
@@ -127,13 +127,17 @@ def execute_local_command(cmd):
     return rc, out
 
 
-def execute_remote_command(client, cmd):
+def execute_remote_command(client, cmd, skip_error=False):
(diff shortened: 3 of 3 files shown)


features/clamav/test/test_clamav_cron.py
```diff
@@ -0,0 +1,8 @@
+from helper.utils import execute_remote_command
+
+
+def test_clamav_cron(client):
+    # This will already fail if file is absent
+    out = execute_remote_command(client, "cat /var/spool/cron/crontabs/root")
```
nanory, line 6: If this file fails, we will get an error message like this then:
```
cat: /var/spool/cron/crontabs/root: No such file or directory\n
```
Can we adjust the error message in order to make it more clear what the issue is about? Maybe something like this:
```
Cron configuration for ClamAV expected but not found.
```

nanory (changes requested):

features/clamav/test/test_clamav_cron.py
gyptazy: Sure we can but I think when it comes to debugging for someone who isn't into PyTest and Garden Linux, the first error message (even it's not catched) makes more sense. It directly tells the user what, where and why there's an issue while your requested change would lead into more questions. Why did it fail, what it the issue and where and how to debug this?

nanory (approved):

features/clamav/test/test_clamav_cron.py
nanory: Is it? I mean if you not into pytest or Garden Linux, you don't really know why a missing `/var/spool/cron/crontabs/root` is a problem. You need then to check what the test is doing and parse Python code. For me at least, it is more confusing if I get a `cat: /var/spool/cron/crontabs/root: No such file or directory\n` error message instead of a error message which tries to explain to me, why this missing file is a problem in this context. This should be more user friendly imho.

gyptazy: Unfortunately, I don't share your opinion here. Let's solve it by using both - your output together with the error message 👌

PR #1239, opened 2022-08-31
tests(aide): Add unit tests for feature `aide`

gyptazy:
Add unit tests for feature `aide`

/kind test
/area os
/os garden-linux

Add unit tests for feature `aide`

Fixes #1173

**Release note**:

```feature user
NONE
```

**Test**:
```
PASSED aide/test/test_aide_cron.py::test_aide_cron
PASSED aide/test/test_packages_musthave.py::test_packages_musthave
```


features/aide/test/test_aide_cron.py
```diff
@@ -0,0 +1,8 @@
+from helper.utils import execute_remote_command
+
+
+def test_aide_cron(client):
+    # This will already fail if file is absent
+    out = execute_remote_command(client, "cat /var/spool/cron/crontabs/root")
```
nanory, line 6: If this file fails, we will get an error message like this then:
```
cat: /var/spool/cron/crontabs/root: No such file or directory\n
```
Can we adjust the error message in order to make it more clear what the issue is about? Maybe something like this:
```
Cron configuration for aide expected but could not be verified.
```

nanory (changes requested):

nanory (approved):

PR #1245, opened 2022-09-01
tests: make test name for platform tests configurable

MrBatschner:
... and avoid code-duplication

/kind test
/area os
/os garden-linux

Allows to configure the name of a test-run for the platform test and as such, modify the prefix that is used for allocated resources in the cloud providers. Of course, not specifying a `test_name` manually will still fall back to the default that was used before: `gl-test-<YYYYmmDDHHMMSS>`.

**Release note**:

```feature developer
tests: make test name for platform tests configurable
```

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -144,6 +144,9 @@ aws:
     # architecture of the image and VM to be used (optional)
     architecture: x86_64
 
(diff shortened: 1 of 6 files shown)


nkraetzschmar (approved):
lgtm

PR #1246, opened 2022-09-01
test: fix tests to run with non-root user on cloud platforms

MrBatschner:

/kind test
/area os
/os garden-linux

Running the feature tests on cloud platforms such as Azure or AWS will have some problems due to the fact that the user used to log in to the test-VM has no root permissions. This PR:
- adds a `sudo` to commands that require root permissions
- gives to the full path to commands/binaries that reside in `/usr/sbin` which is not in a non-root user's `$PATH`

In addition, it makes sure that additional users that get injected by the cloud platforms are accepted and will not fail the `test_users.py` test.

**Release note**:

```feature developer
test: fix tests to run with non-root user on cloud platforms
```

### DIFF
--- a/features/ali/test/test_users.py
+++ b/features/ali/test/test_users.py
@@ -1,0 +1,4 @@
+from helper.tests.users import users
+
+def test_users(client, ali):
+     users(client, "admin")
--- a/features/aws/test/test_users.py
+++ b/features/aws/test/test_users.py
@@ -1,0 +1,4 @@
+from helper.tests.users import users
+
+def test_users(client, aws):
+     users(client, "admin")
--- a/features/azure/test/test_users.py
+++ b/features/azure/test/test_users.py
@@ -1,0 +1,4 @@
+from helper.tests.users import users
+
+def test_users(client, azure):
+     users(client, "azureuser")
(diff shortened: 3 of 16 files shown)


gyptazy (approved):
thanks for your fixes, lgtm so far, but a fixture `non_aws` needs to be added.

gyptazy (changes requested):
thanks for your fixes, lgtm so far, but a fixture `non_aws` needs to be added. Unfortunately, only this one seems to be missing.

MrBatschner:
added, ptal

gyptazy:
Oh sorry, yesterday evening I haven't seen that `non_gcp` is missing, too. Applying this one fixes this issue (at least I only validated the test pipeline by running `kvm-dev` and and `kvm-dev` including the additional `gardener` feature): 
```
diff --git a/tests/conftest.py b/tests/conftest.py
index 6275cb02..464e7ccb 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -407,6 +407,11 @@ def aws(iaas):
     if iaas != 'aws':
         pytest.skip('test only supported on aws')
 
+@pytest.fixture
+def non_gcp(iaas):
+    if iaas == 'gcp':
+        pytest.skip('test not supported on gcp')
+
 @pytest.fixture
 def gcp(iaas):
     if iaas != 'gcp':
```

MrBatschner:
I added the missing fixture for `non_gcp`, but I also added a fixture `non_hyperscalers` so that we do not have to list all cloud platforms individually.

gyptazy (approved):
thanks, lgtm

PR #1247, opened 2022-09-01
test: remove check for storage account name

MrBatschner:

/kind test
/area os
/os garden-linux

Removes a check for name availability of a storage account for platform tests on Azure. The check seems to be broken when used with a subscription only available through a shared tenant and it is unnecessary as creating the storage account would fail if the name were not available.

**Release note**:

```feature developer
- test: remove check for storage account name on Az platform tests
```


nkraetzschmar (approved):
/lgtm

PR #1258, opened 2022-09-06
GitHub actions: S3 upload use full prefix in object key

nkraetzschmar:

/kind enhancement
/area os
/os garden-linux

Use the full prefix (including the minimal feature set) in the S3 object key rather than just the make target.
This will make it easier for the tekton pipeline to find the uploaded images for publishing.


MrBatschner (approved):
/lgtm

PR #1259, opened 2022-09-06
define feature sets for make targets in files

nkraetzschmar:

/kind enhancement
/area os
/os garden-linux

define feature sets for make targets in files, rather than having duplicate code in the Makefile and use generic make targets (`%` and `%-dev`) instead

features lists in `make_targets` are identical to those from the Makefile prior to this change, generated via the following script:

```shell
#!/bin/bash

set -Eeufo pipefail

makefile="$(cat Makefile)"

grep -F ': build-environment $(SECUREBOOT_CRT)' <<< "$makefile" \
| cut -d : -f 1 \
| grep -v '\-dev$' \
| while read -r target; do
	features="$(grep -F "$target"': build-environment $(SECUREBOOT_CRT)' -A 1 <<< "$makefile" \
	| tail -n 1 \
	| grep -Po '(?<=--features )[a-z0-9\-_,]+')"
	echo "$features" > "make_targets/$target"
done
```


gyptazy:
Sorry for sliding in... Currently, I'm rewriting the `garden-feat` in Python where the `features` are being evaluated.

When building any image, the overall base is represented by a feature of the `platform` type. This might be e.g. `KVM`.

According to this PR we define these sub features (see also: https://github.com/gardenlinux/gardenlinux/pull/1259/files#diff-c2604d0d6c1ce427dd5436aae130b8cd7bf9014be12d4ec807dd8318685ec608)
```
target: kvm
    features:
        - server
        - cloud
        - kvm
``` 

I think we do not need to maintain these ones within the Makefiles, while this gets already done by the include features.
Given by the base [kvm](https://github.com/gardenlinux/gardenlinux/blob/main/features/kvm/info.yaml#L3-L5), it includes [cloud](https://github.com/gardenlinux/gardenlinux/blob/main/features/cloud/info.yaml#L4-L5), which includes [server](https://github.com/gardenlinux/gardenlinux/blob/main/features/server/info.yaml#L3-L5), that includes [base](https://github.com/gardenlinux/gardenlinux/blob/main/features/base/info.yaml#L3-L4) and finally includes [_slim](https://github.com/gardenlinux/gardenlinux/blob/main/features/_slim/info.yaml).

However, by just calling the make target with `kvm` we get:
```
target: kvm
    features:
        - cloud
        - server
        - base
        - _slim
```

AFAIK, this should be the case for all available platforms. This works by calling [garden-feat](https://github.com/gardenlinux/gardenlinux/blob/main/bin/garden-build#L52) during the build time and is implemented within [the currently used Go version](https://github.com/gardenlinux/gardenlinux/blob/0fe683f474fa7845aca4aeadc770a469b0fd85d5/bin/garden-feat.go#L175) and the [upcoming Python version](https://github.com/gardenlinux/gardenlinux/blob/6d6e77e12c896fbb407f0b023fc63b64db5cdce3/bin/garden-feat#L140-L141).

jfyi @nanory

nkraetzschmar:
Yes, we can definitely slim down the list of features explicitly given for each make target :+1:

For this change I just wanted to keep the feature lists exactly as they were in the Makefile to ensure we don't accidentally drop any. I'd suggest we merge this as is and slim down the list of features in a separate PR to keep each change more concise.

gehoern (approved):

PR #1274, opened 2022-09-16
Fix : rootfs not resized properly sometimes

5kt:
rootfs is not properly resized all the time, growfs is running before or in parallel with systemd-repart.
More information : [here](https://github.com/systemd/systemd/commit/7b45d6b6f64e9f5c006bdf31559a77294dbe00ad)


MrBatschner (approved):
/lgtm

PR #1289, opened 2022-09-21
feat: disable fail-fast in all GitHub action matrix builds

nkraetzschmar:

/kind enhancement
/area os
/os garden-linux

disabling fail-fast for GitHub action matrix builds will make it easier to diagnose bugs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,6 +44,7 @@ jobs:
     permissions:
       id-token: write
     strategy:
+      fail-fast: false
       matrix:
         architecture: [ amd64, arm64 ]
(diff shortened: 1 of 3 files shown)


Vincinator:
> GitHub will cancel all in-progress and queued jobs in the matrix if any job in the matrix fails. This property defaults to true.
From: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

Setting this to false will help to identify if an issue only occurs for a single matrix target, or concerns multiple - right?

Sounds already good to me - but is there another reason?

gehoern (approved):

PR #1293, opened 2022-09-21
Add policykit again

### LINKED ISSUE #1292: localhost systemd-networkd[322]: Could not set hostname: Access denied
**What happened:**

During boot in 90x we see the following error message in the journal:

```
localhost systemd-networkd[322]: Could not set hostname: Access denied
```

With 576.x we don't see the error. After boot the `/etc/hosts` file has the hostname where in 576 there is the fully qualified domain name. 

[This discussion](https://github.com/systemd/systemd/issues/13501) may shed some light on what is going on, however I don't see a difference to 576. 90x does not come with policykit but installing it did not make any difference.

marwinski:

/kind but
/area os
/os garden-linux

Hostname cannot be set anymore if policykit is missing. This has been removed altough the original commit message stated that it is required for that purpose.

Fixes #1292

**Release note**:

```feature user

```

### DIFF
--- a/features/base/test/test_sgid_suid_files.py
+++ b/features/base/test/test_sgid_suid_files.py
@@ -25,7 +25,9 @@ import pytest
                  "/usr/bin/chfn,root,root",
                  "/usr/bin/gpasswd,root,root",
                  "/usr/bin/sudo,root,root",
-                 "/usr/bin/passwd,root,root"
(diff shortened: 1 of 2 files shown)


gehoern:
i think this was removed because of the policykit security vulnerabilitys.

but nice to have it back

PR #1295, opened 2022-09-22
CI: Make use of Release Manifests Build via GitHub Actions

AndreasBurger:
/kind cleanup
/area os
/os garden-linux

This PR removes the existing image-build from the Pipelines. Afterwards, the Tekton pipeline will simply use the release-manifests created by the GitHub actions to determine the build-artifacts to publish to the infrastructure-providers.
As a consequence, a lot of code relating to the previous image-build was removed.

There is still some subsequent work to be done although this PR should be functional by itself.
- updating the `BuildTarget` model-class: We should remove all mentions of the `build`-target since it effectively does nothing. However, some code relies on it or validates its presence and it is also used in the `branch.cfg` so I've opted to let it remain for now.
- Updating the `branch.cfg` to remove PR-jobs, since they do nothing but publish a snapshot component-descriptor. Also, the head-update- and the release-job will need to be updated once the target mentioned in the previous point is removed.

~Finally, the correct s3 bucket needs to be configured in `ci/cicd.yaml`, as I am not entirely sure which one will be the correct one in future. Currently, this is set to `gardenlinux`, but the github-actions publish to `gardenlinux-github-releases`. If the latter is correct `ci/cicd.yaml` needs to be adjusted, otherwise the github-actions need to be configured to publish to the `gardenlinux` bucket or the published manifests won't be found.~

**Release note**:

```other developer
Use artifacts built by GitHub Actions instead of Tekton
```


AndreasBurger:
Note: Tekton build failures can be ignored for this PR, they cannot succeed without release-manifests being already present.

gehoern (approved):

PR #1296, opened 2022-09-22
Small fixes for start-vm

5kt:
start-vm has some small issues that prohibit the usage of the pxe features.


Vincinator (approved):
looks good

Issue #1304, opened 2022-09-27
Enhance build-container with gardenlinux repository


adaevitncar:
**What would you like to be added**: The gardenlinux repository should be available in the build container.

**Why is this needed**: In the build container we need to install onmetal-image (cmp. https://github.com/gardenlinux/gardenlinux/issues/1302 ) for testing purpose.

Issue #1302, opened 2022-09-26
Package onmetal-image


adaevitncar:
**What would you like to be added**: The binary from https://github.com/onmetal/onmetal-image (must be build) should be available as deb-package in the gardenlinux reposiory.

**Why is this needed**: _oci feature and tests

PR #1309, opened 2022-09-29
Add feature _oci to bundle build articats as OCI compatible 

### LINKED ISSUE #1144: package gardenlinux releases into an oci
oci supports now oci artifacts:
- originally in oci you could only publish layers each layer was a tar containing the data of the relevant option
  (nginx consited of the layer nginx and debian -> nginx had a metadata refering to the layers see https://dlorenc.medium.com/oci-artifacts-explained-8f4a77945c13 -> this had certain impacts: e.g. if you had a debian layer this was automatically deduplicated because it was referenced via a hash and therefore after an update of nginx to the next debian - the dependency to the old stuff was removed)
- now you can store anything as a "layer" or better artifacts. that means we could now release our different flavours as an oci artifact and make sure that descriptive stuff like (logs, os-release, contend description ...) goes into a standard layer but anything that is binary (vmlinuz, initrd, boot disk) goes as an artifact (implicit we have only one version of the kernel everything else is auto deduplicated). So we can boot out of any oci registry.

this is a proposal for a tool to use all this. we need to orchestrate.
https://github.com/onmetal/onmetal-image

adaevitncar:
/kind enhancement
/area os
/os garden-linux

create oci bundled images (kernel, rootfs, initrd, kernel command line) for created build artifacts.

Fixes #1144 

Issues #1302  and #1304 must be solved. Until they are finised the used onmetal-image binary is not available

**Release note**:

```other developer
Add feature _oci to bundle build artifacts as OCI compatible 
```


nanory (approved):

gyptazy (approved):
LGTM, thanks!

PR #1328, opened 2022-10-11
enhancement(integration-tests): Move tests to new file structure

### LINKED ISSUE #1222: `PyTest`: Move legacy tests from `tests/integration/test_gardenlinux.py` to new layout style (`features/$FEAT/test/test_$test.py`)
We should move all tests from the legacy integration test file ([test_gardenlinux.py](https://github.com/gardenlinux/gardenlinux/blob/main/tests/integration/test_gardenlinux.py)) to the new layout style (according to [our docs](https://github.com/gardenlinux/gardenlinux/tree/main/tests#unit-tests)). This may avoid further confusions when adding new tests (regarding they're unit-, integration-, platform tests). All tests are based on PyTest by a single common test platform, now. However, fixtures are used to pin them to the belonging type.

Next to this, we shouldn't only copy them to the new syntax. Instead, we should make sure to integrate them in a proper way. In some cases it may make sense to adjust already present tests and integrate them into these ones or to change to PyTests's parametrize function to parametrize options for tests instead of hardcoded content. An example is given [here](https://github.com/gardenlinux/gardenlinux/commit/bcd2aa16f59c95cfd215f799b402f8759e523b4c), where `test_apparmor` got changed to the parametrize function.

**Why is this needed**:
This avoids further confusions while currently both solution are sourced by PyTest.

**How to proceed**:
I think we should create a single task for each tests within the legacy test file. As a result, everyone can take a single test and convert this one without intercepting someone else. We will discuss this within our today's weekly meeting. Afterwards, I'll create all subtasks and add them here as well.

gyptazy:
enhancement(integration-tests): Move tests to new file structure

/kind enhancement
/area os
/os garden-linux

Move tests to new file structure. This integrates the legacy `integration-tests` into the new Pytest file structure to have all tests in their desired features and test folders.

Fixes #1222

**Release note**:

```feature user
NONE
```


gyptazy:
@MrBatschner please have a short look...

gyptazy:
While @nanory already approved this already it should still be validated on all hyperscaler platforms by @MrBatschner. If everything is fine for you, feel free to ping me to solve the merge conflict.

PR #1329, opened 2022-10-11
feat: Add Garden Linux Repo and onmetal-image to build container

### LINKED ISSUE #1304: Enhance build-container with gardenlinux repository

nanory:
/kind enhancement
/area os
/os garden-linux

This PR adds the Garden Linux Repo to the build-image container in order to install some build tools that are built in the Gitlab Pipeline and published to the Garden Linux Repo. This makes them accessible in the Garden Linux build.

Fixes #1304

### DIFF
--- a/container/build-image/Dockerfile
+++ b/container/build-image/Dockerfile
@@ -1,45 +1,65 @@
-ARG build_base_image=gardenlinux/slim
-FROM 	$build_base_image
+ARG	build_base_image=gardenlinux/slim
+FROM	$build_base_image
 ARG	DEBIAN_FRONTEND=noninteractive
+ARG	GARDENLINUX_MIRROR_KEY=/etc/apt/trusted.gpg.d/gardenlinux.asc
+ARG	VERSION
+
+COPY gardenlinux.asc $GARDENLINUX_MIRROR_KEY
 
 RUN if [ "$(dpkg --print-architecture)" != amd64 ]; then dpkg --add-architecture amd64; fi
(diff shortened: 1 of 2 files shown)


container/build-image/Dockerfile
```diff
@@ -30,6 +32,11 @@ RUN apt-get update \
 		openssl \
 		libengine-pkcs11-openssl
 
+COPY gardenlinux.asc $GARDENLINUX_MIRROR_KEY
+RUN echo "deb [signed-by=$GARDENLINUX_MIRROR_KEY] https://repo.gardenlinux.io/gardenlinux $VERSION main" >> /etc/apt/sources.list \
```
gyptazy, line 36: I know you may use multiple `run` commands for readability when it comes to complex commands but (IMHO) I think we may first copy the file and combine afterwards all runs (but **not** all within a single `apt-get` call) in a single `run` command like in https://github.com/gardenlinux/gardenlinux/blob/main/container/integration-test/Dockerfile. Indention and comments may help here for further readability. What do you think?

nanory: Good point. I think we should streamline this and make sure to create as few layers as possible. Therefore, I adjusted the Dockerfile.

PR #1333, opened 2022-10-13
fix(syslinux): use full name for initrd

5kt:
update-syslinux fails at the moment as the initrd file cannot be found. We need to use the full path when testing and using, e.g. initrd.img-VERSION.


gyptazy (approved):
LGTM
thx for the fix!

PR #1334, opened 2022-10-13
feature(ssh): Make ssh a dedicated feature

### LINKED ISSUE #302: Enable fail2ban service
There is already a lot prepared for fail2ban. However, the fail2ban package is commented. fail2ban should be activated to secure SSH a little.

fail2ban service should be enabled by default.

**Why is this needed**:

Increase the security of the nodes against attacks.

### LINKED ISSUE #1193: Evaluate what it would mean to remove SSH from Garden Linux
SSH isn't explicitly necessary for Garden Linux by default. However, it is installed by the `server` feature so it is part of most GL images.

Typically, SSH is used to debug running Garden Linux VMs on cloud providers but in such cases GL is also used with `gardener` which provides a container runtime and Kubernetes. This way, one could spawn a debug and troubleshot container with root privileges to debug the host system. SSH isn't necessarily needed then.

Next, we use SSH for pytest. Removing SSH could be more complicated, so we will probably not be able to remove SSH completely but we may be able to just install and provide SSH during our pytests but do not need to already provide SSH in the GL image right from the beginning.

There may be other cases, too. As a result, we should analyze the implications of removing SSH from GL, check if it is possible and verify what efforts and issues we would have to solve.

**Why is this needed**:
Without having SSH installed, we also lose a lot of hardening requirements which would make our life much easier.

gyptazy:
feature(ssh): Make ssh a dedicated features

/kind enhancement
/area os
/os garden-linux

Make `ssh` & `fail2ban` a dedicated feature
 * New feature: SSH
 * New feature: Fail2ban
 * Code cleanup (deprecated files, content, comments,...)
 
Fixes #1193 
Fixes #302 

**Release note**:

```feature user
- category: feature
- target_group: operator
```


gyptazy:
Validating the functionality of `fail2ban`:

You can see the entry within the chain `f2b-sshd`: 
```
root@garden:~# iptables -vnL
Chain INPUT (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination         
    2    88 f2b-sshd   tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            multiport dports 22

Chain FORWARD (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination         

Chain OUTPUT (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination         

Chain f2b-sshd (1 references)
 pkts bytes target     prot opt in     out     source               destination         
    2    88 REJECT     all  --  *      *       10.0.2.2             0.0.0.0/0            reject-with icmp-port-unreachable
    0     0 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0           
root@garden:~# 
```

gyptazy:
This features are tested within the following constellations:
 * gardenlinux without ssh-server
 * gardenlinux with ssh-server
 * gardenlinux with ssh-server and fail2ban for ssh

After merge it is in the operators decision to build the needed images with combinations of ssh and fail2ban or just without anything of this new features which makes Garden Linux more modular based on its feature's system. In default, the `server` features includes `ssh` which includes `fail2ban`. Within a default installation (without any further changes) a ssh-server will still be present including fail2ban for ssh services.

CC: @nanory @gehoern

features/ssh/README.md
```diff
@@ -0,0 +1,12 @@
+# SSH
+
+## General
+Ssh (Secure Shell) is a program for logging into a remote machine and for executing commands on a remote machine. It provides secure encrypted communications between two untrusted hosts over an insecure network. X11 connections and arbitrary TCP/IP ports can also be forwarded over the secure channel. It can be used to provide applications with a secure communication channel.
```
nanory, line 4: I wouldn't call the hosts untrusted. The network might be untrusted, though.

features/ssh/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Overview
+In default, `ssh` will be installed by `server` as a dependency. However, it may be safely removed if the operator does not want to use it.
```
nanory, line 7: `By default, ...`

features/ssh/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Configuration
+
+### Defaults
+The configuration gets tweaked according the used keys and ciphers as well as of further configurations regarding the current best practices of security and may change frequently.
```
nanory, line 12: * I would highlight, that we mainly use Debian defaults for ciphers and other security related configurations.
* `according to`

nanory (changes requested):

features/ssh/README.md
gyptazy: Until you verified the host FP's it's untrusted. btw, this is taken from upstream man page.

nanory (approved):

Vincinator:
Hi @gyptazy, this PR will not be part of the next release (within the next few days). 
Additionally, the feature `ssh` and `fail2ban` should be one feature. 
Fail2ban will require some testing, also in gardener context.

gyptazy:
Hey @Vincinator,

thanks for the update. For merging `fail2ban` into `ssh` we should have a small talk within our tomorrow's weekly meeting. I'd like to avoid this and keep them separated.

Thanks!

See you tomorrow,
gyptazy

Vincinator:
Yes, makes sense to sync in a weekly.
I am not invested in either decision. I was the messenger :) 
And the key message is: Do not merge before the next release.

But here is my perspective nonetheless:

**Pro: merge into one feature:**
- design decision is to not replicate a packet manager, but to add a higher level of opinionated functionality for the user
- fail2ban adds another layer to prevent brute force attacks

**Pro: separate features:**
- would help in cases where Gardener users have to (cleanly) turn it off

gyptazy:
Based on the notes from our weekly meeting a new PR has been created. #1351

PR #1345, opened 2022-10-18
fix(start-vm): add garden as supported host_dist

5kt:
We should be able to use the start-vm script when running on a Garden Linux host.

Fixes #


gyptazy (approved):
lgtm

PR #1346, opened 2022-10-18
file.include.stat extras

### LINKED ISSUE #920: Custom permissions for files from file.include
A way to configure custom permissions for files included via file.include from the feature.
This could be done, of course, via exec.config, but maybe it would be nicer to have something like file.include.stat, similar to dpkg statoverride where one would specify the owner/group and whatever permissions are wanted.

**Why is this needed**:
This can also be considered a bug. This is needed because at the moment all files provided via file.include are going to be owned by root:root and have whatever permissions those files have on the build machine.

5kt:
Related to #920, already handled by @gyptazy via #1331

This is just forcing sane defaults (owner root:root, permissions 755/644) for files/directories included via files.include, irrespective of their owner/permissions on the build machine; executable bit will be kept.
Of course, an override will then be possible via file.include.stat.

Setting custom permissions via file.include.stat is now done chrooted to match the users/groups from the build.

Fixes #920

### DIFF
--- a/bin/garden-config
+++ b/bin/garden-config
@@ -44,7 +44,7 @@ printf "## copying from file.include\n"
 for i in $(tr ',' '\n' <<< $features); do
 	if [ -d "$featureDir/$i/file.include" ]; then
 		printf "### $i:\n"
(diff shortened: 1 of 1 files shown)


gyptazy (approved):
LGTM, thanks.

PR #1355, opened 2022-10-24
doc(build): Fix link in `README.md`

### LINKED ISSUE #1348: Provide air gapped/offline build support
Maybe we can provide further air gapped (offline) build support for Garden Linux. While most of the pipeline already works in an air gapped setup, there're still some things left that need to be covered mostly in documentation. This should cover how to get the needed tools, additional resources (cfssl, container images, external utils from pytevts etc.). I think except of unit tests we do not any further integration while this tests won't be able to perform in an air gapped environment. 

**Why is this needed**:
It might come to the point where users are in an air gapped environment and have the requirement to build their artifacts from source (instead of copying random bins). Therefore, we should adjust (if needed) the pipeline and provide a documentation how to build GL in an air gapped env.

gyptazy:
doc(build): Fix link in `README.md`

/kind TODO
/area os
/os garden-linux

Fix link in `README.md`

Fixes #1348

**Release note**:

```feature user
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@
   - Bare-metal systems
 
 ## Quick Start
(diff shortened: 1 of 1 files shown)


chrinorse (approved):
lgtm

PR #1396, opened 2022-11-10
feat: add a gardener flavour set

MrBatschner:

/kind enhancement
/area os
/os garden-linux

Introduces a `gardener` flavor-set to only contain those public cloud platforms currently used by Gardener and leave out anything bare-metal.

**Release note**:

```feature user
feat: add a gardener flavour set
```

### DIFF
--- a/flavours.yaml
+++ b/flavours.yaml
@@ -1,5 +1,15 @@
 ---
 flavour_sets:
+  - name: 'gardener'
+    flavour_combinations:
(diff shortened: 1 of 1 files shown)


Vincinator (approved):
Changes are non breaking. And require user/build pipelines to explicitly call this new feature set. So I see no reason to block this PR.

gehoern (approved):

PR #1421, opened 2022-11-18
fix: set file attributes of key in Dockerfile

### LINKED ISSUE #1420: Set file permission attributes of gardenlinux.asc in container
**What happened:**
- `container/base-test/Dockerfile` and `container/build-image/Dockerfile` copy the `gardenlinux.asc` inside the container image
- On a system with a more restricted umask (e.g. garden linux itself), the build will fail because the `gardenlinux.asc`can not be read

**Fix**
Set ownership and file permissions explicitly in container

Vincinator:
Explicitly sets the file permissions as required by the container image.,

Fixes #1420 

Kudos to @5kt

### DIFF
--- a/container/base-test/Dockerfile
+++ b/container/base-test/Dockerfile
@@ -6,7 +6,9 @@ ENV PYTHONPATH /gardenlinux/bin:/gardenlinux/ci:/gardenlinux/ci/glci:/gardenlinu
 ARG GARDENLINUX_MIRROR_KEY=/etc/apt/trusted.gpg.d/gardenlinux.asc
 ARG VERSION
 
-COPY gardenlinux.asc $GARDENLINUX_MIRROR_KEY
+COPY gardenlinux.asc $GARDENLINUX_MIRROR
+RUN  chown root:root $GARDENLINUX_MIRROR_KEY \
+        && chmod 644 $GARDENLINUX_MIRROR_KEY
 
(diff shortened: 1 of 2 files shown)


5kt (approved):

gyptazy (changes requested):
I think this'll require an additional RUN block or can be extended within the followed one to avoid this error:

```
 > [2/6] COPY gardenlinux.asc /etc/apt/trusted.gpg.d/gardenlinux.asc         && chown root:root /etc/apt/trusted.gpg.d/gardenlinux.asc         && chmod 644 /etc/apt/trusted.gpg.d/gardenlinux.asc:
------
failed to compute cache key: "/root:root" not found: not found```

Vincinator:
> I think this'll require an additional RUN block or can be extended within the followed one to avoid this error:
> 
> ```
>  > [2/6] COPY gardenlinux.asc /etc/apt/trusted.gpg.d/gardenlinux.asc         && chown root:root /etc/apt/trusted.gpg.d/gardenlinux.asc         && chmod 644 /etc/apt/trusted.gpg.d/gardenlinux.asc:
> ------
> failed to compute cache key: "/root:root" not found: not found```
> ```

Wow, you are fast! Fixed it

gyptazy (approved):
LGTM with [73a2d9a](https://github.com/gardenlinux/gardenlinux/pull/1421/commits/73a2d9af05516a79fd5a24a9a8580fe01e3b3390), now. However, we might combine the two runs into a single layer.

PR #1423, opened 2022-11-18
Enable debian mirrors in a developer mode

### LINKED ISSUE #1422: `BuildPipeline`: Allow declaration & usage of Debian packages from official Debian mirror
via @Vincinator:

We want to have the flag `--debian` for the build pipeline that allows us to define and use package in `package.include` that are not part of the Garden Linux repository.

We can rely on the repository priority system to avoid obtaining all packages from the Debian mirrors.

Vincinator:
As a Garden Linux Developer I want to be able to include packages for example via `features/pkg.include` that are not part of the repo.gardenlinux.io. 

Using only the repo.gardenlinux.io must be the default.

Fixes #1422

**Notes**

- instead of checking `-z $nonDebian` check for ` "$debian" == "1"` 
- build.sh accepts `--debian` flag which is passed down to garden-build -> garden-init

### DIFF
--- a/bin/garden-build
+++ b/bin/garden-build
@@ -3,7 +3,7 @@ set -Eeuo pipefail
 
 thisDir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 source "$thisDir/.constants.sh" \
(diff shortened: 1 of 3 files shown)


gyptazy:
deprecated by #1426

PR #1425, opened 2022-11-18
feature(container): Obtain `awscliv2` from pip instead `aws`

### LINKED ISSUE #1405: Replace tools in test containers with proper packages
As of now, we are using multiple external binaries in our `integration-test` container. In order to enhance security, it would be good to validate the integrity of such binaries and tools. Usually, this can be done by checksums like discussed in #1383. However, some binaries like `awscli` are not versioned, so they are updated continuously which makes validation based on checksums complicated.

A better solution would be the usage of signatures which leads to the question: "Why not using a package management system that already provides such functionality".

For that reason, we should take a look into the binaries used in the `integration-test` and `base-test` containers and replace them with equivalent packages from Debian or PIP.  

**Why is this needed**:
This is needed to enhance the overall security of our test pipeline.

gyptazy:
feature(container): Obtain `awscliv2` from pip instead `aws`

/kind cleanup
/area os
/os garden-linux

Obtain `awscliv2` from pip instead `aws`

Fixes #1405

Should be validated by running the full integration tests for `AWS`

**Release note**:

```feature user
NONE
```

### DIFF
--- a/container/integration-test/Dockerfile
+++ b/container/integration-test/Dockerfile
@@ -9,11 +9,8 @@ RUN : "Eval hardware architecture" \
      && [ $arch = "x86_64" ] && aliyun_arch=amd64 || aliyun_arch=arm64 \
      && [ $arch = "x86_64" ] && ossutil_arch=64 || ossutil_arch=arm64 \
     : "Install AWS requirements" \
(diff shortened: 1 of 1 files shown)


Vincinator (dismissed):

gyptazy:
ping @gehoern (as discussed within the backlog grooming)

gehoern (approved):

PR #1426, opened 2022-11-21
feature(devel): Add new build command `--debian-mirror`

### LINKED ISSUE #1422: `BuildPipeline`: Allow declaration & usage of Debian packages from official Debian mirror

gyptazy:
feature(devel): Add new build command `--debian-mirror`
 Add possibility to add the native Debian repository to chroot when
 installing package via `pkg.include` function (only in dev mode).

/kind enhancement
/area os
/os garden-linux

Add new build command `--debian-mirror`. This adds the possibility to add the native Debian repository to chroot when
 installing package via `pkg.include` function and will only work on non production builds. When building a prod artifact this will fail with the following message:
```
Using the native Debian repository on production builds (_prod) is not supported.
```

Fixes #1422

**Release note**:

```feature user
- category: feature
- target_group: developer
```

### DIFF
--- a/bin/garden-build
+++ b/bin/garden-build
@@ -3,7 +3,7 @@ set -Eeuo pipefail
 
 thisDir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
 source "$thisDir/.constants.sh" \
-	--flags 'no-build,debug,skip-tests,suite:,gardenversion:,timestamp:' \
+	--flags 'no-build,debug,skip-tests,suite:,debian-mirror,gardenversion:,timestamp:' \
 	--flags 'ports,arch:,qemu,features:,disable-features:,commitid:,userid:,usergid:' \
 	--flags 'suffix:,prefix:' \
 	--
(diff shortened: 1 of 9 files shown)


bin/garden-build
```diff
@@ -13,25 +13,28 @@ export REPO_ROOT="$(readlink -f "${thisDir}/..")"
...
-		--gardenversion|--timestamp) version="$1"; shift ;; # timestamp is a parameter this time
-		--suffix) suffix="$1"; shift ;; # target name prefix
-		--prefix) prefix="$1"; shift ;; # target name suffix
-		--commitid) commitid="$1"; shift ;; # build commit hash
-		--commitid-long) commitid_long="$1"; shift ;; # build commit hash, untruncated - used for /etc/os-release only
-		--skip-tests) tests=0 shift ;; # skip tests
-		--userid) userID="$1" shift ;;
-		--usergid) userGID="$1" shift ;;
+		--debug)			debug=1 ;;			# for jumping in the prepared image"
+		--debian-mirror)		debianMirror=$1 ; shift ;; 	# for adding Debian native mirrors (only dev; default: 0)
```
Vincinator, line 23: Input value `$1` is not used. Debian repo is hardcoded to `echo "deb https://deb.debian.org/debian testing main"` in file `bin/garden-debian-sources-list`
```suggestion
		--debian-mirror)		debianMirror=1 ;; 	# for adding Debian native mirrors (only dev; default: 0)
```

since this is a developer parameter, i would say to keep the parameter simple and use a hard coded value

bin/garden-init
```diff
@@ -30,30 +30,35 @@ exclude=
...
-		--keyring-plain) keyringPlain="$1"; shift ;;
-		--arch)		arch="$1"; shift ;;
-		--include)	include="${include:+$include,}$1"; shift ;;
-		--exclude)	exclude="${exclude:+$exclude,}$1"; shift ;;
-		--features)	features="${features:+$features,}$1"; shift ;;
-		--check-gpg)	noCheckGpg=  ;;
-		--no-check-gpg)	noCheckGpg=1 ;;
+		--debian)		nonDebian= ;;
+		--debian-ports)		nonDebian= ; debianPorts=1 ;;
+		--debian-mirror)	debianMirror=0 ; debianMirror=1 ;;
```
Vincinator, line 41: Maybe simplify to just `debianMirror=1 ;;`? Maybe a typo, and intention was to set a different variable to 0?

bin/garden-init
```diff
@@ -156,11 +165,29 @@ if [[ -v PKG_DIR ]]; then
...
-	"$thisDir/garden-debian-sources-list" --gardenlinux-only \
-		$([ -z "$debianPorts" ] || echo '--ports') \
-		$([[ -v PKG_DIR ]] && echo --local-pkgs /run/packages) \
-		"$targetDir" "$suite" "$version"
-	cp "$keyringPlain" "$targetDir/etc/apt/trusted.gpg.d/gardenlinux.asc"
+
+	if [ $debianMirror -eq 1 ]; then
+
+		# Integrate Garden Linux repository with additional Debian native mirror
+		"$thisDir/garden-debian-sources-list" --gardenlinux-only --debian-mirror \
```
Vincinator, line 172: Maybe we should refactor `garden-debian-source-list` so that:
- the `--gardenlinux-only` case is the default (not specifying --gardenlinux-only will result in the current `--gardenlinux-only` behavior
- remove `--gardenlinux-only` since it is the default
- if `--debian-mirror`  is specified then we can use the negation of `--gardenlinux-only` logic in `garden-debian-sources-list` which respects the `suite` parameter


**Motivation:**
specifying both; `--gardenlinux-only --debian-mirror` can be confusing


What do you think?

Vincinator (commented):
Please let me know if my feedback makes sense to you. 
Will close my other PR, i think this one is the better foundation. 

Thank you!

bin/garden-init
gyptazy: I think this is a great idea and we should remove legacy code (e.g. `squeeze` etc.). I've just pushed some changes to prepare this for the outer scripts. If this is fine for you, I'd raise a new issue for refactoring the `garden-debian-sources-list` script (I've only adjusted the default for gardenlinux-repo usage).

Vincinator: This makes sense for me. Thanks for creating the issue!

gyptazy:
All checks passed. fyi @Vincinator

Vincinator (approved):

PR #1434, opened 2022-11-23
Enable ignition on vmware

5kt:
/kind enhancement
/area os
/os garden-linux

We'd like to have ignition enabled on VMware.


Vincinator (approved):
Looks good. However, I need VMWare platform access to test this. Merging this with the note that we might need to iterate on it.

PR #1443, opened 2022-11-28
remove deprecated google api modules

### LINKED ISSUE #1325: Garden Linux UEFI secure boot on GCP

Support for UEFI secure boot on GCP. This is the GCP documentation:

https://cloud.google.com/compute/shielded-vm/docs/shielded-vm

(is is apparently named shielded vm)

**Why is this needed**:

This is part of securing the boot chain with secure boot and dm-vertiy on GCP.

chrinorse:

/kind enhancement
/area os
/os garden-linux

* remove deprecated Google API modules
* deploy images with secureboot enabled to GCP

Fixes #1325 

**Release note**:

```feature user

```


tests/integration/gcp.py
```diff
@@ -104,47 +99,27 @@ def _gcp_create_bucket(self, bucket_name):
...
-            client = self._compute.globalOperations()
-        response = client.wait(**kwargs).execute()
-
-        if response["status"] != "DONE":
-            self.logger.error("Operation failed %s" % json.dumps(response, indent=4))
-            error = ""
-            if "error" in response:
-                error = response["error"]
-            raise Exception("Operation %s failed: %s" % (operation, error))
+        while not operation.done():
```
nanory, line 102: We should probably use `operation.result()` here instead of `operation.done()`:
* https://googleapis.dev/python/google-api-core/latest/operation.html
* https://googleapis.dev/python/google-api-core/latest/futures.html#google.api_core.future.polling.PollingFuture.result
* https://github.com/googleapis/python-compute/blob/HEAD/samples/snippets/firewall/delete.py

tests/integration/gcp.py
```diff
@@ -104,47 +99,27 @@ def _gcp_create_bucket(self, bucket_name):
 
 
     def _gcp_wait_for_operation(self, operation):
-        self.logger.info(f"Waiting for operation {operation['name']} to complete...")
```
nanory, line 107: We should probably leave some log messages here, to better understand what is happening in the background. The following snippet is an good example, maybe we can stick to it:
* https://github.com/googleapis/python-compute/blob/HEAD/samples/snippets/firewall/delete.py

nanory (changes requested):

tests/integration/gcp.py
```diff
@@ -519,21 +514,30 @@ def create_vm(self):
             },
         }
 
-        operation = self._compute.instances().insert(project=self.project, zone=self.zone, body=config).execute()
+        if self.config['secureboot']:
+            shielded_instance_config = {
```
nanory, line 518: I wouldn't use a temporary variable here but instead directly add the "shielded_instance_config" to the `config` variable like this:
```
config["shielded_instance_config"] = {
  "enable_secure_boot": True,
  ...
}
```

tests/integration/gcp.py
```diff
@@ -440,21 +436,20 @@ def _get_image(self, project, image_name):
...
+        self.logger.info(f"Waiting for {hostname} to respond...")
         while True:
-            response = os.system("ping -c 1 " + hostname)
+            response = os.system("timeout 1 bash -c \"</dev/tcp/" + hostname + "/22\"")
             if response == 0:
-                self.logger.info(f"Instance {hostname} is reachable, waiting another 20 seconds...")
-                time.sleep(20)
+                self.logger.info(f"Instance {hostname} is reachable...")
                 return
             time.sleep(1)
```
nanory, line 454: Although we old implementation didn't had this as well, we should probably add some sort of retry counter, because otherwise this function could hang forever, if the SSH port never comes up. We should then use a reasonable default for this retry counter.

tests/integration/gcp.py
```diff
@@ -401,36 +386,47 @@ def _upload_image(self, image_name, image):
...
+        if self.config['secureboot']:
+            cert_file_type = self.config['cert_file_type']
+            shielded_instance_initial_state = {
+                'shielded_instance_initial_state': {
+                    'dbs': [self._get_file_content_buffer(self.config['db_path'], cert_file_type)],
+                    'keks': [self._get_file_content_buffer(self.config['kek_path'], cert_file_type)],
+                    'pk': self._get_file_content_buffer(self.config['pk_path'], cert_file_type),
+                }
+            }
+            config.update(shielded_instance_initial_state)
```
nanory, line 415: We should directly write `shielded_instance_initial_state` to `config` instead of using a temporary variable and the `update()` method

tests/integration/gcp.py
```diff
@@ -401,36 +386,47 @@ def _upload_image(self, image_name, image):
...
+                'raw_disk': {
                     'source': blob_url,
                 },
                 'labels': self._tags,
-            },
-        )
+            }
+
+        if self.config['uefi'] or self.config['secureboot']:
+            guest_os_features = {'guest_os_features': [{'type_': "UEFI_COMPATIBLE"}]}
```
nanory, line 403: Same as commend below

tests/integration/gcp.py
```diff
@@ -519,21 +514,30 @@ def create_vm(self):
...
+                    "enable_vtpm": True,
+                }
+            }
+            config.update(shielded_instance_config)
+
+        operation = self._compute_instances.insert(project=self.project, zone=self.zone, instance_resource=config)
         self._gcp_wait_for_operation(operation)
-        url = operation.get("targetLink")
+        url = self._compute_instances.get(project=self.project, zone=self.zone, instance=name).self_link
         self._instance_name = self._get_resource_name(url)
```
nanory, line 530: Does `_get_resource_name` return another name than the already available variable `name` in this function? If not, we may could remove this function here as well.

tests/integration/gcp.py
```diff
@@ -90,6 +86,18 @@ def validate_config(cls, cfg: dict, image: str, test_name: str, credentials):
             cfg['ssh']['ssh_key_name'] = f"key-{test_name}"
         if not 'user' in cfg['ssh']:
             cfg['ssh']['user'] = "gardenlinux"
+        if not 'uefi' in cfg:
+            cfg['uefi'] = False
+        if not 'secureboot' in cfg:
+            cfg['secureboot'] = False
+        if not 'db_path' in cfg:
+            cfg['db_path'] = "/gardenlinux/cert/secureboot.db.auth"
```
nanory, line 94: We should probably group these parameters (so `db_path`, `kek_path`, `pk_path` and `cert_file_type`) underneath the `secureboot` parameter because these parameters only make sense if you also enable `secureboot`. 

We would streamline this with the `ssh` parameter, which does do the same with its parameters.

docs/deployment/gcp-secureboot.md
```diff
@@ -0,0 +1,207 @@
...
+
+From there, you can now use `ssh` to connect to the remote GCP instance running Garden Linux with secure boot enabled. Use the SSH private key you configured in the configuration yaml file.
+
+Additionally, the output also contains which address can be used to access the remote machine:
+```
+INFO     gcp-testbed:gcp.py:461 Waiting for 34.76.177.199 to respond...
+INFO     gcp-testbed:gcp.py:465 Instance 34.76.177.199 is reachable...
+```
+
+The instances is running with the so called `gardenlinux` user, so you can access this machine by the following command:
```
nanory, line 163: `The instance ...`

nanory (changes requested):

tests/integration/gcp.py
```diff
@@ -401,36 +386,49 @@ def _upload_image(self, image_name, image):
...
+                'raw_disk': {
                     'source': blob_url,
                 },
                 'labels': self._tags,
-            },
-        )
+            }
+
+        if self.config['uefi'] or self.config['secureboot']:
+            config.update(
```
nanory, line 403:
```
config['guest_os_features'] = [
  {'type_': "UEFI_COMPATIBLE"}
]
```

tests/integration/gcp.py
```diff
@@ -401,36 +386,49 @@ def _upload_image(self, image_name, image):
...
+                {'guest_os_features': [{'type_': "UEFI_COMPATIBLE"}]}
+            )
+
+        if self.config['secureboot']:
+            cert_file_type = self.config['secureboot_parameters']['cert_file_type']
+            config.update(
```
nanory, line 409:
```
config['shielded_instance_initial_state'] = {
  'dbs': [self._get_file_content_buffer(self.config['secureboot_parameters']['db_path'], cert_file_type)],
  'keks': [self._get_file_content_buffer(self.config['secureboot_parameters']['kek_path'], cert_file_type)],
  'pk': self._get_file_content_buffer(self.config['secureboot_parameters']['pk_path'], cert_file_type)
}
```

tests/integration/gcp.py
```diff
@@ -519,22 +515,31 @@ def create_vm(self):
             },
         }
 
-        operation = self._compute.instances().insert(project=self.project, zone=self.zone, body=config).execute()
+        if self.config['secureboot']:
+            config.update(
```
nanory, line 519:
```
config['shielded_instance_config'] = {
  "enable_secure_boot": True,
  "enable_integrity_monitoring": True,
  "enable_vtpm": True
}
```

nanory (changes requested):

tests/integration/gcp.py
```diff
@@ -440,27 +438,25 @@ def _get_image(self, project, image_name):
...
             self.logger.debug(f"Failed to find image in {project=}")
             return None
 
     def _wait_until_reachable(self, hostname):
-        self.logger.info(f"Waiting for {hostname} to respond to ping ...")
-        while True:
-            response = os.system("ping -c 1 " + hostname)
+        self.logger.info(f"Waiting for {hostname} to respond...")
+        i=0
+        while i<20:
```
nanory, line 452: A python linter would complain about it. There should be spaces between the operands:
```
i = 0
while i < 20:
```

tests/integration/gcp.py
```diff
@@ -440,27 +438,25 @@ def _get_image(self, project, image_name):
...
+                self.logger.info(f"Instance {hostname} is reachable...")
                 return
-            time.sleep(1)
-
-    def _get_resource_name(self, url):
-        """Get resource name from GCP url"""
-        return urlparse(url).path.split("/")[-1]
+            self.logger.info(f"Waiting for {hostname} to respond...")
+            time.sleep(2)
+            i += 1
```
nanory, line 459: We should probably raise some sort of exception, shouldn't we? If we waited 20 seconds for the SSH server to get available and it's still not there, we should raise something like this:
```
raise Exception("Remote host unreachable")
```

In the upper function, we could then catch it with a `try ... except` block to handle this error properly and maybe cancel the whole execution.

nanory (changes requested):

tests/integration/gcp.py
chrinorse: What's wrong with the update function, the code is a lot more readable with the update function?

nanory: It's not the way how it's done in Python. From an programmer perspective this function costs more than simply assigning values. The updates() function should only be used, if you want to merge two dictionaries into a single one. In fact, what you are doing is creating a temporary third dictionary, which contains another dictionary as a value in order to merge the third one with the first dict. 

What you really want to do is a simple value assignment within your first dictionary (technically just linking two memory blocks with each other). Although, the value itself may be an dictionary again, it is recommended to directly assign values, which is a simple task from a CPU perspective compared to the update function which compares dictionaries key one by one with each others and then tries to merge the values accordingly.

nanory (approved):

PR #1444, opened 2022-11-30
add vmware ovf properties for ignition

defo89:
/kind enhancement
/area os
/os garden-linux

Supports ignition ovf properties. I think it still needs the logic added to act upon these params during the first boot.


Vincinator (approved):
Thanks for the fix

gehoern (approved):

PR #1452, opened 2022-12-05
Feature/1440 create cleanup script for gcp

### LINKED ISSUE #1440: Create cleanup script for GCP
We should add a cleanup script for GCP similar to the one used for AWS (`cleanup_orphaned_aws.py`).

**Why is this needed**:
This makes a cleanup on GCP much easier especially if you run the integration tests with the `keep_running` parameter set to `true`.

chrinorse:

/kind enhancement
/area os
/os garden-linux

adds a script to delete resources created during the integration tests on GCP
Fixes #1440 

**Release note**:

```feature user

```

### DIFF
--- a/tests/tools/clean_orphans_gcp.py
+++ b/tests/tools/clean_orphans_gcp.py
@@ -3,6 +3,7 @@
 import argparse
 import re
 import os
+import sys
 
 import google.cloud.compute as compute
 import google.cloud.storage as storage
@@ -96,7 +97,7 @@ def delete_instance(project, zone, name):
 def add_to_inventory(inventory_dict, key, value):
     try:
         if type(inventory_dict[key]) is dict:
-            inventory[key].update(value)
+            inventory_dict[key].update(value)
     except KeyError:
         inventory_dict[key] = value
     return inventory_dict
@@ -109,7 +110,7 @@ def sanity_check(inventory_dict, dependencies):
     return True
 
 
-while True:
+def main():
 
     buckets = get_buckets()
     images = get_images(args.project)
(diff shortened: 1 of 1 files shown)


MrBatschner (changes requested):
Could you add the option to provide a `serviceaccount.json` for authentication, comparable to https://github.com/gardenlinux/gardenlinux/blob/2d5b1d67a52c1a7ce542b5ff3e32b89b5bf44ab8/tests/integration/gcp.py#L244-L258?

tests/tools/clean_orphans_gcp.py
```diff
@@ -0,0 +1,234 @@
...
+parser.add_argument('-z', '--zone', metavar='zone', required=False, dest='zone', default='europe-west1-d',
+                    help='Default value is "europe-west1-d"')
+parser.add_argument('-a', '--auth', metavar='service_account.json', required=False, dest='auth_path', default='',
+                    help='Path to JSON file that holds the service account login credentials')
+args = parser.parse_args()
+
+if os.path.exists(args.auth_path):
+    credentials = google.oauth2.service_account.Credentials.from_service_account_file(args.auth_path)
+else:
+    credentials = None
```
gyptazy, line 27: When falling back to default / well-known paths, we should validate for missing credentials and/or catch `google.auth.exceptions.DefaultCredentialsError`.

gyptazy (approved):
LGTM

tests/tools/clean_orphans_gcp.py
```diff
@@ -0,0 +1,239 @@
...
+    instance_client = compute.InstancesClient(credentials=credentials)
+except google.auth.exceptions.DefaultCredentialsError as e:
+    print(f"No credentials found. {e}\n\n")
+    parser.print_help()
+    sys.exit(os.EX_CONFIG)
+
+
+def wait_for_operation(operation, name):
+    print(f"Waiting for deletion of resource {name} to complete...")
+    result = operation.result(timeout=60)
```
MrBatschner, line 44: That 60 seconds timeout proved to be a bit too short during my tests - especially when it comes to tearing down VM instances. Maybe we can extend that to 5 Minutes and/or make it configurable...

MrBatschner (changes requested):
See my comment above about the timeout, the rest is really nice and will help a lot with testing in the future. 👍

MrBatschner (approved):
/lgtm

PR #1463, opened 2022-12-06
fix(pytest): Do not run orphaned tests on `GithubActionRunner`

### LINKED ISSUE #1461: `Pytest`: Fix `aws,GithubActionRunner` orphaned test
**What happened:**

The Pytest for orphaned packages will fail on `aws,GithubActionRunner`:
```
>       assert not pkgs_found, f"Found orphaned packages: {pkgs_found}"
E       AssertionError: Found orphaned packages: ['containernetworking-plugins']
```

gyptazy:
Do not run orphaned tests on `GithubActionRunner`

/kind bug
/area os
/os garden-linux

Do not run orphaned tests on `GithubActionRunner`

Fixes #1461

**Release note**:

```feature user
NONE
```


features/base/test/test_orphaned.py
```diff
@@ -3,5 +3,5 @@
 # Execute the orphaned check only on chroot
 # since it must install some packages that
 # may not be possible on running machines.
-def test_orphaned(client, chroot, non_container):
+def test_orphaned(client, chroot, non_container, non_feature_githubActionRunner):
```
nanory, line 6: Shouldn't we analyze, where the `containernetworking-plugins` is coming from and understand why it is considered to be orphaned or at least mark this package, so that other orphaned packages can still be detected instead of completely disabling it?

Check this for example:
* https://github.com/gardenlinux/gardenlinux/blob/main/features/base/exec.config#L17-L27

nanory (changes requested):

features/base/test/test_orphaned.py
gyptazy: IMHO: I don't think we should invest any further efforts for testing orphaned pkgs within the `githubActionRunner` feature since it'll never be used by anyone except of the pipeline. And at this point there'll be many upcoming changes in the near future that could lead into needed changes and break the pipeline again. What do you think @gehoern ?

nkraetzschmar (approved):

PR #1464, opened 2022-12-06
fix(buildpipeline): Avoid error message if cache file is not present

### LINKED ISSUE #1457: Fix build pipeline error
**What happened:**
Running the build pipeline will result in error messages that should be avoided:
```
head: cannot open 'make_targets.cache' for reading: No such file or directory
```

gyptazy:
fix(buildpipeline): Avoid error message if cache file is not present

/kind bug
/area os
/os garden-linux

Avoid error message if cache file is not present

Fixes #1457

**Release note**:

```feature user
NONE
```


bin/gen_make_targets
```diff
@@ -3,7 +3,12 @@
 set -Eeufo pipefail
 
 hash="$(sha256sum flavours.yaml | cut -c -64)"
-check="$(head -n 1 make_targets.cache || true)"
+
+if [ -f "make_targets.cache" ]; then
+	check="$(head -n 1 make_targets.cache)"
+else
+	check=true
```
Vincinator, line 10: The value `true` can be confusing  her. Otherwise looks good to me

gyptazy: Hey @Vincinator,
thanks for your feedback. How'd u like to name it? I think this represents that a "check" should be performed and will be fine in L13 by `if [ "$hash" != "$check" ]; then`.

Vincinator: Actually, you are right. If the variable name is `check` then the value `true` actually does make sense. 

I would go for a radical approach for this unimportant detail and rename the variable.
e.g.:
`cached_hash=none`  would be clearer for me.

Vincinator (approved):
Thanks!

PR #1468, opened 2022-12-08
enhancement(flavours): Add container and sort ASC yaml file

gyptazy:
Add container and sort ASC yaml file

/kind enhancement
/area os
/os garden-linux

Add container and sort ASC yaml file

We should add the `container` platform and "sort" the yaml file.

**Release note**:

```feature user
NONE
```

### DIFF
--- a/flavours.yaml
+++ b/flavours.yaml
@@ -1,4 +1,15 @@
 targets:
+  - name: ali
+  - name: aws
+    modifiers:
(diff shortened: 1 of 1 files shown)


Vincinator (approved):
We will need to change the structure soon, but as you suggested - lets include it already so we have it also in the new structure.

Issue #1439, opened 2022-11-25
`Pytest`: Change test platform for test `garden-feat`


gyptazy:
### General
In Garden Linux, Pytest is used for all unit-, platform- and integration tests. Currently, the unit test for `garden-feat` (located in base/test/) uses the `chroot` platform. While this works it may be uncommon to use a remote platform for running local tests. We might change this to the `local` platform as soon as this is present. (Depends on #1437).

chrinorse:
Solved with PR https://github.com/gardenlinux/gardenlinux/pull/1469

Issue #1438, opened 2022-11-25
`Pytest`: Create unit test for feature `_oci`


gyptazy:
### General
In Garden Linux, Pytest is used for all unit-, platform- and integration tests. Currently, we have a dedicated fixture for each major platform that can be accessed by a common ssh client object. However, when #1437 got merged, we may integrate a new test for the flag feature `_oci` which may use the `local` fixture.

PR #1469, opened 2022-12-08
extend local test platform

### LINKED ISSUE #1437: `Pytest`: Create new platform `local`
### General
In Garden Linux, Pytest is used for all unit-, platform- and integration tests. Currently, we have a dedicated fixture for each major platform that can be accessed by a common ssh client object. However, we may have additional tests on a local base where we want to test Garden Linux related scripts (e.g. `garden-feat`, etc.). Within this case, we do not need a hyperscaler related target platform and may run this locally. As a result, we should create a new generic `local` platform which does not use the common ssh client object. Beside this, it should still give some configuration options within the Pytest configuration for basic options like paths etc but should definitely not include test specific variables.

chrinorse:

/kind enhancement
/area os
/os garden-linux

extends the local test platform to be able to configure and run multiple local test with the same configuration file, excludes tests not suited for local execution.
Fixes #1437  #1438 #1439 

**Release note**:

```feature user

```


nanory (changes requested):

tests/local/test_garden_feat.py
```diff
@@ -106,5 +106,5 @@
 )
 
 
-def test_garden_feat(client, test_case, config, input_features, output_features, chroot):
+def test_garden_feat(client, test_case, config, input_features, output_features, local):
```
nanory, line 109: The garden-feat test doesn't need the `client` parameter anymore, no?

nanory (changes requested):

nanory (approved):

PR #1488, opened 2022-12-16
Force permissions for gardenlinux.asc

5kt:
/kind bug
/area os
/os garden-linux

/etc/apt/trusted.gpg.d/gardenlinux.asc will have whatever permissions the file has on the build system.
This is usually problematic when cloning the repo on machines with strict umasks. 
So we should force 644 as permissions for the file (force root:root as owner as well, not really needed, but just to be clearer).

Fixes #

### DIFF
--- a/bin/garden-init
+++ b/bin/garden-init
@@ -188,7 +188,7 @@ fi
 if [ -z "$nonDebian" ]; then
 
 	# Copy Garden Linux repository key
-	cp "$keyringPlain" "$targetDir/etc/apt/trusted.gpg.d/gardenlinux.asc"
+	install -o root -g root -m 644 "$keyringPlain" "$targetDir/etc/apt/trusted.gpg.d/gardenlinux.asc"	
 
 	# Add repositories to Garden Linux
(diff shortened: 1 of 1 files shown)


gyptazy (approved):
Hey @5kt,

thanks, lgtm

bin/garden-init
```diff
@@ -189,6 +189,8 @@ if [ -z "$nonDebian" ]; then
 
 	# Copy Garden Linux repository key
 	cp "$keyringPlain" "$targetDir/etc/apt/trusted.gpg.d/gardenlinux.asc"
```
waldiTM, line 191: You might want to use the usual `install` tool to do that.

```suggestion
	install -o root -g root -m 644 "$keyringPlain" "$targetDir/etc/apt/trusted.gpg.d/gardenlinux.asc"
```

Or install the file via the existing `file.include` mechanism,

gyptazy: Good point, we should use `install` instead. However, using the feature system could be too late within the overall process. We should keep it here.

PR #1491, opened 2022-12-20
fix(feature): Fix perl version in `base` feature

gyptazy:
fix(feature): Fix perl version in `base` feature

/kind bug
/area os
/os garden-linux

Fix perl version in `base` feature that avoids building GL.

**Release note**:

```feature user
NONE
```

### DIFF
--- a/features/base/exec.config
+++ b/features/base/exec.config
@@ -24,6 +24,6 @@ apt-mark manual \
   libfile-find-rule-perl \
   libgdbm-compat4 \
   libnumber-compare-perl \
-  libperl5.34 \
+  "libperl5.*" \
   libtext-glob-perl \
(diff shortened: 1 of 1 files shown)


features/base/exec.config
```diff
@@ -24,6 +24,6 @@ apt-mark manual \
   libfile-find-rule-perl \
   libgdbm-compat4 \
   libnumber-compare-perl \
-  libperl5.34 \
+  libperl5.36 \
```
waldiTM, line 27: This should also work:
```suggestion
  "libperl5.*" \
```

gyptazy: thanks for the useful hint!

gehoern (approved):

PR #1493, opened 2022-12-20
feature(vmware): switch ovf transport to vmware tools

defo89:

/kind enhancement
/area os
/os garden-linux

**What this PR does / why we need it**: this is required for ignition to read data from vApp properties and not use iso by default.

From the [documentation](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.vm_admin.doc/GUID-14A25C25-A6F8-4818-9D0C-C0247C0672FA.html):
```
If you select ISO image, an ISO image that contains the OVF template information is mounted in the CD-ROM drive.

If you select VMware Tools, the VMware Tools guestInfo.ovfEnv variable is initialized with the OVF environment document.
```

Currently OVA file produced by `make vmware` would expect that template config comes from a mounted ISO image. I am not sure if this is what cloud-init requires.

In the case of ignition, one would usually provide ignition config data via vApp properties (when creating a VM from OVA). But for this to happen, the ovf transport has to be set to `com.vmware.guestInfo`. 

image

### DIFF
--- a/features/vmware/vmware.ovf.template
+++ b/features/vmware/vmware.ovf.template
@@ -64,7 +64,7 @@
       </Property>
     </ProductSection>
 
-    <VirtualHardwareSection ovf:transport="iso">
+    <VirtualHardwareSection ovf:transport="com.vmware.guestInfo">
       <Info>Virtual hardware requirements</Info>
       <System>
         <vssd:ElementName>Virtual Hardware Family</vssd:ElementName>


marwinski (changes requested):
While I generally welcome ignition I am not sure whether this will have an impact on our Gardener on vSphere use case (that still uses cloud-init).

gehoern (approved):

marwinski:
@briantopping: can you judge this change. This is the Garden Linux flavour for vSphere

briantopping:
@marwinski: I can't. Maybe @tuxgoose can (or he may know someone on this end who can...)

marwinski:
Hi @briantopping, can you find an VMware expert in your organisation in order to get a verdict on this PR. Thanks

defo89:
I have updated description with more context.

tuxgoose:
@marwinski Folks I have begun testing the build I will relay the impact.

briantopping:
> @briantopping: can you judge this change. This is the Garden Linux flavour for vSphere

My final judgement is this is the correct change, should be applied, and parties that are affected should update accordingly.

PR #1504, opened 2023-01-19
feat(azure): add waagent again

Vincinator:
Adding waagent again, because it seems like waagent is a requirement for azure.

### DIFF
--- a/features/azure/file.include/etc/waagent.conf
+++ b/features/azure/file.include/etc/waagent.conf
@@ -29,7 +29,7 @@ ResourceDisk.SwapSizeMB=0
 ResourceDisk.MountOptions=None
 
 # Enable verbose logging (y|n)
-Logs.Verbose=n
+Logs.Verbose=y
 
 # Enable Console logging, default is y
 # Logs.Console=y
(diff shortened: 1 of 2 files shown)


features/azure/file.include/etc/waagent.conf
```diff
@@ -0,0 +1,87 @@
...
+
+# Size of the swapfile.
+ResourceDisk.SwapSizeMB=0
+
+# Comma-separated list of mount options. See mount(8) for valid options.
+ResourceDisk.MountOptions=None
+
+# Enable verbose logging (y|n)
+Logs.Verbose=n
+
```
MrBatschner, line 33: Used to be `y` in Garden Linux 576, no idea if we should flood the logs or not... probably not.

features/azure/file.include/etc/waagent.conf
```diff
@@ -0,0 +1,87 @@
...
+# Root device timeout in seconds.
+OS.RootDeviceScsiTimeout=300
+
+# If "None", the system default version is used.
+OS.OpensslPath=None
+
+# Set the SSH ClientAliveInterval
+# OS.SshClientAliveInterval=180
+
```
MrBatschner, line 54: Uncomment, Azure requires this setting and will deyn the image for certification is this is not set. Alternatively, we can put it in `sshd_config`.

Vincinator:
Thanks @MrBatschner, added your feedback with the last commit

PR #1509, opened 2023-01-23
make selinux an (optional) feature

### LINKED ISSUE #1508: Make SELinux an (option) feature flag
Make SELinux support optional as an feature flag. 

**Why is this needed**:
SELinux needs extended attributes filesystem, and as a building host, SELinux must be disabled so the security labels created during build times don't clash with active selinux policies.
We discovered that our build system has SELinux enabled by default with clashes with the `makepart` portion of the build system, which unconditionally set selinux attributes for the disk image files.

notandy:

/kind enhancement
/area os
/os garden-linux

This commit adds a feature flag `_selinux` that can be disabled for individual builds.
the `server` feature which is inherited by a few targets enables `_selinux` by default.
Disabling `_selinux` will also disable the selinux attribute tagging during partition build time.

This commit also cleanups the various hardcoded selinux `setfiles`.

Fixes #1508

**Release note**:

```feature user

```


features/_selinux/test/test_kernel_cmdline.py
```diff
@@ -0,0 +1,14 @@
...
+
+@pytest.mark.parametrize(
+    "file,args",
+    [
+        ("/etc/kernel/cmdline", "security=selinux"),
+    ]
+)
+
+
+def test_kernel_cmdline(client, file, args):
```
nkraetzschmar, line 13: please skip this test for githubActionRunner and firecracker feature by changing this to

```python
def test_kernel_cmdline(client, file, args, non_feature_githubActionRunner, non_feature_firecracker):
```

and adding a `non_feature_firecracker` pytest fixture in `tests/conftest.py`

```python
@pytest.fixture
def non_feature_firecracker(testconfig):
    features = testconfig.get("features", [])
    if "firecracker" in features:
        pytest.skip('test is not supported on firecracker')
```

features/firecracker/info.yaml
```diff
@@ -3,3 +3,4 @@ type: platform
 features:
   include:
     - server
+  exclude: [ _selinux ]
```
nkraetzschmar, line 6: exclude does not properly work yet, just skip the test instead

features/githubActionRunner/info.yaml
```diff
@@ -1,2 +1,4 @@
 description: GitHub action self hosted runner
 type: element
+features:
+  exclude: [ _selinux ]
```
nkraetzschmar, line 4: see above

nkraetzschmar (changes requested):

features/firecracker/info.yaml
notandy: Thanks, jeah just noticed it doesn't work

features/githubActionRunner/file.include/etc/kernel/cmdline.d/99-no-lsm.cfg
```diff
@@ -1 +0,0 @@
-CMDLINE_LINUX="$(echo "$CMDLINE_LINUX" | tr ' ' '\n' | grep -v '^security=' | tr '\n' ' ' | sed -e 's/ $//')"
```
nkraetzschmar, line 1: please keep this file for now, until we get exclude features to work correctly

nkraetzschmar (changes requested):

features/githubActionRunner/file.include/etc/kernel/cmdline.d/99-no-lsm.cfg
notandy: true, forget to re-include it again. thanks.

nkraetzschmar (approved):

PR #1510, opened 2023-01-24
Remove erroneous pkg.exclude from main feature dir

5kt:
There is a leftover pkg.exclude in the main features directory and it should be removed.


5kt:
This is actually needed, closing.

PR #1517, opened 2023-02-01
fix: test container: install python3-novaclient

nkraetzschmar:

/kind bug
/area os
/os garden-linux

install python3-novaclient in test container


nkraetzschmar:
replaced by #1518

PR #1522, opened 2023-02-07
feat: garden-feat use venv

nkraetzschmar:

/kind enhancement
/area os
/os garden-linux

add wrapper to garden-feat to run itself in a python venv to ensure all dependencies are available

**TODO**:

- [ ] update github action runner image to include `python3-venv` instead of `python3-networkx`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ __pycache__
 /tests/*.log
 /tests/*.xml
 make_targets.cache
+bin/garden-feat.venv
--- a/bin/garden-feat
+++ b/bin/garden-feat
@@ -1,3 +1,21 @@
+#!/usr/bin/env bash
+
+set -eufo pipefail
+
+exec 3>&1
+exec 4>&2
+exec 1>/dev/null
+exec 2>/dev/null
+
+venv="$(dirname "$BASH_SOURCE")/garden-feat.venv"
+python3 -m venv "$venv"
+source "$venv/bin/activate"
+pip install pyyaml networkx
+
+exec 1>&3
+exec 1>&4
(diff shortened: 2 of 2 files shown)


gyptazy:
Hey @nkraetzschmar,

just some notes bc `garden-feat` is one of the core parts during the build process and may also be executed out of a container on a user's system.

The [Makefile](https://github.com/gardenlinux/gardenlinux/blob/main/Makefile#L21) generates the build targets by invoking [gen_make_targets](https://github.com/gardenlinux/gardenlinux/blob/main/bin/gen_make_targets#L16) which uses `garden-feat`.

Therefore, we should ensure:
 * Ensure the environment is prepared (at least python-pip is currently not required regarding the docs and may fail [on light installs & macOS])
 * Not silently install packages on a users system by default (`pip install pyyaml networkx`, but they're already mentioned for `Debian` and `CentOS`)
 * Maybe add a new param to run without venv (e.g. `garden-feat --without-venv`) which might be helpful in some corner cases
 * `pip` might also be `pip3` on some systems (e.g. `macOS`)

IMHO, I think this could be solved by:
 * Ensure `pip` (or `pip3`) package is present
 * Adding the required pip modules to `macOS` quick start guide: `pip3 install pyyaml networkx` that ensures users are aware of the packages before starting
 * Validate if `pip` or `pip3` should be used in `garden-feat`
 * Maybe add a new param to run without venv (e.g. `garden-feat --without-venv`)

Cheers,
gyptazy

Vincinator:
This is also relevant for the nightly build pipeline. https://github.com/gardenlinux/gardenlinux/pull/1602 
@nkraetzschmar what do you think, can we close this PR?

PR #1547, opened 2023-02-23
Fix: use venv for container

Vincinator:
***Why do we need this?*** 
* A new debian python update causes pip to refuse installation of packages if environment is marked as externally managed. This issue breaks our dev and nightly
* https://github.com/pypa/pip/issues/11381
* alternatively, we can delete the EXTERNALLY-MANAGED files

***Reproduce issue***
```
make --directory=container build-cert
```

Before this PR we run into this:
```
STEP 10: RUN pip install git+https://github.com/awslabs/python-uefivars
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.

### DIFF
--- a/container/base-test/Dockerfile
+++ b/container/base-test/Dockerfile
@@ -24,6 +24,7 @@ RUN echo "deb http://deb.debian.org/debian testing main contrib non-free" > /etc
           gnupg \
           python3-pip \
           python3-pytest \
+          python3-venv \
           vim \
           procps \
           openssh-client \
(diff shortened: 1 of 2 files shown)


nkraetzschmar (approved):
/lgtm

PR #1556, opened 2023-03-13
remove iptables-nft default relation

gehoern:
error in rel934.

we have users that complained about mixed firewall setup. so we are reverting to iptables-legacy for now.

### DIFF
--- a/features/gardener/exec.config
+++ b/features/gardener/exec.config
@@ -8,3 +8,6 @@ chmod u-s /sbin/mount.nfs /sbin/mount.cifs
 # one it uses
 systemctl disable docker
 systemctl disable containerd
+
+update-alternatives --set iptables "/usr/sbin/iptables-legacy" > /dev/null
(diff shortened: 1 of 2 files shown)


marwinski (changes requested):
We had this:

```
update-alternatives --set iptables /usr/sbin/iptables-legacy
update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
update-alternatives --set arptables /usr/sbin/arptables-legacy
update-alternatives --set ebtables /usr/sbin/ebtables-legacy
```

Not sure whether this is really needed but we should probably stick to something that is known to work

MrBatschner:
> 

There are binaries `arptables-legacy` and `ebtables-legacy` in Garden Linux 934 and thus they are obviously also not registered as alternatives for `arptables` and `ebtables`.

Vincinator (approved):

marwinski (approved):

PR #1559, opened 2023-03-13
arptables and ebtables missing for legacy

gehoern:
legacy fix for iptables-nft


Vincinator (approved):

MrBatschner:
cherry picked into PR #1556 as this should be in one single PR
/close in favour of PR #1556

PR #1572, opened 2023-03-21
New Logging feature

gehoern:
Logging is now a standalone feature and supports:

- remote logging via systemd-units and the journalctl format
- but also rsyslog for remote logging
  - even the old way of logging to /var/log can be enabled via rsyslog

- examples are in

improvements:
- logging is now guaranteed sealed (was before implicit)
- there are official auditd rules now integrated


Vincinator (approved):
/lgtm

PR #1614, opened 2023-03-31
[cloud] console order in in 00-default.cfg

defo89:

/kind enhancement
/area os
/os garden-linux

When testing ignition on vmware image I was not getting any kernel output in the console - was not able to see why ignition run was failing. This changed order fixes the issue for vmware environment. Kindly verify the behaviour in other environments.

I noticed that [metal](https://github.com/gardenlinux/gardenlinux/blob/main/features/metal/file.include/etc/kernel/cmdline.d/00-default.cfg) uses the same configuration.

Fixes #

Here is how the output looked before (not able to do anything here):
Screenshot 2023-03-30 at 17 53 43

And this is how it looks after. I'm now able to see a failed igntion run, enter emergency shell, etc.
Screenshot 2023-03-31 at 14 54 59


MrBatschner (approved):
It is even in the documentation of all the cloud providers and still, we did it wrong. D'oh! Thank you for that.
/lgtm

PR #1621, opened 2023-04-03
fix: serial console default on cloud provider

Vincinator:
See #1614 for the reason why vmware needs a different order
Other cloud flavors should keep the serial console as default, also the start-vm feature does not work if serial console is not default.


Vincinator:
Need to check/test this

5kt (approved):

PR #1624, opened 2023-04-04
fix: cloud config for azureuser

Vincinator:
* specify azureuser in azure feature via cloud-init
* tests can safely assume azureuser can execute user tests


MrBatschner (approved):
/lgtm

PR #1631, opened 2023-04-06
fix: increase nslookup test timeout for azure

Vincinator:
/kind test
/area os
/os garden-linux

Azure requires the OS to register its hostname against the azure DNS.

We have a test that checks if `nslookup $(hostname)`  does not return an error exit code 
**AND** the lookup was faster than 2 seconds.

- Increase timeout of azure specific nslookup test to 10 seconds:
    * hostname is registered on the azure dns during boot
    * registration is performed by cloud-init in our case
    * The timeout of 2 seconds can be too small, because the hostname was just registered on the
azure DNS.
    * garden linux responsibility is to register it, but garden linux has no direct impact on
dns performance
- Added description to test docstring

* running `nslookup $(hostname)` on a booted garden linux vm on azure returns a valid answer.
* tested with `make azure-gardener_prod` generated image of current main f130ddb

### DIFF
--- a/features/base/test/test_network.py
+++ b/features/base/test/test_network.py
@@ -4,14 +4,19 @@ from helper.sshclient import RemoteClient
 
 
 def test_hostname_azure(client, azure):
-    """ Test for valid hostname on azure platform """
+    """ Test for valid hostname on azure platform. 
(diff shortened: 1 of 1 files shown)


MrBatschner (approved):
/lgtm

PR #1632, opened 2023-04-06
fix(tests): workaround for azure sudo -l -U check

Vincinator:
/kind test
/area os
/os garden-linux

The user tests are performed by the azureuser
on azure. In theory, azureuser should have full admin priviledges and, just like root, no requirement
to enter a password for sudo.

However:
* `sudo -l -U` requires a password
* `sudo -s sudo -l -U` works passwordless
* this commit is a workaround so the test can do the actual check if any user on the image
has un-intendet sudo priviledges.


MrBatschner (approved):
/lgtm

PR #1638, opened 2023-04-13
fix(azure): cloud-init configuration

### LINKED ISSUE #1633: Startup time for azure in platform tests sometimes takes too long
**What happened:**
Boot time for azure sometimes is over the threshold we have set in our tests.
The variance for the time results are so high, that sometimes the threshold is below 40, and sometimes over 50. 
This high variance in observed boot times is caused presumably by the response times of azure backends during VM registration. Yet to be proven. 

`E       AssertionError: startup time in user space too long: 52.17seconds but only 40 tolerated.
[600](https://github.com/gardenlinux/gardenlinux/actions/runs/4649107880/jobs/8227840502#step:11:601)
E       assert 52.17 < 40`

* https://github.com/gardenlinux/gardenlinux/actions/runs/4664429621/jobs/8257120249#step:11:570
* https://github.com/gardenlinux/gardenlinux/actions/runs/4649107880/jobs/8227840502#step:11:576

Vincinator:

/kind enhancement
/area os
/os garden-linux

Cloud init on azure was not optimal, some modules were redundant to our garden linux configuration (ntp).

* Disable network via config
* Configure datasource to speed up boot process
* Disable the following cloud-init modules
```
<  - migrator
<  - growpart
<  - resizefs
<  - disk_setup
<  - mounts
<  - rsyslog
<  - snap
<  - keyboard
<  - grub-dpkg
<  - apt-configure
<  - ntp
<  - disable-ec2-metadata
<  - chef
<  - power-state-change
```

Fixes #1633 
This PR should be used instead of increasing the time thresholds (e.g. via #1634)

Credits for this PR also to @MrBatschner and @marwinski.

**Release note**:

```feature user

```

### DIFF
--- a/features/azure/exec.config
+++ b/features/azure/exec.config
@@ -1,5 +1,3 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-# disable cloud-init network configuration
-echo -e "network:\n  config: disabled\n" >> /etc/cloud/cloud.cfg.d/00_azure.cfg
--- a/features/azure/file.include/etc/cloud/cloud.cfg.d/02_azure-datasources.cfg
+++ b/features/azure/file.include/etc/cloud/cloud.cfg.d/02_azure-datasources.cfg
@@ -1,0 +1,1 @@
+datasource_list: [ Azure ]
--- a/features/azure/file.include/etc/cloud/cloud.cfg.d/03_azure-modules.cfg
+++ b/features/azure/file.include/etc/cloud/cloud.cfg.d/03_azure-modules.cfg
@@ -7,6 +7,7 @@ cloud_init_modules:
  - update_etc_hosts
  - ca-certs
  - users-groups
+ - ssh
 
 cloud_config_modules:
  - ssh-import-id
--- a/features/azure/file.include/etc/cloud/cloud.cfg.d/03_azure-modules.cfg
+++ b/features/azure/file.include/etc/cloud/cloud.cfg.d/03_azure-modules.cfg
@@ -1,0 +1,42 @@
+cloud_init_modules:
+ - seed_random
+ - bootcmd
+ - write-files
+ - set_hostname
+ - update_hostname
+ - update_etc_hosts
+ - ca-certs
+ - users-groups
+
+cloud_config_modules:
+ - ssh-import-id
+ - locale
+ - set-passwords
+ - apt-pipelining
+ - timezone
+ - runcmd
+ - byobu
+
+cloud_final_modules:
+ - package-update-upgrade-install
+ - fan
+ - landscape
+ - lxd
+ - write-files-deferred
+ - puppet
+ - mcollective
+ - salt-minion
+ - reset_rmc
+ - refresh_rmc_and_interface
+ - rightscale_userdata
+ - scripts-vendor
+ - scripts-per-once
+ - scripts-per-boot
+ - scripts-per-instance
+ - scripts-user
+ - ssh-authkey-fingerprints
+ - keys-to-console
+ - install-hotplug
+ - phone-home
+ - final-message
+ - power-state-change
--- a/features/azure/file.include/etc/cloud/cloud.cfg.d/99_disable-network-config.cfg
+++ b/features/azure/file.include/etc/cloud/cloud.cfg.d/99_disable-network-config.cfg
@@ -1,0 +1,1 @@
+network: {config: disabled}


Vincinator:
I am currently testing this PR. Please feel free to also test and share

Vincinator:
This PR does not seem to fix the issue with test_loadavg.

I have build an `azur-gardener_prod` image, and run the integration tests on azure. Got two failing tests:

```
   def test_loadavg(client, non_kvm, non_chroot):
        """ This test does not produce any load. Make sure no other process does """
        (exit_code, output, error) = client.execute_command("cat /proc/loadavg")
        assert exit_code == 0, f"Expected to be able to show contents of /proc/loadavg"
        load =  float(output.split(" ")[1])
>       assert load  < 0.5, f"Expected load to be less than 0.5 but is {load}"
E       AssertionError: Expected load to be less than 0.5 but is 0.82
E       assert 0.82 < 0.5
```


```
    def test_clock(client):
        """ Test clock skew """
        (exit_code, output, error) = client.execute_command("date '+%s'")
        local_seconds = time.time()
        remote_seconds = int(output)
>       assert (
            abs(local_seconds - remote_seconds) < 5
        ), "clock skew should be less than 5 seconds. Local time is %s and remote time is %s" % (local_seconds, remote_seconds)
E       AssertionError: clock skew should be less than 5 seconds. Local time is 1682494738.2738435 and remote time is 1682494701
E       assert 37.27384352684021 < 5
E        +  where 37.27384352684021 = abs((1682494738.2738435 - 1682494701))
```

Vincinator:
This loadavg test is failing for our azure images in our nightlies (for a while now).
I am trying to understand why we test this, but also why only azure fails this test.

*What does this test do?*

This test reads the second value returned by `/proc/loadavg`, and asserts that it is smaller than `0.5`.
If it is greater than 0.5, the test fails.

*What does a `loadavg` number mean?*

`loadavg` of 0 means that the system is idle. For each process entering in one of the states _running_, _queued_, or _uninterruptible sleep_, the load is increased. If a process transitions out of these states, then the load is decremented again.
Thus, if we have an `loadavg` of 1 over the past 5 minutes, then on average 1 process is in one of the states:
* _running_
* _queued_
* _uninterruptible sleep_

Well this is a simplification. In reality the sum is an exponentially decaying average of the processes in running, queued, or uninterruptible sleep. Also calculation for a `NO_HZ` kernel (which we have) is a bit more complicated. But in the end the simplification is good enough for a valid intuition. (in case you are curious and don't accept a simplification: https://elixir.bootlin.com/linux/latest/source/kernel/sched/loadavg.c#L162)

*Why is a value above 0.5 considered too high?*
A value above 0.5 is too high if we expected the system to idle a lot in the the recent past of the time window (last 5 minutes).

Now this is where I got confused. The time we read the `loadavg` value of the past 5 minutes include multiple tasks:
* boot process (including systemd, cloud init)
    * We expect the available CPUs to be fully utilised during boot. While on the other hand we want to reduced the time processes are blocked by i/o. 
* Test overhead
    * (network) initialise first ssh connection between tester machine and tested VM [see](https://github.com/gardenlinux/gardenlinux/blob/main/tests/helper/sshclient.py#L133). initialisation time is variable depending on how fast the tester can connect to the tested VM
    * (network) SSH connections to send commands to execute and read output values
    * (local process) execution test commands on that machine

Multiple problems here.
1. Why do we expect the system to idle if we put load on the system? 
2. The start of `test_loadavg` is variable depending on network. The `loadavg` 5 minute window is fixed, while we do not know what is included in that 5 minutes window. 
3. `loadavg` does not differentiate between running processes and processes blocked by I/O. 
4. The VMs we use to run the tests on have different amounts of CPUs per cloud provider: (azure: Standard_D4_v4  `4 cpus`, aws: t3.micro `2 cpus`, gcp: n1-standard-2 `2 cpus`)
5. Systemd and cloud-init both have components that can be executed concurrently. 
6. The test does not always fail. Thus, we can not trust this test setup if it is green

A test that just checks if the system is booting fast enough is implemented with [test_startup_time](https://github.com/gardenlinux/gardenlinux/blob/main/features/base/test/test_basics.py#L49).


*What can cause a high `loadavg` during a given time period?*
* Obviously, a high demand on computing time during that time window. 
* A process that is in uninterruptible sleep because of a blocking i/o operation. And I think, this is where it gets interesting. Remember, we are talking about having failing tests on azure: loadavg and also boot time taking too long. 


*My Conclusion*
I think we should remove `test_loadavg`,  or create a better setup and document what this test is actually testing. Is it testing boot load? Is it testing idle load?

Vincinator:
Feel free to re-open if required.

PR #1655, opened 2023-05-09
fix(ci): turn off auto release note gen

Vincinator:
* the github action runner require python3-click installed as dependency.
* Open PR: https://github.com/gardenlinux/gardenlinux/pull/1654
* disable auto generation until dependency is available


MrBatschner (approved):
/lgtm

PR #1665, opened 2023-05-17
chore: Gitignore uefi vars files

fwilhe:
Those binary files are created by the 'start-vm' script and most likely should not be committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ __pycache__
 /tests/*.log
 /tests/*.xml
 make_targets.cache
+/bin/*.vars


Vincinator:
Thanks, good catch. Appreciate that you directly open a PR. :) 

looks good to me. approved.


Now, let me blabber a bit of context and share a friendly reflection regarding the practice of prefixing the name of a constant with "curr_".

[bin/start-vm:494](https://github.com/gardenlinux/gardenlinux/blob/c366ca18b8ce61c200e2eff0acd84b83ee74a261/bin/start-vm#L494):
```bash
        [ -e $CURR_DIR/$mac.vars ] || cp $uefi_vars $CURR_DIR/$mac.vars
```
This line is responsible for the leftover. Thus, you need to be on osx to reproduce this.

The constant `CURR_DIR` should be renamed to `THIS_FILE_DIR` or something less ambiguous than "CURR_DIR".
However, it is set in [bin/start-vm:6](https://github.com/gardenlinux/gardenlinux/blob/c366ca18b8ce61c200e2eff0acd84b83ee74a261/bin/start-vm#L6) to
```
readonly CURR_DIR="$(dirname "$(readlink -f "$BASH_SOURCE")")"
```

"curr" and "constant" do not work well together, does it? 🤨

Vincinator (approved):

PR #1669, opened 2023-05-26
hack(gl-search): fix bash subshell function export

Vincinator:
This PR fixes hack/gl-search.sh for osx users.

The issue was that osx bash does not load exported functions into sub shells automatically.
This commit puts the functions in a separate file, which is then sourced on demand.

@fwilhe Thanks for reporting!

### DIFF
--- a/hack/.gl-search-functions.sh
+++ b/hack/.gl-search-functions.sh
@@ -3,17 +3,28 @@
 
 function filter_package_info() {
   pkg=$1;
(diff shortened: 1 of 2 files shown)


fwilhe:
Thanks for looking into it, I can confirm that this resolves the issue I had 👍

Vincinator:
@fwilhe Thanks! Once this PR gets approved I will merge it

MrBatschner (approved):
/lgtm

PR #1670, opened 2023-05-30
test: increase timeout for GCP Image import

MrBatschner:

/kind test
/area os
/os garden-linux

Importing an image to GCP for testing might take longer than two minutes which used to be the max. timeout for this operation. This PR increases the timeout to 10 minutes.

**Release note**:

```feature user
GCP platform tests now have a longer timeout of 10 minutes for the image import.
```


Vincinator (approved):
looks good to me, thanks!

PR #1671, opened 2023-05-30
test: abort all tests if SSH connection fails

MrBatschner:

/kind test
/area os
/os garden-linux

Platform tests should fail during fixture setup if an SSH connection to a VM cannot be established within time. This removes tons of unnecessary log output if the SSH connection fails to establish.

**Release note**:

```feature user
Garden Linux platform tests will fail early if an SSH connection cannot be established after a default timeout of 5 minutes.
```


Vincinator (approved):
lgtm

PR #1672, opened 2023-05-30
fix(test): backport updated gcp integration test to rel 934

Vincinator:

/kind test
/area os
/os garden-linux

We need to cherry-pick a7ed52f into rel-934 branch.  
We changed the used api in the gcp platformtest setup in this commit 7b35ac4.

This PR brings the required changes from main to rel-934.

**Note for Reviewer**
I think we should keep the history and NOT squash this PR into a single commit. We added a bunch of commits via cherry-pick from main. History will be more meaningful if we keep the original commit messages.

### DIFF
--- a/tests/Pipfile
+++ b/tests/Pipfile
@@ -13,10 +13,8 @@ scp = "*"
 dacite = "*"
 # Google/GCP SDK packages
 google-auth = "*"
-google-cloud = "*"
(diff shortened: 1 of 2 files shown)


MrBatschner (approved):
Looks good so far and should help to keep the tests on the `rel-934` branch in line with the latest tests on `main`.

/lgtm

PR #1676, opened 2023-06-01
tests: improve resource leakage

MrBatschner:

/kind test
/area os
/os garden-linux

This PR introduces two small fixes to the platform tests which - under certain conditions - could leak resources in GCP and Azure:

- on GCP, the GCS bucket used for uploaded the image to test will be deleted immediately after the image is inserted into GCE (so far, we only deleted the image blob immediately after import but left the empty bucket)
- on Azure, the SSH key used for the test will be explicitely removed and a longer timeout is granted for deleting the resource group of the test 

**Release note**:

```feature user
The cleanup code for automatic Garden Linux platform tests was improved to prevent resource leakage.
```

### DIFF
--- a/tests/integration/azure.py
+++ b/tests/integration/azure.py
@@ -143,6 +143,16 @@ class AZURE:
                 'tags': self._tags
             }
         )
+    
+    def az_delete_ssh_key(self, name: str, force: bool = False):
+        key = self.az_get_ssh_key(name)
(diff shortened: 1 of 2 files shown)


Vincinator (approved):
Nothing to add. Looks good to me

gehoern (approved):

PR #1682, opened 2023-06-13
builder

nkraetzschmar:
use https://github.com/gardenlinux/builder build toolchain


Vincinator:
The `features/example` feature is not up to date, this PR should either remove the example in favour of the [gardenlinux/builder_example](https://github.com/gardenlinux/builder_example) or update the example. In case we do move the example to somewhere else, we should also update [gardenlinux/builder/README.md](https://github.com/gardenlinux/builder/blob/main/README.md#config-directory)

What is outdated?
- [gardenlinux/features/example/exec.post](https://github.com/gardenlinux/gardenlinux/tree/main/features/example/exec.post) comment is outdated
    - the comment that it is not executed in chroot is now false. The new builder does execute exec.post in chroot.
- [gardenlinux/features/example/exec.pre](https://github.com/gardenlinux/gardenlinux/tree/main/features/example/exec.pre)
    - is not implemented by the new builder
- new stage `exec.late` is not documented

PR #1713, opened 2023-07-21
feat(gh_actions): version tag for container images

Vincinator:
* The Garden Linux container will be used for Garden Linux Package builds. To build packages for supported versions, we need the corresponding package build container in the correct version. E.g. `1227` instead of `nightly`. 

* This commit uses `bin/garden-version now` instead of `bin/garden-version today` This will set the actual version, instead of the string "today".

* This commit adds version tags to the garden linux container

### DIFF
--- a/.github/workflows/publish_container.yml
+++ b/.github/workflows/publish_container.yml
@@ -4,7 +4,7 @@ on:
     inputs:
       version:
         type: string
-        default: today
+        default: now
       default_modifier:
         type: string
         default: ""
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_container.yml
```diff
@@ -4,7 +4,7 @@ on:
     inputs:
       version:
         type: string
-        default: today
+        default: now
```
nkraetzschmar, line 7: having a different default here than for `build.yml` is probably not a good idea. Otherwise using build and publish with default options would fail at the download artifact step.

.github/workflows/publish_container.yml
```diff
@@ -21,6 +21,9 @@ jobs:
         run: |
           bin/garden-version "${{ inputs.version }}" | tee VERSION
           git update-index --assume-unchanged VERSION
+      - name: get current garden linux version
+        run: |
+          echo "gl_version=$(cat VERSION)" | tee -a "$GITHUB_ENV"
```
nkraetzschmar, line 26: maybe we should use `./get_version` instead.... won't make  a difference with the set version step above, but if that step ever changes using `./get_version` should be more stable

.github/workflows/publish_container.yml
Vincinator: agree, but I thought `now` is actually used
https://github.com/gardenlinux/gardenlinux/blob/main/.github/workflows/nightly.yml#L17

Vincinator:
As agreed in the daily, we do not want version tags for unsupported versions

Issue #1709, opened 2023-07-17
start-vm: Is there any reason to exclude ubuntu from finding the uefi code?


fwilhe:
We are testing for the os [here](https://github.com/gardenlinux/gardenlinux/blob/ac4d82e3cea0df3bd63cc5ab291129c14af716fb/bin/start-vm#L442). Ubuntu does not get recognised as debian but has its own entry which is not tested here. In my ubuntu vm, the path for debian also works for ubuntu, so I guess we can just extend the if branch for "if debian or ubuntu".

Anything that I overlooked?

Vincinator:
fixed by https://github.com/gardenlinux/gardenlinux/pull/1714

PR #1714, opened 2023-07-26
start-vm: Make uefi boot default, fix logic for ubuntu

fwilhe:
It seems that uefi boot should be the default.
For compatibility, the uefi flag still exists but it is not needed.
A new 'legacy-bios' flag was added to disable uefi if needed.

Also add ubuntu to the conditional case for loading uefi code/vars for debian.
Without this change, running in uefi mode on an ubuntu host fails when paths are not configured by hand.

### DIFF
--- a/bin/start-vm
+++ b/bin/start-vm
@@ -458,6 +458,9 @@ qemu_opt_uefi () {
 
     elif [ "$arch" = arm64 ]; then
         # Always enable UEFI mode on aarch64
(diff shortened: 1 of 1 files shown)


Vincinator:
fixes #1709

fwilhe:
@Vincinator I've added the warning for aarch64 we discussed about.

From my perspective this should be ready for review.

Vincinator (approved):
/lgtm

PR #1733, opened 2023-08-16
refactor: generate release note

Vincinator:
* remove click as dependency. Click is nice, but having less dependencies in the pipeline is nicer.
* make commitish parameter required
* If image cannot be found, print helpful error message instead of aborting
* Added usage doc in top of script for now.

This is currently not part of the automated pipeline. A garden linux maintainer must execute this locally

```
# install python dependencies 
python3 -m venv venv
source venv/bin/activate
pip install boto3 pyyaml

# Execute the command (example for 934.10):
.github/workflows/generate_release_note.py generate --version 934.10 --commitish f057c9b 
```

### DIFF
--- a/.github/workflows/generate_release_note.py
+++ b/.github/workflows/generate_release_note.py
@@ -3,13 +3,13 @@
 # Generate Release Notes for a Garden Linux Release
 #
 
-import click
 import os
 import boto3
 import yaml
 import urllib.request
 import sys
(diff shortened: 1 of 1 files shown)


MrBatschner:
We should integrate the call somewhere here: https://github.com/gardenlinux/glci/blob/main/.ci/pipeline_definitions

MrBatschner (approved):

PR #1751, opened 2023-08-31
clean: remove garden-integration-test-config

Vincinator:
This script was required by the `build.sh`, which used this to setup the kvm/chroot environment.

The new builder has two separate `build` and `test` scripts now. 

`test` script [starts](https://github.com/gardenlinux/gardenlinux/blob/main/test#L50) the [test container](https://github.com/gardenlinux/gardenlinux/blob/main/tests/Containerfile) which runs this [init](https://github.com/gardenlinux/gardenlinux/blob/main/tests/init) to setup and trigger the tests. 

In this new process we do not need the `garden-integration-test-config` file anymore.


MrBatschner (approved):
/lgtm

Failing pipeline is unrelated to these changes.

PR #1756, opened 2023-09-07
Lima: Convert image to qcow2 for smaller download size

fwilhe:

/kind enhancement
/area os
/os garden-linux

Convert image files from raw to qcow2 in the lima feature.
As it is common for lima to download images via http, having raw images means a lot of useless traffic will happen so it makes sense to convert the images to qcow2 as they are smaller.

Fixes #

**Release note**:

```feature user

```


Vincinator (approved):
Tested - looks good to me

PR #1758, opened 2023-09-12
test: limit azure nsg to incoming ssh from test system

MrBatschner:

/kind test
/area os
/os garden-linux

The network security group that gets created for a platform test in Azure should not allow incoming SSH traffic from 0.0.0.0/0 but only from the machine that runs and controls the tests, thus increasing security.

**Release note**:

```feature user
The network security group for Azure base platform tests will only allow incoming SSH traffic from the test system.
```


Vincinator (approved):
nice, thank you.

PR #1762, opened 2023-09-22
adr: dists can never change retroactively

Vincinator:
This is the ADR for the decision to freeze apt distributions from repo.gardenlinux.io (e.g. `1270.0`).

### DIFF
--- a/docs/architecture/decisions/0002-dists-can-never-change-retroactively.md
+++ b/docs/architecture/decisions/0002-dists-can-never-change-retroactively.md
@@ -20,7 +20,7 @@ One of the mission statements of Garden Linux is to ensure reproducibility of Ga
 ## Decision
 
 With the exception of `next` distributions, all apt distributions must be immutable. 
-Any apt distribution used to bootstrap its corresponding Garden Linux version should never be altered retroactively.
+Any apt distribution used to bootstrap its corresponding Garden Linux version must never be altered retroactively.
 
 
 ## Consequences
--- a/docs/architecture/decisions/0002-dists-can-never-change-retroactively.md
+++ b/docs/architecture/decisions/0002-dists-can-never-change-retroactively.md
@@ -8,16 +8,28 @@ Accepted (for the GitHub package pipeline)
 
 ## Context
 
-The Garden Linux apt repository produces daily distributions. 
-These distributions are assembled through a dedicated package repository pipeline. 
-Garden Linux maintainers define the package specifications for upcoming apt distributions.
+Garden Linux uses the apt package manager and bootstraps via the `repo.gardenlinux.io` apt repository. 
+Every Garden Linux version corresponds directly to an apt distribution on this repository, sharing the same name.
+For instance, Garden Linux version `934.10` aligns with the `934.10` apt distribution.
+
+The major apt distributions are assembled daily automatically by a GitHub pipeline. 
+After the apt distribution (e.g. `1234.0`) is created, the respective Garden Linux version is build by another automated pipeline.
+
+One of the mission statements of Garden Linux is to ensure reproducibility of Garden Linux releases.
 
 ## Decision
 
-Apt distributions, identified by their version numbers, are immutable. They cannot be altered once established. Essentially, when a distribution is named with a specific version, its contents remain unchanged.
+With the exception of `next` distributions, all apt distributions must be immutable. 
+Any apt distribution used to bootstrap its corresponding Garden Linux version (e.g. `1234.0`) should never be altered retroactively.
+
 
 ## Consequences
 
 1. Introduction of the dynamic Garden Linux apt distribution named `next` 
 2. This `next` distribution is adaptable and can undergo changes at any given moment, ensuring it always encompasses the most recent package iterations.
 3. When creating a `major.0` release, the existing state of `next` is utilized. 
+
+## Notes
+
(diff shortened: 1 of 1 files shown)


docs/architecture/decisions/0002-dists-can-never-change-retroactively.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Status
+
+Accepted (for the GitHub package pipeline)
+
+## Context
+
+The Garden Linux apt repository produces daily distributions. 
+These distributions are assembled through a dedicated package repository pipeline. 
+Garden Linux maintainers define the package specifications for upcoming apt distributions.
```
fwilhe, line 13: > upcoming apt distributions.

What does 'distributions' mean in this case? Is it the same as above?

This might be debian lingo that I'm not too familiar with. In debian, a distro is something like 'bookworm' or 'trixie', right?

docs/architecture/decisions/0002-dists-can-never-change-retroactively.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Decision
+
+Apt distributions, identified by their version numbers, are immutable. They cannot be altered once established. Essentially, when a distribution is named with a specific version, its contents remain unchanged.
+
+## Consequences
+
+1. Introduction of the dynamic Garden Linux apt distribution named `next`
```
fwilhe, line 21: how is 'next' different from 'today'?

docs/architecture/decisions/0002-dists-can-never-change-retroactively.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Context
+
+The Garden Linux apt repository produces daily distributions. 
+These distributions are assembled through a dedicated package repository pipeline. 
+Garden Linux maintainers define the package specifications for upcoming apt distributions.
+
+## Decision
+
+Apt distributions, identified by their version numbers, are immutable. They cannot be altered once established. Essentially, when a distribution is named with a specific version, its contents remain unchanged.
```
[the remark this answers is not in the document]
Vincinator: > What does this mean? We have an automated process to make new versions each day, but here we are talking about versions we actually maintain/release, right?

We are talking here about every release, also the versions created each day. 
Once a `.0` is created we should not be able to modify it retroactively anymore. I


> In my own words:
What you mean here is that we consider the apt repo immutable (even when technically it is not), so when we have version 1234 no package in that directory changes and if we need to make updates (for security), we will copy the contents from 1234 into 1234.1 with the patched packages changed.
Is this summary correct?


Technically it must be immutable up to the level of our pipelines and other interfaces used to create apt distributions for repo.gardenlinux.io. 

Yes absolutely, the basis for a `1234.1` release is `1234.0`, plus the list of additional package version changes we want to include.

The `next` branch introduced with the proposed ADR is **not** used for patch releases. I will propose another ADR for a similar approach for patch releases. 

To outline it briefly, here is an example of the relevant process for creating a patch release:

1.  Packages candidates for the next patch release will be placed in the `1234-next` apt dist, which is writable and can change at any time, just as the `next` apt dist. 
2. We decide to release the 1234.2 minor version
3. We trigger a repo pipeline that assembles the apt distribution for `1234.2`, the pipeline uses `1234.1` as base, and includes the changes required specified in a custom config file (like the package yaml files we currently have e.g. [934.10.yaml ](https://gitlab.com/gardenlinux/gardenlinux-package-build/-/blob/b3013e418d1f78e4135755b4aa4c6213d1587dca/packages/934.10.yaml)  
4. we tag the respective commits on the `rel-1234` branch in the gardenlinux/gardenlinux repo
5. The tagged commit `1234.2` of  gardenlinux/gardenlinux repo will be used to build the release artifacts

This process makes sure that we enforce reproducibility.

docs/architecture/decisions/0002-dists-can-never-change-retroactively.md
Vincinator: today is like a soft link to the xyz.0 release. 

next is the development for "tomorrows" xyz.0 release

fwilhe:
Does anyone else want to leave review comments?

Had an offline discussion with @Vincinator about `next` and `today` and I think it makes sense to me now, not sure if I would understand it just by reading this. This is probably out of scope for this document.

I'm fine with merging the document but I'll wait for anyone else to look at it before officially approving.

Vincinator:
> Does anyone else want to leave review comments?
> 
> Had an offline discussion with @Vincinator about `next` and `today` and I think it makes sense to me now, not sure if I would understand it just by reading this. This is probably out of scope for this document.
> 
> I'm fine with merging the document but I'll wait for anyone else to look at it before officially approving.

Thanks for your feedback and the discussion! 

Keeping the ADR minimal but also unambiguous is a challenge. We will develop a better understanding about the "right" scope on the go. 

Currently, I see the ADRs more as a tool for us justify design decisions and less as a holistic documentation. 
While that statement above might sound obvious, I think it is difficult to find the right balance. 

We currently have a `today` repo, and it helps for general understanding to clarify the differences. But as we are currently re-working the repository/package build processes and pipelines, I do not want to include too much legacy definitions (e.g. today) and documentation.

fwilhe (approved):

fwilhe:
@Vincinator Feel free to merge when no one else wants to review

PR #1771, opened 2023-10-12
Install kube-tools via apt repository

damyan:
Install all kubeernetes tools, such as `kubelet`, `kubectl`, `kubeadm`, `crictl` etc. from official repository.

Background: kuebernetes deprecated `kubepkg` ([1], [2]), so systemd units and configuration files can be installed inline only, greatly reducing flexibility. Better: install everything from repository, as recommended ([3], [4])

[1]: https://github.com/kubernetes/release/issues/3265
[2]: https://groups.google.com/g/kubernetes-sig-release/c/jhpoJ1t0fDo/m/bvrbApUODQAJ
[3]: https://kubernetes.io/blog/2023/08/31/legacy-package-repository-deprecation/
[4]: https://kubernetes.io/blog/2023/08/31/legacy-package-repository-deprecation/


damyan:
/cc @gehoern @5kt 
Kindly, review the changes.

features/khost/exec.late
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+K8S_VERSION=${K8S_VERSION:-$(wget --quiet -O- https://storage.googleapis.com/kubernetes-release/release/stable.txt)}
+K8S_VERSION_REPO="${K8S_VERSION%.*}"
+
+wget --quiet -O- "https://pkgs.k8s.io/core:/stable:/${K8S_VERSION_REPO}/deb/Release.key" | gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg
```
Vincinator, line 7: There is no standard for retrieving keys for third party apt repos. 
By downloading and trusting the `Release.key`, we are telling apt to trust everything signed by that key.

If the packages from the apt repo of pkgs.k8s.io are compromised, then we can also not trust the apt key from pkgs.k8s.io. 

I would prefer to have the apt key as a file in a `feature/khost/release.key` file for example.

damyan: @Vincinator Should have been resolved now

damyan:
@Vincinator @5kt I go for a hard-coded kubernetes version, s. https://github.com/gardenlinux/gardenlinux/pull/1771/files#diff-bd72176852b769ba67f234fbc3197001a064b58226d9270a87d07fe943bc3e27R4, but I still suspect patch changes would result in non-reproducible builds.

As an alternative, we could gardenize[sic] all the kube* tools we need. Any opinions on that one?

PR #1795, opened 2023-10-24
fix: move update-bootloaders to _boot and run update-syslinux last

5kt:
_update-syslinux_ should run last when using the _update-bootloaders_
Also, it should be part of the _boot feature.

Fixes #


Vincinator (approved):
It does make sense to me. Looks good.

PR #1796, opened 2023-10-24
Pin debian images for dev env

fwilhe:

/kind technical-debt
/area os
/os garden-linux

Pin the images for dev env to a specific debian testing image. Currently there is a incompatible combination of the kernel and crun in debian testing ([bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1053821)), the selected images are old enough to not have this issue.

It might make sense to always have pinned images here but they would need to be updated from time to time.

Fixes #

**Release note**:

```feature user

```


Vincinator (approved):
workaround looks reasonable to me.

Vincinator:
I think the `build / container image for tests` failed job is unrelated - I just restarted the failed jobs

PR #1809, opened 2023-11-01
Add a script to update the dev env on demand

fwilhe:
This PR addresses an issue with the dev environment.

Previously, the dev environment automatically always used the latest debian testing image available which changes every day. The issue is that debian testing might be broken on some days, so we never know if this will be a working environment. The current example for this issue is [crun and the linux kernel](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1053821) where a new kernel broke crun, and crun was fixed upstream but not in debian testing for a while.

The alternative approach is to use specific debian testing images as introduced in #1796. This has the upside of producing a more reliable environment, but the downside of additional maintenance effort as this image will have to be updated from time to time to avoid it getting too old.

This PR introduces a script to automate the update process, so we can have reliable environments but still update them regularly (probably once or twice a month) without much effort. Running the script will get the most recent debian testing image and rewrite the lima manifest yaml file. Note that testing if the new image works for building Garden Linux is not automated and has to be done before merging changes produced by this script.

### DIFF
--- a/hack/lima-dev-env.yaml
+++ b/hack/lima-dev-env.yaml
@@ -21,12 +21,12 @@ containerd:
   user: false
 
 images:
-  - location: "http://cloud.debian.org/images/cloud/trixie/daily/20231101-1551/debian-13-genericcloud-amd64-daily-20231101-1551.qcow2"
+  - location: "https://cloud.debian.org/images/cloud/trixie/daily/20231010-1529/debian-13-genericcloud-amd64-daily-20231010-1529.qcow2"
     arch: "x86_64"
-    digest: "sha512:b5c4ea6c4554e73b7ab2ed1f85a5e3fc145191169f04a82f03fab284ccdf03b25b271ac8caf1109e6938118991d7c1bdbb02c509718c97f465f1a6701c6bc3cb"
-  - location: "http://cloud.debian.org/images/cloud/trixie/daily/20231101-1551/debian-13-genericcloud-arm64-daily-20231101-1551.qcow2"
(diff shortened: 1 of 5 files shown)


Vincinator:
The update script idea and reasoning sounds good to me. 

Wouldn't it makes sense at this point to outsource the dev environment scripts to a dedicated folder, and add a README.md that describes the setup? 

My Tests for this PR: 
- Tested the `./hack/update-lima-dev-env.py` script within the gl-dev lima VM got an error: 
`OSError: [Errno 30] Read-only file system: 'SHA512SUMS'`
- Tested the `./hack/update-lima-dev-env.py` on my osx, and installed dependencies manually via pip.

fwilhe:
Thanks for the feedback, good suggestions @Vincinator 

Will update

fwilhe:
> Read-only file system: 'SHA512SUMS'

good point, this was caused by a read-only mount of the gl directory. Honestly I've never used that mounted directory and having a local copy of the gl repo inside the vm feels much safer to me. This means that this local checkout will be destroyed when the vm is re-created but as I think pushing code on a regular basis is a good idea this seems like a good idea to me. We could also make it to work in a mounted directory from the host, but I'm not sure if this is safe/fast.


> installed dependencies manually via pip

yeah my idea was that this script is run inside the vm, that's why I've added bs4 as a apt dependency. this will work if you'll do it inside the home directory of the vm user which should now be the default because I removed the unused mount.

Vincinator (approved):
lgtm

PR #1816, opened 2023-11-06
Pin all gh actions 

fwilhe:
For consistency it makes sense to just apply actions pinning everywhere.

This was done using mheap/pin-github-action

Dependabot should be clever enough to update the pinned versions.


Vincinator (approved):

Vincinator:
Makes sense. 

> Dependabot should be clever enough to update the pinned versions.
Yes, let us continue to monitor the behaviour - we will see. Couldn't find anything clear about updating pinned versions in the docs yet

PR #1818, opened 2023-11-09
fix(ci): build artifacts have 8 chars of commitish

Vincinator:
* the release page generation downloads the artifacts from s3, to do that, it needs to construct the name of the artifacts. the new builder uses 8 chars of the commitish.

* Rename the gh action to a fitting name


fwilhe (approved):
LGTM

PR #1822, opened 2023-11-09
fix(release notes): install dependencies for generate release py tool

Vincinator:
Add missing boto3 dependency for the `generate_release_notes.py`


fwilhe (approved):
Consider using apt-get instead of apt as it I think still is not recommended for use in scripts.

PR #1823, opened 2023-11-09
fix: use short commitish also for generation of release note script

Vincinator:
- Use 8 character long commit hash instead of full commit hash

- the `generate_release_notes.py` script requires the commit hash. The commit hash is used to retrieve the artifact. Therefore it must be the same length as in the artifact


MrBatschner (approved):
/lgtm

PR #1831, opened 2023-11-15
fix: check in Pipefile.lock for test container

### LINKED ISSUE #1815: bug(ci): test container can sometimes not be build
**What happened:**
The [build / container image for tests](https://github.com/gardenlinux/gardenlinux/blob/c25f1695147c38b0a9e5bb6f8b26df96714a5df6/.github/workflows/build.yml#L100) fails sometimes, which blocks nightly and dev pipelines of Garden Linux 

**Debug Logs:**
Example of failed Pipeline: https://github.com/gardenlinux/gardenlinux/actions/runs/6768874994/job/18394147551#step:5:21663

```
pipenv.exceptions.ResolutionFailure: ERROR: Getting requirements
to build wheel exited with 1
✘ Locking Failed!

Traceback (most recent call last):
  File "/opt/venv/bin/pipenv", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pipenv/cli/options.py", line 58, in main
    return super().main(*args, **kwargs, windows_expand_args=False)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pipenv/vendor/click/decorators.py", line 92, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pipenv/cli/command.py", line 209, in install
    do_install(
  File "/opt/venv/lib/python3.11/site-packages/pipenv/routines/install.py", line 164, in do_install
    do_init(
  File "/opt/venv/lib/python3.11/site-packages/pipenv/routines/install.py", line 672, in do_init
    do_lock(
  File "/opt/venv/lib/python3.11/site-packages/pipenv/routines/lock.py", line 65, in do_lock
    venv_resolve_deps(
  File "/opt/venv/lib/python3.11/site-packages/pipenv/utils/resolver.py", line 849, in venv_resolve_deps
    c = resolve(cmd, st, project=project)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pipenv/utils/resolver.py", line 718, in resolve
    raise RuntimeError("Failed to lock Pipfile.lock!")
RuntimeError: Failed to lock Pipfile.lock!
Error: error building at STEP "RUN pipenv install --system --dev --verbose": error while running runtime: exit status 1
Error: Process completed with exit code 125.
```

Vincinator:
* Pipfile.lock will be  updated via Dependabot
* Generating Pipfile.lock dynamically in container fails. This commit removes the dynamic Pipfile.lock generation in the container build step, thus removes the error source that the Pipfile.lock generation can fail during the container test image build step
* Fixes nightly and dev builds 

Fixes #1815 

Tested locally with 
```
 podman build --squash --tag test --build-arg base="$(./build --print-container-image)" tests 
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,7 @@ updates:
       - "gardenlinux/garden-linux-maintainers"
 
   - package-ecosystem: "pip"
-    directory: "/"
+    directory: "/tests"
     schedule:
       interval: "weekly"
     reviewers:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,11 @@ updates:
       interval: "weekly"
     reviewers:
       - "gardenlinux/garden-linux-maintainers"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    reviewers:
+      - "gardenlinux/garden-linux-maintainers"
+
--- a/tests/Containerfile
+++ b/tests/Containerfile
@@ -9,6 +9,7 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 RUN pip install pipenv
 COPY Pipfile Pipfile
+COPY Pipfile.lock Pipfile.lock
 RUN pipenv install --system --dev --verbose
 
 COPY init /init
--- a/tests/Pipfile.lock
+++ b/tests/Pipfile.lock
@@ -1,0 +1,2488 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "cf57ff7c68afc5926e42c8d6efa88efb9e7fb6756ccc51cc397e2a50c8e338e7"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.11"
+        },
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.python.org/simple",
+                "verify_ssl": true
+            }
+        ]
+    },
+    "default": {
+        "adal": {
+            "hashes": [
+                "sha256:2a7451ed7441ddbc57703042204a3e30ef747478eea022c70f789fc7f084bc3d",
+                "sha256:d74f45b81317454d96e982fd1c50e6fb5c99ac2223728aea8764433a39f566f1"
+            ],
+            "version": "==1.2.7"
+        },
+        "aliyun-python-sdk-core": {
+            "hashes": [
+                "sha256:c806815a48ffdb894cc5bce15b8259b9a3012cc0cda01be2f3dfbb844f3f4f21"
+            ],
+            "index": "pypi",
+            "version": "==2.14.0"
+        },
+        "aliyun-python-sdk-ecs": {
+            "hashes": [
+                "sha256:7c1af78ff351dbde3617fc74f2a4c709bd3dc0058a0d1d1134f03509565a089d",
+                "sha256:dbe604938ac42b5aa2451378863a821d157d23037c631717d5e6a4284a0448e2"
+            ],
+            "index": "pypi",
+            "version": "==4.24.68"
+        },
+        "aliyun-python-sdk-kms": {
+            "hashes": [
+                "sha256:83166468817a4fbc4c958af43ec22856e1bd80f1363f56acf822206febe6b059",
+                "sha256:f87234a8b64d457ca2338f87650db18a3ce7f7dbc9bfef71efe8f2894aded3d6"
+            ],
+            "version": "==2.16.2"
+        },
+        "antlr4-python3-runtime": {
+            "hashes": [
+                "sha256:3cd282f5ea7cfb841537fe01f143350fdb1c0b1ce7981443a2fa8513fddb6d1a",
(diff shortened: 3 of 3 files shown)


tests/Containerfile
```diff
@@ -8,7 +8,7 @@ RUN python3 -m venv $VIRTUAL_ENV
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 RUN pip install pipenv
-COPY Pipfile Pipfile
```
fwilhe, line 11: Is it correct to remove the pipfile here? Not a python expert, but I'd expect both files are needed.

fwilhe:
I think it makes sense to configure depenabot right away. [Here is an example for a python config](https://github.com/paperless-ngx/paperless-ngx/blob/128594584e00794f7d76cdbb23a5fc0ba101712d/.github/dependabot.yml#L22).

I think we need to adjust the path.

[Dependabot docs for reference](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pip-and-pip-compile)

tests/Containerfile
Vincinator: The Pipefile is used to generate the `Pipfile.lock`. 

The generation of the Pipfile.lock is moved outside of the container, thus `pipenv` must not re-generate a different `Pipfile.lock` during the container build step. 
Therefore I removed the Pipfile, so pipenv does not even try to generate it. 

However, we could add a `--ignore-pipfile` to pipenv.

Vincinator: Hm.. let me check. if we even have to use `--ignore-pipfile`.

Vincinator:
> I think it makes sense to configure depenabot right away. [Here is an example for a python config](https://github.com/paperless-ngx/paperless-ngx/blob/128594584e00794f7d76cdbb23a5fc0ba101712d/.github/dependabot.yml#L22).
> 
> I think we need to adjust the path.
> 
> [Dependabot docs for reference](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pip-and-pip-compile)

Yes, will do

tests/Containerfile
Vincinator: Ok, it's either `--ignore-pipfile` or adding the `Pipfile`to the container image. 

Added Pipfile to the container image now

.github/dependabot.yml
```diff
@@ -8,3 +8,11 @@ updates:
       interval: "weekly"
     reviewers:
       - "gardenlinux/garden-linux-maintainers"
+
+  - package-ecosystem: "pip"
+    directory: "/"
```
fwilhe, line 13:
```suggestion
    directory: "/tests"
```

To my understanding this should point to the directory where our pipfile is located

Vincinator: Do we need to limit dependabot to check only Pipfile in tests? Why?

Vincinator: Checking if dependabot scans recursively or needs explicit paths

Vincinator: Yep - you were right

fwilhe (approved):
lgtm

.github/dependabot.yml
fwilhe: fyi, dependabot docs describe this: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directory

PR #1840, opened 2023-11-22
Rename Bare Container Images Feature

fwilhe:
We chose Bare Container Images as the descriptive term for the feature, this PR attempts to rename accordingly (note that this is not just search/replace as the term is used in different contexts)


Vincinator (approved):
lgtm

PR #1843, opened 2023-11-22
Rename bare container images to bare

Vincinator:
* Use shorter "bare-$config" naming instead of hierarchical names like "bare-container-image/libc"
* Remove legacy `features/_distroless`
  * unbase_oci is used for building the distroless equivalent "bare-$config"
  * remove this feature to reduce complexity and prevent confusion between bare and the legacy feature
* add versioning for bare containers


Vincinator:
Still in draft, testing if the GitHub Actions don't complain

fwilhe (approved):
Only looked at the diff, did not try building the images

LGTM

PR #1852, opened 2023-11-29
github: add labeler action 

Vincinator:

* Add github labeler action
    * must be merged to main before the new labeler action can assign labels to PRs 
    * test repo: https://github.com/Vincinator/test-labeler/pull/1 
* use PR template from gardenlinux/.github org repo

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -32,6 +32,7 @@ features:
 
 tools:
 - 'hack/**'
+- 'tools/**'
 
 logo:
 - 'logo/*'
@@ -48,7 +49,6 @@ checksums:
 
 packages:
 - 'features/**/pkg.*'
-
 
 bare-container:
 - 'bare_flavors/**'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -53,3 +53,34 @@ packages:
 bare-container:
 - 'bare_flavors/**'
 - 'unbase_oci'
+
+gcp-platform:
+- 'features/gcp/**'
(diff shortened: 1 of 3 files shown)


Vincinator:
Requirement: Need label automation

Vincinator:
Requirement: Conventional Commit Checker (other PR?)
Requirement: Static Code Checker (other PR?)

Vincinator:
**Gardener Robot:** you write labels in the PR instead of using GitHub's UI. 
Pro:
- you don't need to use the GitHub UI  
Con:
- you need to know the naming of the Labels

**Labeler GH Action:** Automatically assign labels based on code path
Pro:
- don't need to do anything
Con:
- Things like `breaking-change` must still be assigned manually

fwilhe:
Screenshot 2023-12-04 at 13 13 07

This sounds like it would affect us, but I don't see that config being set.

Vincinator:
Set `dot: true` for labeler workflow

fwilhe:
> Thanks! Added platform labels.
> 
> The `dot: true` setting is a good idea, don't see a reason why not to enable it. You?

No I don't see a reason against, but it does not seem to make a difference for our config as far as I understood now.

Vincinator:
Change PR template here: 
https://github.com/gardenlinux/.github/blob/main/.github/pull_request_template.md

Vincinator:
https://github.blog/changelog/2023-01-10-github-actions-support-for-organization-wide-required-workflows-public-beta/

This should be a required org-wide workflow

PR #1863, opened 2023-12-08
Use guestfish instead of libguestfs-tools in test containers

MrBatschner:

/kind test
/area os
/os garden-linux

The only binary from the guestfs universe that is used in Garden Linux test containers is `virt-copy-in` which is actually part of `guestfish`. Therefore, all the other dependencies (especially Perl) of `libguestfs-tools` can be omitted.

**Release note**:

```feature user
Guestfish is used instead of libguestfs-tools in test containers
```


fwilhe (approved):
Did not try it out, LGTM

PR #1866, opened 2023-12-11
test: Require azure-cli >= 2.44.0 for tests

MrBatschner:

/kind test
/area os
/os garden-linux

Require azure-cli >= 2.44.0 for Azure platform tests. Also make sure that the `Pipfile.lock` is actually used in building the integration test containers so that we have tighter control over which versions of the packages are used.

**Release note**:

```feature user
test: Require azure-cli >= 2.44.0 for tests
```

### DIFF
--- a/container/base-test/Dockerfile
+++ b/container/base-test/Dockerfile
@@ -49,7 +49,7 @@ ENV VIRTUAL_ENV="$VIRTUAL_ENV_PARENT/.venv"
 RUN python3 -m venv $VIRTUAL_ENV
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
-COPY Pipfile* "$VIRTUAL_ENV_PARENT/"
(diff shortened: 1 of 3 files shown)


container/base-test/Dockerfile
```diff
@@ -49,7 +49,7 @@ ENV VIRTUAL_ENV="$VIRTUAL_ENV_PARENT/.venv"
 RUN python3 -m venv $VIRTUAL_ENV
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
-COPY Pipfile "$VIRTUAL_ENV_PARENT"
+COPY Pipfile* "$VIRTUAL_ENV_PARENT/"
```
Vincinator, line 52:
```suggestion
COPY Pipfile Pipfile.lock "$VIRTUAL_ENV_PARENT/"
```
nitpick/style: explicitly control what is included. 
wildcard is also fine for me.

Vincinator (approved):

PR #1867, opened 2023-12-11
remove vim-tiny from server feature

Vincinator:

* less potential security vulnerabilities

* as an alternative; vim can open files via scp

**Release note**:

```feature user
Remove vim-tiny from Garden Linux. 
* less potential security vulnerabilities
* as an alternative; vim can open files remotely via scp
```

### DIFF
--- a/features/server/pkg.include
+++ b/features/server/pkg.include
@@ -28,5 +28,4 @@ systemd-cron
 systemd-timesyncd
 tcpdump
 traceroute
-vim-tiny
 wget


MrBatschner:
If that means no vim at all, then debugging issues in a Gardener environment will be very difficult or even impossible (the scp approach is likely not working that easily there). How about moving vim-tiny to the gardener feature (even though that means no less compliance overhead)?

PR #1874, opened 2023-12-12
Define builder version in dockerfile

fwilhe:
This change is needed to allow keeping the builder updated via dependabot.

### DIFF
--- a/builder.dockerfile
+++ b/builder.dockerfile
@@ -1,4 +1,3 @@
-# This file intentionally only has the 'FROM' line.
-# Defining the builder version this way allows us to update it via dependabot which was not possible any other way.
+# Dependency management via Dependabot
 
 FROM ghcr.io/gardenlinux/builder:002a2ff4d1dc6a39d04543b6a6e92a1465d9e226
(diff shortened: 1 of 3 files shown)


builder.dockerfile
```diff
@@ -0,0 +1,4 @@
+# This file intentionally only has the 'FROM' line.
+# Defining the builder version this way allows us to update it via dependabot which was not possible any other way.
```
Vincinator, line 2:
```suggestion
# Dependency management via Dependabot
```

Would keep the explanation simple here

Vincinator (commented):
We had a follow up meeting days ago. Not sure if this PR is still relevant? 

I guess yes, and LGTM.

fwilhe:
> We had a follow up meeting days ago. Not sure if this PR is still relevant?
> 
> I guess yes, and LGTM.

yes, still relevant imo. Updated to include it in the dependabot updates. We should then also switch to builder releases, but I'd do this in a separate PR.

Vincinator (approved):

PR #1877, opened 2023-12-18
Add isc dhcp client to kvm and metal

Vincinator:

Fixes build of Garden Linux `kvm` and `metal` based flavours
- only effects flavours using `_boot`, flavours using `_secureboot` were not affected 
- `dracut[I]: Module 'network-legacy' will not be installed, because command 'dhclient' could not be found!`
  
- might be related https://github.com/dracutdevs/dracut/commit/e863807

### DIFF
--- a/features/kvm/pkg.include
+++ b/features/kvm/pkg.include
@@ -1,2 +1,3 @@
 dosfstools
+isc-dhcp-client
 qemu-guest-agent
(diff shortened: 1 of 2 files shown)


Vincinator:
network-legacy is just a fallback option. We want to use systemd-networkd instead. See #1879

PR #1879, opened 2023-12-18
Dracutnetworkfix

gehoern:
there was a build error that url-lib could not be installed in dracut because of missing dhcpsupport.

Preferring systemd-networkd fixed the problem.


5kt:
IMHO this should be fixed a bit differently, please look at #1880

PR #1911, opened 2024-01-12
update sap machine in bare flavour + checksum updates

Vincinator:
updates sap machine to LTS 🆙 
fixes nightly 🌙 
motivated by 🐰


fwilhe (approved):
looks good. I've built it locally for aarch64, but not for x64

PR #1922, opened 2024-01-17
Fix build after breaking it in #1874

fwilhe:
The change in #1874 had unintended side-effects. This PR attempts to mitigate those issues, but might not be the most elegant solution for that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -115,7 +115,9 @@ jobs:
       - name: build test container
         run: |
           OCI_IMAGE_TAG="$(./build --print-container-image)"
-          podman build -t "$OCI_IMAGE_TAG" -f builder.dockerfile .
+          if [[ -f builder.dockerfile ]]; then
+            podman build -t "$OCI_IMAGE_TAG" -f builder.dockerfile .
+          fi
           podman build --squash --tag test --build-arg base="$OCI_IMAGE_TAG" tests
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -114,7 +114,9 @@ jobs:
       - uses: ./.github/actions/setup
       - name: build test container
         run: |
-          podman build --squash --tag test --build-arg base="$(./build --print-container-image)" tests
+          OCI_IMAGE_TAG="$(./build --print-container-image)"
+          podman build -t "$OCI_IMAGE_TAG" -f builder.dockerfile .
```
fwilhe, line 118: this now assumes the existence of `builder.dockerfile`, which I think is okay in this case. It would be possible to make it more generic, but since this is not re-usable library code here, I think that's okay. Simplicity over being generic.

.github/workflows/build.yml
```diff
@@ -114,7 +114,9 @@ jobs:
       - uses: ./.github/actions/setup
       - name: build test container
         run: |
-          podman build --squash --tag test --build-arg base="$(./build --print-container-image)" tests
+          OCI_IMAGE_TAG="$(./build --print-container-image)"
```
fwilhe, line 117: This will always print `localhost/builder`, but I think it's better not to have that string hard coded here, so it will also work in other cases.

PR #1930, opened 2024-01-23
Update docs for new builder

### LINKED ISSUE #1929: Update outdated references to `make $TARGET` in docs
Just noticed that for example [here](https://github.com/gardenlinux/gardenlinux/blob/a2644846d713501989c6293ae9f0e78edbeefbdb/docs/02_operators/deployment/install-non-default.md?plain=1#L36) there seem to be outdated references. Should be updated to the newer build system using the `./build` shell script.

fwilhe:
Fixes #1929


Vincinator (approved):
LGTM

PR #1932, opened 2024-01-29
Fix build after breaking it in #1874

fwilhe:
Additional output created by building the image breaks assumptions about what ./build --resolve-cname does. Fix by grepping for the cname.


fwilhe:
closed in favour of solution by @nkraetzschmar

PR #1935, opened 2024-01-31
Update builder to include gardenlinux/builder#57

fwilhe:
This should fix the failed nightly builds because they make assumptions about what is written to stdout.

### DIFF
--- a/build
+++ b/build
@@ -3,6 +3,9 @@
 set -euo pipefail
 shopt -s nullglob
 
+exec 3>&1
+exec 1>&2
+
 container_image=localhost/builder
 container_engine=podman
 target_dir=.build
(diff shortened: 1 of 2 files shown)


fwilhe:
CI is red now because the builder image was not yet pushed. WIP here https://github.com/gardenlinux/builder/actions/runs/7724005235/job/21055243308, should work once that succeeded.

nkraetzschmar (approved):
lgtm

gehoern (approved):

fwilhe:
Sadly this currently does not work because the builder is (currently) based on debian stable which does not contain [systemd-ukify](https://packages.debian.org/trixie/systemd-ukify)

brdanin:
This will soon change, because the TPM2 feature is using debian testing for the builder and includes `systemd-ukify`

PR #1938, opened 2024-02-02
Add TPM2 emulation support to start-vm using `swtpm`

brdanin:
This PR adds the optional flag `--tpm2` to the `start-vm` shell script. If enabled, an emulated TPM chip is launched by a process in the background and injected into the VM as a new device.

Very useful for debugging and developing TPM related features, `swtpm` needs to be installed on the hosts OS though.


fwilhe (approved):
Didn't test, but looks sensible to me. Feel free to merge as you see fit.

PR #1948, opened 2024-02-10
fix(startvm): Remove watchdog option due to incompatibility in newer versions

### LINKED ISSUE #1854: `Bug`: Unavailable option `-watchdog` in `startvm`
**What happened:**
Using QEMU on newer systems (even Debian Bookworm) will report, that the watchdog option is not more present. Therefore, `startvm` will fail:

```
Press any key to continue...
qemu-system-x86_64: -watchdog: invalid option
```

gyptazy:
fix(startvm): Remove watchdog option due to incompatibility in newer versions

Remove watchdog option due to incompatibility in newer versions of `qemu-system-x86_64`. However, this also only worked with Debian version and on x86.

Fixes #1854

**Release note**:

```feature user
- category: bugfix
- target_group: user
```

### DIFF
--- a/bin/start-vm
+++ b/bin/start-vm
@@ -103,7 +103,6 @@ set_defaults () {
     monitor=1
     bridge=
     daemonize=
-    no_watchdog=0
     pxe=
     pxe_binary=
     pxefile=${CURR_DIR}/../examples/ipxe/start-vm.ipxe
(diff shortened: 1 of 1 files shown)


Vincinator (approved):
LGTM - thanks

PR #1957, opened 2024-02-13
ci: dependabot strategy to increase-if-necessary

Vincinator:
* we do not need to upgrade unnecessary pip or gh actions

* feedback on how to improve this configuration is very welcome

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,6 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-    versioning-strategy: "increase-if-necessary"
     reviewers:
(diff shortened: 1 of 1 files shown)


Vincinator:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy 

according to documentation "increase-if-necessary" is N/A for GitHub Action ecosystem

fwilhe:
Thanks, I was not aware of this option.
For reference: https://github.blog/changelog/2022-10-24-reduce-dependabot-version-updates-in-your-python-projects-with-the-increase-if-necessary-strategy/

fwilhe (approved):

Issue #1842, opened 2023-11-22
Platform test on Azure fails due to a timeout in creating a ssh connection to the VM under test in azure


Vincinator:
**What happened:**

The azure platform tests run by the nightly github action failed.
The test-VM in azure is not booting fast enough for our timeout.

```
paramiko.transport:transport.py:1893 Auth banner: b'"System is booting up. Unprivileged users are not permitted to log in yet. Please come back later. For technical details, see pam_nologin(8)."\n'
```

-  [Call to wait_systemd_boot](https://github.com/gardenlinux/gardenlinux/blob/90300db8ec95a0f2b7857b0821fd793a16d9532e/tests/integration/azure.py#L451)
- [wait_systemd_boot(client) definition](https://github.com/gardenlinux/gardenlinux/blob/90300db8ec95a0f2b7857b0821fd793a16d9532e/tests/helper/utils.py#L119)

Ultimately, the tests do not start until this synchronously blocking command is done. 
```
systemctl is-system-running --wait
```
see systemctl [docs](https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#--wait)

**Debug Logs:**
- 1312.1 build pipeline: https://github.com/gardenlinux/gardenlinux/actions/runs/6955241114/job/18925068987#step:13:99

**Environment:**
- GL 1312.1 in rel-1312 branch

Vincinator:
Azure Issue, won't fix in garden linux pipeline - currently no issues. We could implement further workarounds (retry or increase wait).. remove this issue

PR #1971, opened 2024-02-20
test: handle pam_nologin during connection attempt

MrBatschner:

In some rare cases, the Garden Linux platform test framework attempts to connect to a machine during its boot-up in that small time frame when SSH already accepts connections but PAM refuses them with the "System is booting up" message. In this case, the tests should not be aborted with an error but the connection should be retried.

Along the lines, a little code cleanup.

**Release note**:

```feature user
Garden Linux platform tests no longer abort on pam_nologin but attempt to retry.
```

### DIFF
--- a/tests/helper/sshclient.py
+++ b/tests/helper/sshclient.py
@@ -7,8 +7,11 @@ from os import path
 from binascii import hexlify
 
 from paramiko import SSHClient, AutoAddPolicy, RSAKey
-from paramiko.auth_handler import AuthenticationException, SSHException
-from paramiko.ssh_exception import NoValidConnectionsError
+from paramiko.ssh_exception import (
+    NoValidConnectionsError,
+    AuthenticationException,
+    SSHException
+)
 from scp import SCPClient, SCPException
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


tests/helper/sshclient.py
```diff
@@ -89,6 +93,17 @@ def __init__(
         self.ssh_key_filepath = path.expanduser(sshconfig['ssh_key_filepath'])
 
 
+    def _increase_retry_count(self, wait: bool = True):
+        if self.retry_count >= self.ssh_max_retries:
+            max_timeout = self.retry_count * self.ssh_retry_timeout_seconds
+            # FIXME: this should not be pytest.exit() as it will exit immediately without cleanup
+            pytest.exit(f"Unable to establish an SSH connection after {max_timeout=} seconds. Aborting all tests.", returncode=5)
```
Vincinator, line 100: Nit pick: do we need to support the case where `wait` param is set to false? 

In the wait=False case we would not do a `time.sleep(self.ssh_retry_timeout_seconds)`, but after `self.retry_count >= self.ssh_max_retries` retries, we exit with the message that we have waited for `max_timeout`

Vincinator: If we remove the wait param, we should reconsider the name. 

before it was clear that this function has the side effect of a sleep. But without a param I would prefer to have it in the function name e.g.`_increase_retry_with_wait`

Vincinator:
Fixes #1842

MrBatschner:
Failing CI is unrelated to this change, merging.

PR #1974, opened 2024-02-26
test: generalize check for installed kernel modules

Vincinator:
This commit introduces the check_kernel_module_exists test helper function.

* considers compression 
    * adding new compression types by extending the list in check_kernel_module_exists function
* added support to run kernel module availability tests also on chroot targets 
    * previously not possible due to kernel version detection via uname
    * now we can run `./test` locally and the kernel module checks are also executed 
* refactored and generalized checks to not only check for wireguard

Fixes nightly platform tests, because wireguard is now compressed in new 6.6 kernel. It is `wireguard.ko.xz` instead of `wireguard.ko`.

* Added a note in the `tests/README.md` that the chroot environment spawns an sshd inside the chroot in a container, and the tests communicate over ssh with the chroot environment.

* We should add more kernel module availability check tests in follow up PRs.
food for thought:
    - check gvnic driver in `features/gcp/tests/test_kernel_modules.py`
    - check nvme driver


tests/helper/utils.py
```diff
@@ -72,6 +72,17 @@ def get_file_perm(client, fname):
     if not "cannot statx" in error:
         return int(output)
 
+def check_kernel_module_exists(client, module_name):
+    """Check if a kernel module exists in any subdirectory of /lib/modules, considering compression."""
+
+    for ext in ["ko", "ko.gz", "ko.xz", "ko.bz2", "ko.zst"]:
+        filename = f"{module_name}.{ext}"
+        command = f"find /lib/modules/ -type f -name {filename}"
```
MrBatschner, line 80: Before, it was explicitly looked into `/lib/modules/$(uname -r)`. Is it safe to become more generic here?
Since this runs in a running system, would 
```suggestion
        command = f"find /lib/modules/$(uname -r) -type f -name {filename}"
```
be an option?

MrBatschner (changes requested):

tests/helper/utils.py
MrBatschner: Ok, sometimes, reading the PR description is helpful. Still, what would be if we have multiple kernel packages in the system, one comes with the module, one does not?

Vincinator: > Is it safe to become more generic here?
I asked myself the same question. Given our Gardener OS build targets, it is safe to assume that only install one kernel package is installed per image.

However, this might change for some reason in Gardener OS or a spin-off project in the future. 

I really want to have these checks available in the default "chroot" test environment. I will try to find a better solution that is future-proof.

features/cloud/test/test_only_one_kernel.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+import helper.utils as utils
+
+def test_number_installed_kernels(client):
+    """ Only one kernel should be installed """
+    installed_kernels = utils.get_installed_kernel_versions(client)
+    nr_kernels = len(installed_kernels)
+    assert nr_kernels == 1, f"Number of installed kernels {nr_kernels} != 1"
```
fwilhe, line 8: super small nitpick, but 'nr' does not seem to be a common abbreviation for 'number' in English.. I'd prefer the longer number of instead.

```suggestion
    number_of_kernels = len(installed_kernels)
    assert number_of_kernels == 1, f"Number of installed kernels {nr_kernels} != 1"
```

tests/helper/utils.py
```diff
@@ -72,6 +72,42 @@ def get_file_perm(client, fname):
...
+    """ Checks if a kernel module exists for a given kernel version """
+    
+    for ext in ["ko", "ko.gz", "ko.xz", "ko.bz2", "ko.zst"]:
+        filename = f"{module_name}.{ext}"
+        command = f"find /lib/modules/{kernel_version} -type f -name {filename}"
+        (exit_code, output, error) = client.execute_command(command, quiet=True)
+        if exit_code == 0 and output:
+            return True
+
+    return False
```
fwilhe, line 95: I'm missing experience with pytest, so for me it's not clear here if this fails the test or not. Should it fail the test? If so is there a more self-describing way like raising a new exception?

fwilhe:
Two very minor comments/questions. Looks good otherwise. Feel free to merge when @MrBatschner sees his comments as addressed.

tests/helper/utils.py
Vincinator: This is the helper function for the actual pytest (with a proper assertion)

features/cloud/test/test_kernel_modules.py:
```
def test_kernel_modules(client, module_name):
     """Check if the module_name kernel module is available as loadable module """
     # NOTE: check all kernel versions, in case multiple kernels are installed
     assert utils.check_module_exists_in_all_kernels(client, module_name), f"{module_name} does not exist in image"
```

Vincinator: We could also raise an exception in that helper function true.

 I suggest to keep the style coherent with the other helper functions for now. Not sure if a proper exception handling is beneficial or more elegant.

fwilhe: thanks, makes sense

fwilhe (approved):

MrBatschner (approved):

PR #1982, opened 2024-02-27
Patch crun

fwilhe:
Workaround as suggested in https://github.com/actions/runner-images/issues/9425

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -3,11 +3,13 @@ runs:
   steps:
 
     # Workaround for https://github.com/actions/runner-images/issues/9425, to be removed once the issue is resolved
-    - run: crun --version
+    - name: Show crun version before the patch
+      shell: bash
+      run: |
+        crun --version
     - name: patch crun (Workaround for https://github.com/actions/runner-images/issues/9425)
       shell: bash
       run: |
-        set -x
         curl -Lo ./crun https://github.com/containers/crun/releases/download/1.14.3/crun-1.14.3-linux-amd64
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yml
```diff
@@ -1,6 +1,14 @@
 runs:
   using: composite
   steps:
+
+    # Workaround for https://github.com/actions/runner-images/issues/9425, to be removed once the issue is resolved
+    - name: patch crun (Workaround for https://github.com/actions/runner-images/issues/9425)
+      shell: bash
+      run: |
+        curl -Lo ./crun https://github.com/containers/crun/releases/download/1.14.3/crun-1.14.3-linux-amd64
+        sudo install crun /usr/bin/crun
```
Vincinator, line 10: I suggest to verify signature of the downloaded crun binary
https://github.com/containers/crun/releases/download/1.14.3/crun-1.14.3-linux-amd64.asc

fwilhe: Good suggestion, thanks. I had issues with doing it with gpg for reasons, probably I don't understand the gpg cli good enough. Resorted to using sha sum for now, sadly they don't seem to publish it for releases. I think this should be safe enough for our purposes now, but I'm also open to other suggestions.

fwilhe: So the 'good' sha sum in here was created by me because sadly the github release does not have sha sums

Vincinator: Yepp, that works for me

Vincinator (approved):

PR #1983, opened 2024-02-27
tests: add kernel config tests

### LINKED ISSUE #1981: Take ownership of kernel configuration - remove implicit dependencies to debian across kernel versions
**Context**
The current kernel build for Garden Linux uses debian kernel config as starting point, and adapts this with own Garden Linux specific configurations (using the debian kernel config tooling).

- [ ] Linux package owns kernel configuration, no implicit dependency from a debian git repo
    * kernel config will be maintained at `gardenlinux/package-linux` repo(s)
- [ ] Kernel config Tests in gardenlinux/gardenlinux
    * GL features that install a kernel, can add tests to check for important kernel configuration (firecracker, metal, cloud) 
    * These tests should be moved to this repository, testing the actual kernel installed in the Garden Linux image - removing potential de-sync error sources 
 - [ ] Add automation to detect kernel config changes in debian 
     * our kernel config is based on the debian kernel config.
     * we need to be notified about changed configs 

**Why is this needed**:
To simplify and to eliminate potential error sources, we want to own and maintain all parts of the kernel config, so that if debian changes a kernel config, we would need to actively pull those changes.

Vincinator:
Refactors the kernel config tests
* Checks all configurations of all installed kernels
    * in theory, only one kernel is installed during build time. But if we include two different kernels in the image (intentionally or unintentionally), then we will have multiple /boot/config-* files, and the result of the kernel config test is useless- can not be trusted. Therefore this test does not assume anything about the number of installed kernels, and just verifies that all included configs are correct.  
* Check each config individually to identify wrong config directly
* Three test types:
    * is enabled: either `y` or `m`
    * is disabled: either not set or `n`
    * is exact: check if exactly this line is contained in the config

Add another layer of verification that our kernel configuration is correct.

Fixes #1981

### DIFF
--- a/features/cloud/test/test_kernel_config.py
+++ b/features/cloud/test/test_kernel_config.py
@@ -12,7 +12,7 @@ import helper.utils as utils
 def test_kernel_configs_enabled(client, kernel_config_item):
     kernel_config_paths = utils.get_kernel_config_paths(client)
     for kconf in kernel_config_paths:
-        assert utils.check_kernel_config_enabled(client, kconf, kernel_config_item), f"Kernel Config: {kernel_config_item} is not set, but expected it to be set to y or m"
+        assert utils.check_kernel_config_exact(client, kconf, f"{kernel_config_item}=(y|m)"), f"Kernel Config: {kernel_config_item} is not set, but expected it to be set"
 
 @pytest.mark.parametrize(
     "kernel_config_item",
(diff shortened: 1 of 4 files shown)


features/metal/test/test_kernel_config.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+import helper.utils as utils
+
+
+@pytest.mark.parametrize(
+    "kernel_config_item",
+    [
+        "CONFIG_X86_SGX",
```
fwilhe, line 8: maybe it would be good to add a string field here to provide a reasoning for why this config is desired (or not desired)

string field would be nicer than a comment imo, can be part of the assertion error message

Vincinator: I like the idea. This must be extended to contain the following information 
(config name, reasoning string, architecture)

Vincinator: further simplified to 
(key, value, rational, arch) 

value should not be implicitly encoded in function calls imo

fwilhe (approved):
Looks good to me

features/cloud/test/test_kernel_config.py
```diff
@@ -0,0 +1,32 @@
...
+        if arch in kconf or arch == "all": 
+            if kernel_conf_value == "n":
+                print(f"Check if {kernel_conf_key} is disabled for {arch}")
+                assert not utils.check_kernel_config_exact(client, kconf, f"{kernel_conf_key}=(y|m)"), f"Kernel Config: {kernel_conf_key} is set, but expected it to be disabled or not set"
+            else:
+                print(f"Check if {kernel_conf_key} is enabled for {arch}")
+                assert utils.check_kernel_config_exact(client, kconf, f"{kernel_conf_key}={kernel_conf_value}"), f"Kernel Config: {kernel_config_item} is not set to {kernel_conf_value}"
+        else:
+            print(f"Skipping config test of {kernel_conf_key} for architecture {arch}")
+
```
MrBatschner, line 32: Would it make sense to you to externalize this function to helpers so that it exists only once?

PR #1985, opened 2024-02-28
disabling ssh now by default

gehoern:
since the Gardener adaptation of a new security requirement: ssh is now off by default.

Gardener can enable this feature over its cloud provider tooling.

Just for documentation:
- the other nodes are not more unsecure because of SSH since
- there is no public key / passwort installed and needs cloud provider tooling to upload (google tools / cloudinit etc)
- every gardener node is in a private network (not reachable from outside anyhow)
- can only reached via bastion host functionality


MrBatschner (changes requested):
/hold this PR until a discussion with Gardener DoDs came to a result

fwilhe (approved):

dguendisch:
I don't see how we would be able to inspect/debug failures like nodes not being able to join the cluster (if the node cannot join, it won't be able to retrieve cluster configuration like ssh enablement, so this node would never be able to enable ssh).
Hence I think this change would greatly decrease our possibility to debug such issues and therefore I would not want to run such an image in our landscapes 😉.
(maybe it would be an option to produce different images with different defaults should there be GL consumers that want to have this disabled by default, but for our landscapes I'd strongly prefer to keep it enabled).

dimityrmirchev:
Unless I am missing something it should not matter much in the Gardener context because on/off is controlled by a shoot setting (as of now it defaults to `true`) https://github.com/gardener/gardener/blob/master/docs/usage/shoot_workers_settings.md

MrBatschner:
Was talking about that with @scheererj: if Gardener were to deploy the setting of `sshAccess: {enabled: true}` through cloud-init (i.e. osc-provision), then we could disable SSH by default in the Garden Linux image. If a user wished to have SSH enabled for a Shoot, it would get enabled early on by cloud-init and subsequently reconciled by gna. If a node is so broken that even cloud-init would not run, then having SSH would not help in any way because there would be no login user.

MrBatschner (approved):
Said logic is already present in Gardener ([ref](https://github.com/gardener/gardener/blob/20c560a3756473fa691f975eea056a30f6b3b456/pkg/component/extensions/operatingsystemconfig/operatingsystemconfig.go#L698-L708)), so this PR would be ok from Gardener POV.

Therefore, LGTM.

PR #1988, opened 2024-02-28
Support generation of 2048 bit RSA keys for TPM PCR value signing

brdanin:
As it often cannot be guaranteed that all hardware TPM chips support the standard key size of 4096 bit, this PR allows the generation of specific RSA keypairs for signing TPM PCR values, which only have a size of 2048 bit.

image


nkraetzschmar (approved):
lgtm

PR #2011, opened 2024-03-05
Fix platform tests apply crun workaround

Vincinator:
Platform tests fail because of empty cname, and download-artifacts step tries to download all because no input was provided due to empty cname

```
Error: OCI runtime error: chmod `run/shm`: Operation not supported
cname=
```

Therefore this PR does two things:
1. Include fix, so that cname can be resolved with `./build --resolve-cname` in pipeline (crun workaround)
2. Make `get cname` steps fail in github actions bash environment that has **not** set `pipefail`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -219,7 +219,9 @@ jobs:
         with:
           path: tests/test.xml
       - name: get cname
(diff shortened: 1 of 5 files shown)


nkraetzschmar (approved):
lgtm

PR #2041, opened 2024-03-13
use packages.gardenlinux.io for release page generation

Vincinator:
Note: release page action does not need to be run again for 1443.0, I triggered it from the  fix-generate-release-page-for-new-package-pipeline branch


fwilhe (approved):
thanks

PR #2043, opened 2024-03-15
Add script to automatically bump SapMachine version

fwilhe:
So far the SapMachine version we include needs to be updated manually.

This PR adds a script to help with that, but the update is still a manual process.

The script finds the latest minor/patch version of a given major version along with the checksums of the tarballs.

Running `update-sapmachine.py` will print out the variables to be used in `exec.config`.
A potential improvement is to edit the `exec.config` script automatically via `update-sapmachine.py`.

We still want the checksum and the exact SapMachine version to be stored in the git repo, so we don't want to hide this by automatically running `update-sapmachine.py` during the build for reproducibility reasons.

### DIFF
--- a/features/sapmachine/update-sapmachine.py
+++ b/features/sapmachine/update-sapmachine.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+import json
+import argparse
+
+from urllib.request import urlopen
+
+sapmachine_releases_json_url = 'https://sap.github.io/SapMachine/assets/data/sapmachine-releases-website.json'
+
+
+def loadReleases():
+    with urlopen(sapmachine_releases_json_url) as response:
+        releases = json.loads(response.read().decode())
+        return releases
+
+
+def loadChecksum(url):
+    with urlopen(url) as checksum:
+        checksum_file_content: str = checksum.read().decode()
+        # Need to split because the file contains the checksum and the file name
+        # We only want the checksum which is the first element in the line
+        checksum = checksum_file_content.split(' ')[0]
+        return checksum
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--majorJreVersion', type=str)
+    args = parser.parse_args()
+    major_jre_version = args.majorJreVersion
+    releases = loadReleases()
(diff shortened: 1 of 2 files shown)


features/sapmachine/update-sapmachine.py
```diff
@@ -0,0 +1,45 @@
...
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--majorJreVersion', type=str)
+    args = parser.parse_args()
+    major_jre_version = args.majorJreVersion
+    releases = loadReleases()
+    tag: str = releases['assets'][major_jre_version]['releases'][0]['tag']
+    version = tag.split('-')[1]
+    downloadUrlSapMachineBinary_amd64: str = releases['assets'][major_jre_version]['releases'][0]['jre']['linux-x64']
+    downloadUrlSapMachineBinary_aarch64: str = releases['assets'][major_jre_version]['releases'][0]['jre']['linux-aarch64']
```
fwilhe, line 34: I'm not sure about the 'releases' array here.. It seems to always only have a single entry which is the lastest minor-patch version, but if this assumption is false this might not give us the latest patch release here.. need to clarify if this is okay.

Vincinator:
Thanks, this update helper is an improvement. As a next step, let us align if we can do package version management via package repository. Then we would have automated this manuell process. We would still need to decide on what java LTS we want to include in the nightly, but this would be by design imo.

features/sapmachine/update-sapmachine.py
```diff
@@ -0,0 +1,45 @@
...
+        checksum_file_content: str = checksum.read().decode()
+        # Need to split because the file contains the checksum and the file name
+        # We only want the checksum which is the first element in the line
+        checksum = checksum_file_content.split(' ')[0]
+        return checksum
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--majorJreVersion', type=str)
```
Vincinator, line 27:
```suggestion
    parser.add_argument('--majorJreVersion', required=True, type=str)
```

Vincinator: just for convenience

Vincinator (approved):

PR #2060, opened 2024-03-25
fix: release page automation - rm package updates section

Vincinator:
* security update annotations are not tracked in package repository, but in security tracker and other internal tooling.
* new package pipeline does not contain annotations for CVEs, thus we need to remove the packages updates section in the release page automation => need to include it manually


fwilhe (approved):
CI fails because of

```
 E: Unable to locate package fail2ban
```

That's unrelated to this PR, cf https://tracker.debian.org/news/1514146/fail2ban-removed-from-testing/

PR #2081, opened 2024-04-17
Rootless podman feature

fwilhe:
Feature for setting up podman to be used in rootless mode.

This is intended to be used in combination with the lima feature, but it might also be used standalone.


fwilhe:
Would be nice to have a few tests for the feature, but I'm not sure how to approach that. As this feature is not part of the core offering, I think it's okay to relax the rest requirements a bit.

Vincinator (approved):
looks ready for the lima images - LGTM 👍🏻

PR #2082, opened 2024-04-17
fix: start-vm: correctly get homebrew_prefix

nkraetzschmar:


fwilhe (approved):
tried it out locally, fixed the issue for me. thanks @nkraetzschmar

PR #2083, opened 2024-04-17
Build with SELinux enabled

fwilhe:
Possible thanks to https://github.com/gardenlinux/builder/pull/73

Tested with podman machine on macos with a new vm that has SELinux enabled


Vincinator (dismissed):
Nice!

EDIT: Sorry, made the connection to our chat a bit too late -> removed approval

PR #2100, opened 2024-05-03
docs: use new builder in oras and oci feature README

### LINKED ISSUE #1689: Outdated docs about build.sh
There's still lots of outdated docs, which reference the `build.sh`. We should remove those docs/replace them with info reflecting the builder build-process.
We should not describe details of how the builder works in the gardenlinux repo - this should be done in the builder repo. In the gardenlinux repo we should describe details about gardenlinux-specific implementation.

Vincinator:
Updates remaining references to old build.sh.  fixes #1689


fwilhe (approved):
Thanks

PR #2101, opened 2024-05-03
Fix 1996 remove garden feat tests

### LINKED ISSUE #1996: Delete 'garden-feat' tests if they are obsolete
If my understanding is correct, the 'garden-feat' tests can't work anymore as the script was removed [here](https://github.com/gardenlinux/gardenlinux/commit/ef0e0e6c0383494f9d72670a1e8585e709c808e2). Not sure what qualities the tests ensured but I don't think they work with the current code base. I think the old 'garden-feat' is now (partially) the parse features script in the builder, maybe that should have those tests?

https://github.com/gardenlinux/gardenlinux/blob/34bb8c96fcda3c53a5573e320ab5534141915c6e/tests/local/test_garden_feat.py#L1

Vincinator:
garden-feat does not exist anymore in favour of the [parse_features](https://github.com/gardenlinux/builder/blob/main/builder/parse_features) from the new builder.

Thus, this PR removes a test of that non existing script. 

fixes #1996


fwilhe (approved):
Thanks for taking care 💯

PR #2113, opened 2024-05-15
Umask test use umask cmd

### LINKED ISSUE #2112: fix GCP umask platform test
- [test_umask](https://github.com/gardenlinux/gardenlinux/blob/main/features/base/test/test_umask.py) test  fails in nightly platform tests
- Running this test in chroot environment works
- Manually creating a `gcp-gardener_prod` image in GCP and testing on a VM in GCP also works fine 

While investigating this issue, I implemented a new test that additionally also checks value of umask command instead of only reading /etc/login.defs UMASK value. Why? Because umask value can be set in root/.profile of root/.bashrc or somewhere in the bash environment => also test it with umask command.  We still need the /etc/login.defs check, because pam_umask for example reads from there.

Vincinator:
**What this PR does**
- repair `non_gcp` flag for python platform tests
- add a second umask test `test_umask_cmd` (check via `umask`) 
- disable `test_umask` for platform gcp tests (check by reading UMASK value from /etc/login.defs) 

**Why we need it?**
- pam_umask reads the umask value from /etc/login.defs: we need to test /etc/login.defs
- `/root/.bashrc`, `/root/.profile`, ... can overwrite default value from /etc/login.defs
    - ➡️  we need to check via umask cmd in root bash environment

Conclusion: we need both tests. Disabling `/etc/login.defs` for gcp only is only acceptable, because we additionally have chroot tests that still check the file content.

|               | platform test (gcp) | chroot tests (gcp) |
|---------------|---------------|--------------|
| login.defs    |        ❌       |         ✅       |
| umask cmd     |         ✅        |        ✅        |

Fixes #2112 

We still need to find out why test_umask only fails on gcp platform tests (could not find /etc/login.defs).  

Tested gcp-gardener_prod image:
- ✅  Tested locally (all green)
- ✅  Tested manually in gcp by creating a custom image and spawning a vm (all green)
- ❌  Tested automatically via nightly platform tests (test_umask fails)

**Release note**:

```feature user

```

### DIFF
--- a/features/base/test/test_umask.py
+++ b/features/base/test/test_umask.py
@@ -1,4 +1,5 @@
 import pytest
+from helper.tests.file_content import file_content
 
 # Parametrize the test unit with further
 # options.
--- a/features/base/test/test_umask.py
+++ b/features/base/test/test_umask.py
@@ -5,4 +5,4 @@ def test_umask(client):
     (exit_code, output, error) = client.execute_command(
         cmd, quiet=True)
     assert exit_code == 0, f"Could not execute umask cmd: {error}"
-    assert output == "0027", "umask is not set to 0027 "
+    assert output == "0027\n", "umask is not set to 0027 "
--- a/features/base/test/test_umask.py
+++ b/features/base/test/test_umask.py
@@ -1,8 +1,29 @@
 import pytest
 
-def test_umask(client):
+# Parametrize the test unit with further
+# options.
+# First: File to process
+# Second: Key to search
+# Third: Value of Key
+@pytest.mark.parametrize(
+    "file,dict",
+    [
+        ("/etc/login.defs", {"UMASK": "027"})
+    ]
+)
+
+
+# Run the test unit to perform the
+# final tests by the given artifact.
+def test_umask(client, file, dict):
+    # Check /etc/login.defs
+    file_content(client, file, dict)
+
+def test_umask_cmd(client, non_container):
+    # Additionally check umask via cmd
     cmd = f"sudo su root -c umask"
     (exit_code, output, error) = client.execute_command(
         cmd, quiet=True)
+
     assert exit_code == 0, f"Could not execute umask cmd: {error}"
     assert output == "0027\n", "umask is not set to 0027 "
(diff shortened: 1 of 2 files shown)


features/base/test/test_umask.py
```diff
@@ -16,5 +16,15 @@
 
 # Run the test unit to perform the
 # final tests by the given artifact.
-def test_umask(client, file, dict):
```
fwilhe, line 19: Very nice pr description, but for visibility reasons I'd suggest to add a comment here to explain why it's disabled on gcp (for now). Will make it easier to come back to it later when we know more.

features/base/test/test_umask.py
```diff
@@ -16,5 +16,15 @@
...
+def test_umask(client, file, dict, non_gcp):
+    # Check /etc/login.defs
     file_content(client, file, dict)
+
+def test_umask_cmd(client, non_container):
+    # Additionally check umask via cmd
+    cmd = f"sudo su root -c umask"
```
fwilhe, line 25: Out of curiosity, I stumbled over the `sudo su` idiom. I assume there is a reason to do it, but it's not obvious to me. Not important enough to add a comment I guess.

tests/conftest.py
```diff
@@ -316,8 +316,8 @@ def aws(iaas):
 
 @pytest.fixture
 def non_gcp(iaas):
-    if iaas != 'gcp':
-        pytest.skip('test only supported on gcp')
+    if iaas == 'gcp':
```
fwilhe, line 319: good find. Looks like this was not used anywhere, otherwise it might have enabled/disabled random tests.

fwilhe (approved):
Minor comments, nothing that should stop us from merging.

Thanks.

features/base/test/test_umask.py
Vincinator: The authentication is different
- `sudo su` authentication is handled via whatever sudo permissions the user has. In this case, the user is conceptional already privileged and does not need to enter a password for sudo
- `su` requires a password

features/base/test/test_umask.py
Vincinator: addressed this in latest commit

fwilhe (approved):
Comment got much better compared to before, thanks

PR #2139, opened 2024-06-11
Include checksums and additional tags to S3 artefact uploads

MrBatschner:

During the upload to S3 job of the `nightly` Action:
- md5sums and sha256sums will be calculated for every artefact
- both values will be supplied to S3 to verify proper upload ([ref](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html))
- md5sum and sha256sum will be added to the release manifest
- md5sum and sha256sum will be added as tags to every uploaded artefact (along with Garden Linux version, platform, architecture and comittish)

All this to have the possibility to track and check integrity of image artefacts right from the moment they come out of the build pipeline.

Tested in https://github.com/gardenlinux/gardenlinux/actions/runs/9465059139

Do not cherry-pick to `rel-*` branches as this will require changes to glci as well.

**Release note**:

```feature user
Checksums (md5 and sha256) are now calculated for and added to each release artefact.
```

### DIFF
--- a/.github/workflows/upload_to_s3.sh
+++ b/.github/workflows/upload_to_s3.sh
@@ -37,15 +37,25 @@ EOF
 
 for file in "$name/$name"*; do
 suffix="$(basename "$file" | sed "s/^$name//")"
+md5sum=$(md5sum -b "$file" | cut -d " " -f 1)
+md5bin=$(openssl md5 -binary "$file" | base64 -w0)
+sha256sum=$(sha256sum -b "$file" | cut -d " " -f 1)
+sha256bin=$(openssl sha256 -binary "$file" | base64 -w0)
 cat << EOF >> "$meta_yml"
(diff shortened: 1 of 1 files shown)


fwilhe (approved):
Thanks, looks good.

It took me a moment to realise the difference between the `foosum` and the `foobin` variables. I assume we need the `foobin` encoding for the s3 api, maybe this would deserve a comment as (at least for me) this is not obvious and might lead to errors when someone is touching this code later.

MrBatschner:
@fwilhe: You are right, comments always help. 😉 
Added with  https://github.com/gardenlinux/gardenlinux/pull/2139/commits/496a30557a011a9abedb70428d9ac43c114b3ada

fwilhe:
> comments always help. 😉 

🫠

PR #2143, opened 2024-06-14
fix release page use long commitish

Vincinator:

* GitHub API requires the full commitish to create a release page
* To retrieve the correct metadata of Garden Linux published images from S3, we need to use the short commitish (8 characters). 
* Fixes output format of kmodbuild container


fwilhe (approved):
Didn't run the script locally, but the changes look good. Thanks for taking care.

PR #2167, opened 2024-06-24
feat: better serial console output

nkraetzschmar:
- setup console correctly for qemu virt machine type in arm64 kvm build
- add earlycon/earlyprintk entries working with qemu to kvm feature
- add systemd.journald.forward_to_console=1 to _dev feature

### DIFF
--- a/features/_dev/file.include/etc/kernel/cmdline.d/42-forward-journal.cfg
+++ b/features/_dev/file.include/etc/kernel/cmdline.d/42-forward-journal.cfg
@@ -1,0 +1,1 @@
+CMDLINE_LINUX="$CMDLINE_LINUX systemd.journald.forward_to_console=1"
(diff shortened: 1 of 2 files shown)


5kt (approved):

fwilhe:
I have issues booting images with this commit added. I'm not sure if it is this commit's fault.

@5kt @nkraetzschmar you know this stuff much better, have you verified this builds images that still work fine with `start-vm`?

5kt:
@fwilhe you're right, the issue here arises when updating the kernel cmdline when running outside the build environment - e.g. when disabling ignition after first boot - my commit fixes the issue.

fwilhe:
Yes now I can boot the image but it's not the kind of improvement I was hoping for. Feel free to merge if you're confident it does not break anything.

PR #2173, opened 2024-06-27
Fix start-vm on mac

### LINKED ISSUE #2172: start-vm uses `cp` option that is specific to gnu cp, breaks on macos
Just noticed that this line breaks start-vm on macos because 'cp' is not gnu cp

https://github.com/gardenlinux/gardenlinux/blob/3b4fc5cd87e9d2969dfba25d23f6386ec82523e1/bin/start-vm#L523

```
$ start-vm --skipkp .build/kvm_dev-arm64-today-3b4fc5cd.raw
cp: illegal option -- -
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-aclpSsvXx] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-aclpSsvXx] source_file ... target_directory
```

Possible options to resolve:

- Make gnu coreutils a dependency and call cp as `gcp` on macos (might be an issue if someone has a custom alias for `gcp`)
- find a portable way to do what we're doing now that also works non bsd cp

fwilhe:
start-vm is broken on mac

Fixes #2172

### DIFF
--- a/bin/start-vm
+++ b/bin/start-vm
@@ -520,14 +520,7 @@ qemu_opt_uefi () {
         [ -r $uefi_vars ] || eusage "Missing uefi vars at $uefi_vars.\n Run: apt-get install ovmf qemu-efi-aarch64"
 
         # The source file might be read-only.
-        # cp on macos does not preserve attributes by default
-        if [ ! -e "$CURR_DIR"/"$mac".vars ]; then
-            if [ "$os" = macos ]; then
-                cp "$uefi_vars" "$CURR_DIR"/"$mac".vars
-            else
-                cp --no-preserve=mode,ownership "$uefi_vars" "$CURR_DIR"/"$mac".vars
-            fi
-        fi
+        [ -e $CURR_DIR/$mac.vars ] || cat "$uefi_vars" > "$CURR_DIR"/"$mac".vars
 
         # Add support for amd64 architecture
         if [ "$arch" = amd64 ]; then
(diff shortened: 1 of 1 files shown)


nkraetzschmar:
> If the destination file does not exist, the mode of the source file is used as modified by the file mode creation mask (umask, see csh(1)). [^1]

So this would still preserve ownership on macos. While it likely isn't an issue since the file should be writable in the default brew install setup, it might be better if we always handle the case that the file may have an unwanted permission mode.

How about just a simple `cat "$uefi_vars" > "$CURR_DIR"/"$mac".vars`?
This should work on any os to copy only contents and always using default permission mode.

[^1]: BSD manpage for cp

fwilhe:
useful use of cat? 🐱 

sure if that solves our problem then feel free to push it in a new commit on this pr branch @nkraetzschmar

blitz:
Sorry for breaking MacOS support! :see_no_evil: 

> How about just a simple `cat "$uefi_vars" > "$CURR_DIR"/"$mac".vars`? This should work on any os to copy only contents and always using default permission mode.

That's a great solution. I like it!

The alternative solution is to do `chmod u+r` on the file unconditionally.

blitz (approved):
Thanks for fixing this! :pray:  I'll keep in mind that coreutils might not be GNU next time!

fwilhe:
> Thanks for fixing this! 🙏 I'll keep in mind that coreutils might not be GNU next time!

Thanks for your contribution and your interest in what we're doing :)

fwilhe:
I can't approve my own PR but I've tried @nkraetzschmar's commit so I'll approve by merging.

PR #2174, opened 2024-06-27
Fix/disable IMDSv1

Akendo:

We have to disable the IMDSv1. The test will fail without a token. So we request a token from IMDSv2 and provide it for our current test case.

### DIFF
--- a/features/base/test/test_metadata_connection.py
+++ b/features/base/test/test_metadata_connection.py
@@ -4,9 +4,15 @@ from helper.sshclient import RemoteClient
 
 def test_metadata_connection(client, non_azure, non_ali, non_chroot, non_kvm):
     metadata_host = "169.254.169.254"
+    # request the IMDSv2 token to allow access to the metadata_host on AWS.
+    (exit_code, token, error) = client.execute_command(
+            f"curl -sqX PUT 'http://{metadata_host}/latest/api/token'\
+                    -H 'X-aws-ec2-metadata-token-ttl-seconds:60'"
+    )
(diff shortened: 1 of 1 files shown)


features/base/test/test_metadata_connection.py
```diff
@@ -5,7 +5,9 @@
 def test_metadata_connection(client, non_azure, non_ali, non_chroot, non_kvm):
     metadata_host = "169.254.169.254"
     (exit_code, output, error) = client.execute_command(
-        f"wget --timeout 5 http://{metadata_host}"
+        f"wget --timeout 5 \
+               --header=\"X-aws-ec2-metadata-token: $(curl -sqX PUT 'http://{metadata_host}/latest/api/token' -H 'X-aws-ec2-metadata-token-ttl-seconds: 21600')\" \
```
fwilhe, line 9: From the perspective of readability, i would prefer not to get this token value in a subshell command.
Any reason against splitting this into two? Might make it more readable and more easy to debug in case of failure. Also this can include an assertion on the expected return value of the `api/token` endpoint.

fwilhe (approved):
Approving, I have limited context on this change. If the comment provided on the subshell is useful, feel free to change this and merge then, otherwise ignore it and just merge.

fwilhe:
@Akendo Thanks for taking care. Porting it to `rel-1312` now, and I've noticed you merged with a merge commit. I'm used to squash merge which makes it easier to cherry pick PRs. This is a thing where github remembers what you did last, I'd ask you to do squash merges in the future.

PR #2177, opened 2024-06-28
document package pipeline design

fwilhe:
Add high level documentation for our package pipeline, outlining the design decisions and qualities it brings.

### DIFF
--- a/docs/00_introduction/package-pipeline.md
+++ b/docs/00_introduction/package-pipeline.md
@@ -7,14 +7,14 @@ This document aims to explain the decisions and tradeoffs made.
 ## The bigger picture
 
 Garden Linux release artifacts are disk images for various platforms such as kvm, cloud providers or bare metal hardware.
-To build those artifacts, Garden Linux makes use of an APT repository.
+To build those artifacts, Garden Linux makes use of an [APT repository](https://wiki.debian.org/DebianRepository).
 APT is the Debian package management tool.
 
 Unlike Debian, Garden Linux is focussed on containers.
 This means that for users of Garden Linux, the APT repo has much less direct importance.
 Typically, we don't expect our users to install packages from the APT repository.
 
-Still, for building our release artifacts (the disk images), we need the APT repo.
+The APT repo exists because it is needed for building our release artifacts (the disk images).
 
 ## Qualities of the Garden Linux package pipeline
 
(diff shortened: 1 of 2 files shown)


docs/00_introduction/package-pipeline.md
```diff
@@ -0,0 +1,44 @@
...
+This pipeline might look overly complex on first sight, but is based on it's requirements and past experience.
+This document aims to explain the decisions and tradeoffs made.
+
+## The bigger picture
+
+Garden Linux release artifacts are disk images for various platforms such as kvm, cloud providers or bare metal hardware.
+To build those artifacts, Garden Linux makes use of an [APT repository](https://wiki.debian.org/DebianRepository).
+APT is the Debian package management tool.
+
+Unlike Debian, Garden Linux is focussed on containers.
```
Vincinator, line 13: > Garden Linux is focussed on containers

Claim not relevant for package pipeline documentation.

docs/00_introduction/package-pipeline.md
```diff
@@ -0,0 +1,44 @@
...
+This means that for users of Garden Linux, the APT repo has much less direct importance.
+Typically, we don't expect our users to install packages from the APT repository.
```
Vincinator, line 15: Not relevant. The package pipeline design also works in case we would advertise the apt repos to be used for live-upgrades. which technically should work already if you consider a few configurations (immutability and apt sources).

docs/00_introduction/package-pipeline.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Qualities of the Garden Linux package pipeline
+
+Based on previous experience, the design goals of our package pipeline are the following:
+
+- We want to ship a stable operating system with a modern software stack
+- We want the latest LTS linux kernel with all the Debian benefits as described in [Linux Kernel](./kernel.md)
+- We want to be able to reproduce release artifacts (disk images) and patch individual packages as needed
+- We require updated packages (rebuilt from source, with patches applied) to be absolutely binary compatible with existing releases
+  - This is a challenge because Garden Linux is based on Debian *Testing* where components such as libc or the go compiler might change in incompatible ways without notice
```
Vincinator, line 27: > without notice
reads like this silently fails, but apt package management will explicitly complain about dependency issues.

But, yes, we cannot subscribe to a mailing list, consume an API or stuff like that..

docs/00_introduction/package-pipeline.md
```diff
@@ -0,0 +1,44 @@
...
+- We want the pipeline to work with short-lived (ephemeral) runners, so no long-running "pet" vm is needed for the process
```
Vincinator, line 28: Yes. On a higher level, we want to reduce the dependencies to vendor specific stuff (GitLab or GitHub), and designed the package build based on containers only that could run anywhere, where containers can run (e.g. your local machine, or ephemeral github runners).

docs/00_introduction/package-pipeline.md
```diff
@@ -0,0 +1,44 @@
...
+- We require the APT repos to be consistent. For this reason, APT repo distributions (like `1443.8`) must be immutable.
+- We need to keep the required tooling providers low. In the past, complex setups lead to integration issues. Our new tooling is fully geared towards GitHub, allowing us to make use of the benefits GitHub's tooling provides.
```
Vincinator, line 30: I don't think so. The package pipeline could also be ported back to GitLab with reasonable low effort. Package build stuff is written in bash and build OS is a container. GH Actions only define one way of using the package build ([reference](https://github.com/gardenlinux/package-build/blob/main/.github/workflows/build.yml))

Vincinator (commented):
Good!

A few remarks only. I tried to add a bit of context where my feedback is coming from.

fwilhe:
Thanks for your feedback @Vincinator 

Made some adjustments based on that, hope that's more accurate/relevant now. Re-review appreciated.

PR #2190, opened 2024-07-03
test: add test to check if containerd+runc can pull and spawn an image

Vincinator:
Adds a test to gardener feature, that checks if containerd can pull and spawn a container successfully. 

No issue reported yet. 

In 1312 we have docker.io included and a test using docker to spawn a container. See [test_docker](https://github.com/gardenlinux/gardenlinux/blob/rel-1312/features/base/test/test_docker.py).

However, we can utilise `ctr` which is the cli for containerd to test it in version that do not have a user friendly overlay like podman or docker.

### DIFF
--- a/features/gardener/test/test_containerd.py
+++ b/features/gardener/test/test_containerd.py
@@ -14,6 +14,6 @@ def test_containerd(client, non_chroot):
     (exit_code, output, error) = client.execute_command("sudo ctr image pull ghcr.io/gardenlinux/gardenlinux:nightly")
     assert exit_code == 0, f"no {error=} expected"
 
-    (exit_code, output, error) = client.execute_command("sudo ctr run --rm -t ghcr.io/gardenlinux/gardenlinux:nightly gardenlinux echo 'hello from container'")
+    (exit_code, output, error) = client.execute_command("sudo ctr run --rm -ti ghcr.io/gardenlinux/gardenlinux:nightly gardenlinux echo 'hello from container'")
     assert exit_code == 0, f"no {error=} expected"
     assert output == "hello from container", f"unexpected output: {output}"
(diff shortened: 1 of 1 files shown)


Vincinator:
need #2191 first so this PR can be tested

fwilhe (approved):
Thanks, good call to add this test

Tried the commands locally. I don't think this test is run in the pr workflow, but I think it will be run as part of the nightly once we merge. That is fine from my POV.

Vincinator:
@fwilhe you are right. The tests are not run with dev workflow, because dev workflow only uses ./test, and ./test uses chroot test environment only. Hmm. need to test with kvm locally

Vincinator:
https://github.com/gardenlinux/gardenlinux/actions/runs/9792450496

Vincinator:
https://github.com/gardenlinux/gardenlinux/actions/runs/9793004182

features/gardener/test/test_containerd.py
```diff
@@ -0,0 +1,19 @@
...
+    (exit_code, output, error) = client.execute_command("sudo systemctl enable containerd")
+    assert exit_code == 0, f"no {error=} expected"
+
+    (exit_code, output, error) = client.execute_command("sudo systemctl start containerd")
+    assert exit_code == 0, f"no {error=} expected"
+
+    (exit_code, output, error) = client.execute_command("sudo ctr image pull ghcr.io/gardenlinux/gardenlinux:nightly")
+    assert exit_code == 0, f"no {error=} expected"
+
+    (exit_code, output, error) = client.execute_command("sudo ctr run --rm -ti ghcr.io/gardenlinux/gardenlinux:nightly gardenlinux echo 'hello from container'")
```
fwilhe, line 17: `-ti` will be an issue when running not in an interactive shell afaik

PR #2192, opened 2024-07-04
fix: set umask value even if /etc/login.defs did not set it before

### LINKED ISSUE #2191: test_umask fails in nightly because value is set to 022 and not 027
We set the umask 027 value by searching for existing UMASK value and replace it with 027. however, the /etc/login.defs does not contain a UMASK anymore.
 
https://github.com/gardenlinux/gardenlinux/blob/b4982d519f78feb86007ec79656c8b54d1bac3ad/features/base/exec.config#L25

we need to investigate why UMASK is not set anymore and adapt eatures/base/exec.config to set UMASK if it does not exist 
 

Additionally, also the STIG feature fails.

Vincinator:
Fixes the case when `/etc/login.defs`  does not contain a UMASK value by adding a check, and appending the desired configuration in this case.  

Fixes #2191

### DIFF
--- a/features/base/exec.config
+++ b/features/base/exec.config
@@ -21,9 +21,13 @@ if [ -f /etc/update-motd.d/05-logo ]; then
 fi
 
 # set default umask to a more conservative value
-# inplace sed would change file access permissions
-login_defs="$(sed 's/UMASK\t\t022/UMASK\t\t027/' /etc/login.defs)"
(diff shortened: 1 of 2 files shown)


Vincinator:
@fwilhe as discussed, you can take over this PR and fix the STIG feature with additional commits - thanks!

fwilhe (approved):

PR #2220, opened 2024-07-17
Add azure-nvme-utils to Azure feature

MrBatschner:

The `azure-nvme-utils` package contains udev rules to populate `/dev/disk/azure` with NVMe disks (in addition to traditional SCSI disks) which is required by the Azure CSI driver for Kubernetes to use NVMe disks as persistent volumes.

**Release note**:

```feature user
Package azure-nvme-utils is added to the Azure feature.
```

### DIFF
--- a/features/azure/pkg.include
+++ b/features/azure/pkg.include
@@ -1,3 +1,4 @@
 chrony
 cloud-init
 python3-cffi-backend
+azure-nvme-utils


Vincinator (approved):
Package `azure-nvme-utils` is available in garden linux package repo as of today

PR #2231, opened 2024-07-24
Resize metal /efi to 300MiB

### LINKED ISSUE #1893: Efi Partition too small for Kernel Update
At some point you changed something with the efi Partition. On my latest Garden Linux Machines it's only 65MB large. With the kernel right after installation it's already 34MB/65MB used. Installation of another kernel fails kind of silently (you have to look into the log):

```
...
Setting up linux-image-6.1.68-gardenlinux-amd64 (6.1.68-0gardenlinux2) ...
/etc/kernel/postinst.d/dracut:
dracut: Generating /boot/initrd.img-6.1.68-gardenlinux-amd64
/etc/kernel/postinst.d/zz-systemd-boot:
Installing kernel version 6.1.68-gardenlinux-amd64 in systemd-boot...
install: error writing '/efi/Default/6.1.68-gardenlinux-amd64/initrd': No space left on device
Error: could not copy '/tmp/kernel-install.staging.cbUUJP/initrd' to '/efi/Default/6.1.68-gardenlinux-amd64/initrd'.```
...
```

```
root@gardenlinux:/home/myuser# lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda      8:0    0   30G  0 disk 
├─sda1   8:1    0   65M  0 part /efi
└─sda2   8:2    0 29.9G  0 part /
```

So, apt does not throw an error. after the upgrade no failed upgrade is offered again. After a reboot the old kernel is still used, which makes Kernel Updates basically fail. is there a good reason for it?

tl/dr: 65MB for efi size is in my opinion too small for Kernel Updates, as this example shows. Can you please increase it again? 

(Also: Where in the build script can i define the size of the efi partition myself?)

dnan0s:

Garden Linux images have a default /efi size of `64MiB`. This is not sufficient for more than one kernel, which often prevents Kernel Updates from working. We experienced this with our images and customers also experienced it. I propose to enlarge the /efi partition to `300MiB` have no issues if the need for two (or even three) images arises. 

Fixes #1893

Although this is against the "create a small/minimal immutable image" rule, but i think practically this would help many users in the future.

### DIFF
--- a/features/metal/fstab.mod
+++ b/features/metal/fstab.mod
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+while read -r line; do
+	read -r source target fs options args <<< "$line"
+	if [ "$target" = "/efi" ]; then
+		args="$args,size=300MiB"
+		echo "$source $target $fs $options $args"
+	else
+		echo "$line"
+	fi
+done


Vincinator:
@dnan0s I do not see this in the metal feature directly yet. I assume you have your own Garden Linux image build pipelines and can add a custom feature?

I suggest that you move the changes to a new feature and call it something like `_bigefi` (maybe you can come up with a better name) and include it in the cname when building it, like `metal-kvm_bigefi_prod`. This way you can still consume our changes from metal and other features without code conflicts. Still, you would need to test this of course if your changes still are applied and not "overwriten" by some other feature. 

This PR should not be merged yet to gardenlinux/gardenlinux. If you have a new _bigefi feature (or a feature with a better name), you are very welcome to open a PR to contribute it.

PR #2240, opened 2024-07-31
additional support for NVMe on Azure

MrBatschner:

Adds additional configuration for Azure NVMe according to the guidelines in
https://techcommunity.microsoft.com/t5/running-sap-applications-on-the/converting-virtual-machines-running-linux-from-scsi-to-nvme/ba-p/4162497 
and
https://raw.githubusercontent.com/Azure/SAP-on-Azure-Scripts-and-Utilities/main/NVMe-Preflight-Check/azure-nvme-preflight-check.sh.

**Release note**:

```feature user
additional configuration to support NVMe on Azure was added to the Azure feature
```


Vincinator (approved):
LGTM, lets test with platform test in nightly if it breaks stuff.

PR #2246, opened 2024-08-05
Add script to compare apt repo versions

fwilhe:
Helpful script to investigate differences in repo versions between different Garden Linux versions.

This script will download the apt source file of two given versions, unpack and compare them. Example output:
Screenshot 2024-08-05 at 13 32 35

### DIFF
--- a/hack/compare-apt-repo-versions.sh
+++ b/hack/compare-apt-repo-versions.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 set -o nounset
 set -o errexit
-set -o pipefail
 
 SCRIPT_NAME="${0##*/}"
 readonly SCRIPT_NAME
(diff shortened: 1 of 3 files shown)


hack/compare-apt-repo-versions.sh
```diff
@@ -0,0 +1,38 @@
...
+
+    trap 'rm -rf $TEMP_DIR' EXIT
+
+    curl -s https://packages.gardenlinux.io/gardenlinux/dists/"$VERSION_A"/main/binary-amd64/Packages.gz | gunzip > "$TEMP_DIR"/"$VERSION_A"
+    curl -s https://packages.gardenlinux.io/gardenlinux/dists/"$VERSION_B"/main/binary-amd64/Packages.gz | gunzip > "$TEMP_DIR"/"$VERSION_B"
+
+    python3 "$SCRIPT_DIR"/parse-aptsource.py "$TEMP_DIR"/"$VERSION_A" > "$TEMP_DIR"/a
+    python3 "$SCRIPT_DIR"/parse-aptsource.py "$TEMP_DIR"/"$VERSION_B" > "$TEMP_DIR"/b
+
+    diff "$TEMP_DIR"/a "$TEMP_DIR"/b
```
Vincinator, line 34:
```suggestion
    diff -U 0 --minimal  "$TEMP_DIR"/a "$TEMP_DIR"/b | grep -v '^@'
```

I suggest a different output format that is a bit easier to read and parse.

Vincinator: ` -U 0`:  removes the context lines that did not change and uses the `+ -` format 

`--minimal`:  not sure if we need it, but it sounds cool 
> -d, --minimal
              try hard to find a smaller set of changes

` | grep -v '^@'`: removes lines starting with @, which are part of output from diff to indicate where a "chunk" starts

Vincinator: We could further improve usability by adding more flags to this tool.

`--ignore-removed`: does a `| grep -v '^-'` to remove lines that show what packages were 
`--only-new`: only shows packages that did not exist before 
`--only-version-changes`: only shows packages that existed before in a different version

just food for thought. Not an issue.

Vincinator (approved):
Useful tool! feel free to merge. Don't see any blocker, just some feedback on preferences

hack/compare-apt-repo-versions.sh
fwilhe: Thanks for your suggestions. Another option I was thinking about is `side-by-side`.. does not work with `U 0` but might also be useful.

Screenshot 2024-08-06 at 14 14 26

I'd like to merge it with your suggestion and then evolve the tool as we see fit. I'd like to avoid too complex argument parsing, but some more convenience features would be welcome for sure.

PR #2255, opened 2024-08-06
Switch from fail2ban to sshguard

### LINKED ISSUE #2221: Implement enabling PerSourcePenalties with SSHguard
The purpose of this issue is to follow up on https://github.com/gardenlinux/security/issues/26 with configuring Garden Linux going forward to use PerSourcePenalties with SSHguard.

@gehoern will go for it.

CC @Akendo

fwilhe:
The decision was made to switch from fail2ban to sshguard.
This PR replaces the packages and removes fail2ban config.

Also the Garden Linux apt repo from tomorrow on contains  openssh 9.8p1-2 which includes the PerSourcePenalties.
This is enabled by default, custom configuration can be made.

Fixes #2221

**Reviewer notes:**

I'm not sure if we need special configuration for sshguard. We had quite some config for fail2ban which is deleted in this PR.

I did not find any specific tests for fail2ban, not sure if we need tests for sshguard.


VERSION
```diff
@@ -1 +1 @@
-today
```
fwilhe, line 1: needs to be reverted once sshguard is in the today dist (tomorrow)

features/ssh/exec.config
```diff
@@ -1,5 +1,34 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
+set -x
```
Vincinator, line 3: do we need this?

Vincinator, line 3:
```suggestion
```

fwilhe: thanks, forgot to remove after debugging

5kt (approved):

PR #2256, opened 2024-08-06
Include sbom information in release notes

fwilhe:
Add a list of selected packages with their respective versions to the release notes workflow.
This should give Garden Linux developers and users a way to quickly identify the versions of relevant components in a specific Garden Linux version.


Vincinator (approved):
LGTM, creating a follow up PR with general improvements unrelated to this PR.
I can not test this without creating an actual release or modifying the release_note.py.

Will refactor it, so the release note content can be created in a separate step from creating a GH release and writing that content string to the body.

PR #2270, opened 2024-08-13
Release note dry run

Vincinator:
Adds a simple dry-run flag, so that we can re-generate the release note locally without uploading changes directly to a release in GitHub. 

Follow up for #2256


fwilhe (approved):
Good idea. 👍  I've changed the script so I could run it manually, but that was pretty hacky. Much better that way.

fwilhe:
it fails for me. am I using it wrong? Dry run should not attempt to download images, right?

Screenshot 2024-08-13 at 09 20 36

PR #2275, opened 2024-08-14
revert changes from #1874

fwilhe:
Simplify the build by removing the dockerfile indirection.
This was intended to be used with dependabot but never really worked with how the builder image is published, so we can just remove it to simplify.


Vincinator (approved):
👍🏻

PR #2283, opened 2024-08-16
Add Script to compare the apt repo version of Garden Linux with Debian

fwilhe:
Script might be useful to chose which packages to upgrade in Garden Linux patch releases

Run the script like this `./hack/compare-apt-repo-versions-gl-debian.sh 1592.0`

Output looks like this
```
base-files 13.3 13.5
bash 5.2.21-2.1 5.2.21-2.1+b1
binutils 2.43-1 2.43-2
```

where the first version number is the version in Garden Linux, the right version number is the version in debian testing as of now.

### DIFF
--- a/hack/compare-apt-repo-versions-gl-debian.sh
+++ b/hack/compare-apt-repo-versions-gl-debian.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+set -o errexit
+
+SCRIPT_NAME="${0##*/}"
+readonly SCRIPT_NAME
+
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
+readonly SCRIPT_DIR
+
+usage() {
+    echo "Script to compare the apt repo version of Garden Linux with Debian."
+    echo "Usage: $SCRIPT_NAME GL_VERSION"
+    echo "Example: $SCRIPT_NAME 1599.0"
+    exit 1
+}
+
+main() {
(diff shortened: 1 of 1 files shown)


Vincinator (approved):
Good stuff. I was also thinking about writing a tool for this. 
Thanks, I like it 👍🏻 

I approve, since it is a good addition to the `hack/` scripts and tools, and we can improve on it whenever. 

Dumping out my thoughts, but feel free to dismiss them for now: 
- Only packages from `binary-amd64` are compared with this script. Would be nice to have the option to compare packages in `binary-arm64` and `binary-all` as well. 
- We can generalise this tool to compare any two given apt dists. 
    - compare two Garden Linux dists (what [compare-apt-repo-versions.sh](https://github.com/gardenlinux/gardenlinux/blob/main/hack/compare-apt-repo-versions.sh) does)
    - compare with debian snapshot repos
    - compare with other debian dist than testing 
    - compare with ubuntu dists, does it explode?

Feel free to merge!

fwilhe:
Thanks @Vincinator 

I think there might be a few other improvements, for example it makes not much sense to have the kernel in there as we follow the LTS versions and comparing to debian's kernel version does not tell you much.

Also maybe we should filter packages we build from upstream?

I think it makes sense to use it and see where we can improve and/or generalise the scripts.

Just quickly also tested the script in a debian vm, just to make sure, will remove the comment and merge.

PR #2292, opened 2024-08-20
integrate oci delivery

Vincinator:
Provides a workflow for OCI publishing 

Fixes https://github.com/gardenlinux/python-gardenlinux-lib/issues/15

Todos:
- [x] check if the image targets are actually published
- [ ] use correct pub/priv keypair for signing


mxmxchere:
I will close this one in favor of #2370 The diff became too old to be reviewable nicely.

Issue #2296, opened 2024-08-26
Maintain (active) releases in machine readable form


fwilhe:
**What would you like to be added**:
The linux kernel maintains a file at https://www.kernel.org/releases.json where releases are present in a machine-readable form.
This is pretty nice for automation purposes.
We could do something similar, creating such a file could be done by our release-notes workflow.

**Why is this needed**:
In automation, we sometimes need to know (maintained) Garden Linux versions.
We can use git tags or github releases in this repo, but this is missing some information/meta data.

This could be useful to provide a similar view to https://packages.debian.org/search?suite=all&searchon=names&keywords=wget where we want to know the package versions present in (relevant) Garden Linux versions.

This could also be useful for GLVD.

Just a rough sketch of what this file could look like:

```json
{
    "releases": [
        {
            "moniker": "something-needs-to-be-aligned",
            "version": "1592.1",
            "iseol": false,
            "released": {
                "timestamp": 1724569631,
                "isodate": "2024-08-25"
            }
        },
        {
            "moniker": "something-needs-to-be-aligned",
            "version": "1592.0",
            "iseol": true,
            "released": {
                "timestamp": 1724569631,
                "isodate": "2024-08-25"
            }
        },
        {
            "moniker": "something-needs-to-be-aligned",
            "version": "1443.10",
            "iseol": false,
            "released": {
                "timestamp": 1724569631,
                "isodate": "2024-08-25"
            }
        }
  ]
}
```

pnpavlov:
@yeoldegrove would that help you instead of :latest cat /etc/os-release?

yeoldegrove:
@pnpavlov @fwilhe

https://github.com/gardenlinux/gardenlinux/commit/87e2dc920e906345d19b19a40c78b08d5af8e389#diff-9dd0c2a02c4e49d2c385f2033fa07902c78b11ef7013e8fbfbf90d19d6e6057c would be a quick hack to parse `ghcr.io/gardenlinux/gardenlinux:latest` and extract the actual release tag.

```
❯ bin/garden-version-latest
1592.1
```

pnpavlov:
@yeoldegrove , please check if this is in same direction as your current work? Can we do that?

fwilhe:
Alternative suggestion for file layout

The idea here would be to group releases by major versions (one new every day) and to mark which of those are 'maintained', i.e. get any patch releases. I'm not sure if this adds much value, or if we should leave out those nightly releases.

```json
{
[...]

PR #2307, opened 2024-08-29
Initial releases.json file

fwilhe:
This file is supposed to be updated automatically when a new release is performed. It should always contain up-to-date information on maintained Garden Linux releases. This is relevant for glvd and it might also be useful for other tooling.

There are ways to find out about existing releases (for example git tags, github releases, ..) but they don't provide life-cycle information (for example if a version is eol or not).

This file should have more fields indicating the exact status of a release, and that needs to be updated automatically (for example by the 'create release notes' workflow).

see https://github.com/gardenlinux/gardenlinux/issues/2296

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1,0 +1,16 @@
+{
+    "releases": [
+        {
+            "version": "1592.1",
+            "iseol": false,
+        },
+        {
+            "version": "1592.0",
+            "iseol": true,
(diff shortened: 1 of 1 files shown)


fwilhe:
Obsolete thanks to @yeoldegrove and his work on glrd

PR #2331, opened 2024-09-12
ci: update actions/upload-artifact and actions/download-artifact to v4

yeoldegrove:

v3 is scheduled for deprecation on November 30, 2024

There are breaking changes:
- https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes
- https://github.com/actions/download-artifact?tab=readme-ov-file#breaking-changes

checked:
- We are not uploading to the same named Artifact multiple times.
- I think we do not have the use case of downloading artifacts that were created from action/upload-artifact@v3 and below are not supported.

### DIFF
--- a/.github/workflows/stig.yaml
+++ b/.github/workflows/stig.yaml
@@ -20,7 +20,7 @@ jobs:
       - name: Build the image
         run: ./build ${{ matrix.platform}}-stig
       - name: Upload build logs
-        uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # pin@v4.4.0
         with:
           name: STIG-${{ matrix.platform}}-logs
           path: .build/*log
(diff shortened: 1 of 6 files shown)


.github/workflows/stig.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Build the image
         run: ./build ${{ matrix.platform}}-stig
       - name: Upload build logs
-        uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # pin@v4.4.0
```
fwilhe, line 23:
```suggestion
        uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # pin@v4.4.0
```

wait a second, this introduces a syntax error

fwilhe: fixed

fwilhe (approved):
looks good to me, thanks

PR #2340, opened 2024-09-16
cleanup: remove _secureboot, _readonly, _trustedboot

nkraetzschmar:
update to new builder and remove features no longer supported (will be replaced following this)


fwilhe:
nitpick: 'deprecate' is misleading when what we're actually doing is removing. I don't know if anyone uses those features, but a more obvious heads up to potential users would be nice, imo.

build
```diff
@@ -6,11 +6,12 @@ shopt -s nullglob
 exec 3>&1
 exec 1>&2
 
-container_image=ghcr.io/gardenlinux/builder:e2dff5417ff2c6220c4226c585c006053957de5e
+container_image=ghcr.io/gardenlinux/builder:357fbe01a5a95854ccb5065c28fc6eca13466a46
 container_engine=podman
 target_dir=.build
 
 container_run_opts=(
+	--memory 4G
```
fwilhe, line 14: out of curiosity: what is this change about? It seems unrelated to the deletion of features, and I'd like to understand why it's there. I'm okay with having it in this pr.

fwilhe (commented):
I'm assuming you are aware of the ci failures, not ready yet for merge.

build
nkraetzschmar: was some change in the builder a while ago, and looks like we just never pulled it in

Akendo (approved):

PR #2342, opened 2024-09-16
feat: add _trustedboot and _ephemeral features

nkraetzschmar:
_trustedboot provides the core trustedboot functionality; it requires an additional feature to provide handling of mutable data

_ephemeral provides the default implementation of this, it is implemented in a seperate feature to allow overriding with alternatives (using feature excludes)


features/_trustedboot/file.include/etc/update-motd.d/25-secureboot
```diff
@@ -0,0 +1,16 @@
...
+efivars_path=/sys/firmware/efi/efivars/SecureBoot-8be4df61-93ca-11d2-aa0d-00e098032b8c
+
+if [ -f "$efivars_path" ]; then
+	secureboot_state="$(od -An -j4 -N1 -t u1 "$efivars_path" | tr -d ' ')"
+fi
+
+if [ "$secureboot_state" = 1 ]; then
+	printf 'secureboot: \e[32menabled\e[0m\n'
+else
+	printf '\n    \e[37;41m┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓\e[0m\n    \e[37;41m┃                                    ┃\e[0m\n    \e[37;41m┃    !!! secureboot disabled !!!     ┃\e[0m\n    \e[37;41m┃                                    ┃\e[0m\n    \e[37;41m┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛\e[0m\n\n'
```
Akendo, line 15: Should we not have this called trustedboot?

Akendo (changes requested):
Please add some `README.md` to the features. Take a look at [kvm](https://github.com/gardenlinux/gardenlinux/tree/b243a1726a14c8dd070b0319b0e4c09b82f9c32e/features/kvm)

features/_trustedboot/exec.post
```diff
@@ -0,0 +1,17 @@
...
+
+mkdir -p "$rootfs/etc/gardenlinux"
+
+rm -rf "$rootfs/opt"
+ln -sT /var/opt "$rootfs/opt"
+
+rm -rf "$rootfs/home"
+ln -sT /var/home "$rootfs/home"
+
+for key in pk null.pk kek db; do
```
Akendo, line 15: Where do we get this files from?

features/_trustedboot/file.include/etc/kernel/cmdline.d/99-no-gpt-auto.cfg
Akendo, line 1: Why do we disable the `systemd.gpt_auto` feature?

features/_ephemeral/file.include/etc/systemd/system/ephemeral-cryptsetup.service
```diff
@@ -0,0 +1,12 @@
...
+Requires=systemd-repart.service
+After=systemd-repart.service
+DefaultDependencies=no
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/cryptsetup open --type plain --cipher aes-xts-plain64 --key-size 512 --key-file /dev/urandom /dev/disk/by-partlabel/EPHEMERAL ephemeral
+ExecStart=/usr/bin/dd if=/dev/zero of=/dev/mapper/ephemeral bs=1M count=1
+ExecStart=/usr/sbin/mkfs.ext4 -O '^has_journal' /dev/mapper/ephemeral
```
Akendo, line 11: Why do we need this option?

features/_trustedboot/convert.raw~esp.tar
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+input="$1"
+output="$2"
+
+esp_dir="$(mktemp -d)"
+mount -t tmpfs -o size="$TEMPFS_SIZE" tmpfs "$esp_dir"
```
Akendo, line 9: Where do we set `TEMPFS_SIZE`?

features/_trustedboot/file.include/etc/tmpfiles.d/var-home.conf
```diff
@@ -0,0 +1 @@
+d /var/home 0755 - - -
```
Akendo, line 1: This seems too permissive for a home folder.

features/_trustedboot/file.include/etc/update-motd.d/25-secureboot
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -eufo pipefail
+
+secureboot_state=0
+efivars_path=/sys/firmware/efi/efivars/SecureBoot-8be4df61-93ca-11d2-aa0d-00e098032b8c
```
Akendo, line 6: Could that path change?

features/_trustedboot/file.include/usr/bin/repart-esp-disk
```diff
@@ -0,0 +1,13 @@
...
+
+set -eufo pipefail
+
+efi_part_dev="$(realpath "$1")"
+efi_part_name="$(basename "$efi_part_dev")"
+
+disk_name="$(basename "$(realpath "/sys/class/block/$efi_part_name/..")")"
+disk_dev="/dev/$disk_name"
+
+echo | sfdisk "$disk_dev"
```
Akendo, line 11: Are we creating an empty partition here? Maybe some expiation could be helpful why.

features/_trustedboot/file.include/usr/lib/dracut/modules.d/99sysroot-loop/module-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+	What=/root.img
+	Where=/sysroot
+	Type=erofs
+	EOF
+
+	mkdir -p "$initdir/etc/systemd/system/initrd-root-fs.target.requires"
+	ln -s "$initdir/etc/systemd/system/sysroot.mount" "$initdir/etc/systemd/system/initrd-root-fs.target.requires/"
+
+	inst /root.img /root.img
+	instmods erofs
```
Akendo, line 26: Just as a logic, should we not start with a modprobe first?

features/_trustedboot/file.include/usr/lib/systemd/system-generators/detect-disk-by-efivars
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+set -eufo pipefail
+
+efi_part_uuid="$(cat /sys/firmware/efi/efivars/LoaderDevicePartUUID-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f | tail -c +4 | tr -cd '[:print:]' | tr '[:upper:]' '[:lower:]')"
```
Akendo, line 5: Can this part change?

features/_trustedboot/file.include/etc/kernel/cmdline.d/99-no-rd-shell.cfg
Akendo, line 1: Maybe you add why we do not like to have this here. It might be not, obviously.

features/_trustedboot/image.esp.tar
```diff
@@ -0,0 +1,144 @@
...
+
+mkdir "$esp_dir/EFI/Linux"
+mv "$uki_signed" "$esp_dir/EFI/Linux/$BUILDER_CNAME+3.efi"
+
+esp_tar="$(mktemp)"
+
+find "$esp_dir" -mindepth 1 -printf '%P\n' | tar --create --mtime="@$BUILDER_TIMESTAMP" --sort name --numeric-owner --pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime --directory "$esp_dir" --no-recursion -T - > "$esp_tar"
+rm -rf "$esp_dir"
+
+cp --sparse always "$esp_tar" "$output"
```
Akendo, line 141: Why do we need this?

features/_trustedboot/image.esp.tar
Akendo, line 1: I have a hard time to understand when this file is invoked. I would love to see this with some function and a main function. I feel like that main steps can be put into function without much effort.

features/_trustedboot/convert.raw~esp.tar
Akendo, line 1: Maybe we can make the code into some function and add it into a main function?

Akendo (changes requested):
First, this is great work! 

I've added some more comments. This change is huge and not simple to get. I might need some more details to get this.

features/_trustedboot/file.include/etc/update-motd.d/25-secureboot
nkraetzschmar: nope, our naming is that trustedboot = secureboot + immutability. the feature has both, this check is just a runtime check to test if the host we are running on was actually bootet with sec boot enabled.

features/_trustedboot/convert.raw~esp.tar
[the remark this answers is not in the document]
nkraetzschmar: no, the suffix here is explicitly required by the builder, the suffix of the convert script is used to determine which script to use to produce each target artifact

nkraetzschmar:
> Please add some `README.md` to the features. Take a look at [kvm](https://github.com/gardenlinux/gardenlinux/tree/b243a1726a14c8dd070b0319b0e4c09b82f9c32e/features/kvm)

that's part of #2344 since it is a combined doc for both trustedboot and tpm, so would make no sense to have this here already

features/_trustedboot/exec.post
nkraetzschmar: builder maps in the cert folder

features/_trustedboot/file.include/etc/kernel/cmdline.d/99-no-gpt-auto.cfg
nkraetzschmar: to have explicit control over what gets mounted where. otherwise an attacker could conceivably create a malicious disk with our ESP and an attacker controlled usr partition that would be detected by systemd gpt auto fstab generator => mounted over our root => inject mallicious replacements for binaries in the boot chain.

plus we don't need it anyways, so should also marginally reduce bootup latency

features/_ephemeral/file.include/etc/systemd/system/ephemeral-cryptsetup.service
nkraetzschmar: We don't strictly need it, but this is a performance tuning measure. the disk get's discarded on each boot anyways so keeping a journal only adds unnecessary overhead and decreases write performance

features/_trustedboot/convert.raw~esp.tar
nkraetzschmar: builder

nkraetzschmar: [builder](https://github.com/gardenlinux/builder/blob/357fbe01a5a95854ccb5065c28fc6eca13466a46/build#L121)

features/_trustedboot/file.include/etc/update-motd.d/25-secureboot
nkraetzschmar: nope, that's how efivars work, they are store in firmware under 2 keys, vedor UUID + name, and the linux kernel maps this to `$name-$uuid` in the efivars fs

only if the kernel ever decides to change this, but "don't break userspace", so not likely
the secureboot UUID is part of EFI spec, so this won't change unless some new EFI version with new incompatible spec

features/_trustedboot/file.include/usr/bin/repart-esp-disk
nkraetzschmar: This is to handle a resized disk, so when our small disk image got imaged to a larger disk. now the secondary GPT table won't be at the end of the disk anymore.

we are essentially applying an identity transform on the disk with sfdisk, however, even when not making changes sfdisk will always fix up the secondary GPT table to be at the end of the disk.

nkraetzschmar: added a comment to clarify

features/_trustedboot/file.include/usr/lib/dracut/modules.d/99sysroot-loop/module-setup.sh
nkraetzschmar: that should all be handled by dracut, this is not code that is actually run on the target system, just the logic dracut uses to generate the initrd. instmods here is some dracut specific magic, dracut handles all the details for us.

features/_trustedboot/file.include/usr/lib/systemd/system-generators/detect-disk-by-efivars
nkraetzschmar: same as with the secboot vars, except these are defined by the systemd / UAPI group boot loader spec. so far more likely to change than the EFI spec, but still small enough chance to bet on it being fixed.

features/_trustedboot/file.include/etc/kernel/cmdline.d/99-no-rd-shell.cfg
nkraetzschmar: was a finding by anvil in the sec audit of the old secboot / readonly features. if you can somehow cause the boot process to glitch (was easily feasible in the old design, probably way harder now) and you get dropped into an emergency shell you could use it to get around the trustedboot architecture by now just mounting whatever you want with full access to the machine and even worse the TPM state

features/_trustedboot/image.esp.tar
nkraetzschmar: we could also just directly write the tar to $output, but this way, if there are any sparse files inside the tar (quite likely) they well also be stored spares on disk inside the tar. otherwise tar will happily write explicit zero bytes there rather than allocate holes

Akendo (approved):
Thanks for all the info. Thinks make sense.

PR #2346, opened 2024-09-16
add aws_trustedboot and aws_trustedboot_tpm2 to platform tests

nkraetzschmar:


features/base/test/test_basics.py
```diff
@@ -40,7 +40,7 @@ def test_ls(client):
 
 def test_startup_time(client, non_chroot, non_kvm, non_azure):
     """ Test for startup time """
-    tolerated_kernel_time = 15
```
Akendo, line 43: Please add why this is the case, since yestday I know why. But let other know.

Akendo (approved):

features/base/test/test_basics.py
nkraetzschmar: since the initrd now contains the entire rootfs (although compressed) the kernel naturally takes longer before giving control to userspace, since it needs to unpack the initrd first. this should not negatively impact total boot time though (in all test boot time was the same or better) since we are only moving tasks from user space to kernel space.

PR #2360, opened 2024-09-19
feat: enable trustedboot tests on AWS arm

nkraetzschmar:


Akendo:
I may get here the change wrong, but you're just removing things? Is this change complete or still a draft?

nkraetzschmar:
it's complete, we used to exclude running trustedboot tests on arm since aws did not support tpm yet on arm, but now they do, so might as well enable it

nkraetzschmar:
no longer needed, replaced by https://github.com/gardenlinux/gardenlinux/blob/f8b532ff0a3e4e86a412cebce41038d01b379530/flavors.yaml#L73

PR #2383, opened 2024-10-01
Add feature to install glvd client and enable summary in motd

fwilhe:

Add feature to install glvd client and enable summary in motd

Example screenshot:
glvd-motd

Fixes https://github.com/gardenlinux/glvd/issues/118


Akendo (approved):
Looks good.

features/glvd/file.include/etc/update-motd.d/99-glvd
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+if [ -x /usr/bin/glvd ]; then
+    exec /usr/bin/glvd executive-summary
```
fwilhe, line 4: comment from a private chat:

this might be slow and thus undesired

should have an aggressive timeout to better not show anything than delay operations for multiple seconds

could make use of caching

PR #2409, opened 2024-10-15
 create releases in GLRD for nightly and patch releases

yeoldegrove:

- This PR introduces wrapper scripts to run `bin/glrd` and `bin/glrd-create` from https://github.com/gardenlinux/glrd within Makefiles or workflows.
- nightly builds now creates GLRD releases [f223b59](https://github.com/gardenlinux/gardenlinux/pull/2409/commits/e5b11a9f72dd4d1c55c16f66f3c1cbfe6956cf91)
- patch releases now create GLRD releases [88e1352](https://github.com/gardenlinux/gardenlinux/pull/2409/commits/ff32cc7aedafd218e8e693ed843426cce802cb58)
  - the unused `beta` releases are removed and `dev` releases are possible now
  - No github releases will be created for `dev` releases, only a GLRD entry.


bin/glrd
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+cmd=$(basename "${0}")
+args="$*"
+if [ -z "${GARDENLINUX_BUILD_CRE}" ]; then
+    GARDENLINUX_BUILD_CRE="sudo podman"
```
fwilhe, line 5: why is sudo required here (by default)?

bin/glrd-manage
```diff
@@ -0,0 +1 @@
+glrd
```
fwilhe, line 1: what's the use-case for this script?

fwilhe (commented):
looks good on a first sight, leaving some time for others to review

bin/glrd
yeoldegrove: This is "comman practice" in other places, too:
https://github.com/gardenlinux/gardenlinux/blob/c8b506dd79bd1fc6f8f10a7bf3e2d2eef625f3a9/bin/start-vm#L98
https://github.com/gardenlinux/gardenlinux/blob/c8b506dd79bd1fc6f8f10a7bf3e2d2eef625f3a9/.github/workflows/platform-test-images.yml#L83
https://github.com/gardenlinux/gardenlinux/blob/c8b506dd79bd1fc6f8f10a7bf3e2d2eef625f3a9/.github/workflows/build_container.yml#L19

Just adopted it. Should be changed globally if we don't need this feature anymore. I am not sure who needs to run this as root.

bin/glrd-manage
yeoldegrove: `bin/glrd-manage` is just a symlink to `bin/glrd`. It has some magic in it ...
https://github.com/gardenlinux/gardenlinux/pull/2409/files#diff-ff3a914bba97885ac56a104c745a59a74bae8df634b5cf25396dc029f17e98bfR2
https://github.com/gardenlinux/gardenlinux/pull/2409/files#diff-ff3a914bba97885ac56a104c745a59a74bae8df634b5cf25396dc029f17e98bfR15
... which enables us to use the same code (as in the link target) without having a very similar file twice.
It basically calls the `${cmd}` in the container which is gotten from the filename/basename of `bin/glrd-manage` OR `bin/glrd`.

fwilhe (approved):
LGTM

PR #2410, opened 2024-10-16
adaptions and clarifications on "Garden Linux Release Plan Overview"

yeoldegrove:

Here a few adaptions and clarifications, based on the discussions we had this week.

### DIFF
--- a/docs/00_introduction/release.md
+++ b/docs/00_introduction/release.md
@@ -6,17 +6,29 @@ On this page you will learn about the Garden Linux releases and their lifecycle.
 
 ### Stable releases
 
-You can expect up to four Garden Linux releases per year, published up to once per quarter. A new release _(e.g. 1443.0 or "release 1443")_ includes major changes like newer major version of software package or library. Within a release maintenance period, changes are published as release patches _(e.g. 1443.1 or "patch 1 of release 1443")_. Each release is maintained nine months, while going through following maintenance phases:
+You can expect up to four Garden Linux stable releases per year, published up to once per quarter. Each new stable release (e.g., "1443" or "stable-1443" or "release 1443") introduces major changes like updated versions of software packages or libraries. The initial stable release is designated as "patch 0" (e.g., "1443.0" or "patch-1443.0" or "initial release 1443"), establishing the initial frozen codebase. During the nine-month maintenance period, updates are provided as release patches (e.g., "1443.1" or "patch-1443.1" or "patch 1 of release 1443"). Each stable release progresses through the following maintenance phases:
 
-- **Standard maintenance** starts with the publishing of a new stable release and continues for six months. In this phase active maintenance is performed on the release. You can expect regular patches including _(but not limited to)_ fixes for issues, updates to higher minor release of selected software packages and libraries included, proactive monitoring & assessment of security vulnerabilities, performance corrections and enhancements and more. General best practice is to always use the latest patch of the latest release.
+- **Standard maintenance** starts with the publishing of a new stable release and continues for six months. In this phase active maintenance is performed on the stable release. You can expect regular patch releases including _(but not limited to)_ fixes for issues, updates to higher minor release of selected software packages and libraries included, proactive monitoring & assessment of security vulnerabilities, performance corrections and enhancements and more. General best practice is to always use the latest patch of the latest release.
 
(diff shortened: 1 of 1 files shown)


pnpavlov (approved):
OK with me, thank you!

docs/00_introduction/release.md
```diff
@@ -6,17 +6,29 @@ On this page you will learn about the Garden Linux releases and their lifecycle.
...
+Every night, a new build is triggered via the GitHub Action [Nightly Build](https://github.com/gardenlinux/gardenlinux/actions/workflows/nightly.yml) (e.g., "nightly-1660.0"). You can monitor their status publicly on the action page. The maintainers aim to have a successful build every 24 hours. Occasionally, there might be nightly patches (like "nightly-1660.1") released if the initial nightly build requires urgent fixes. Nightly releases contain the latest software packages and are geared toward rapid iterations for advanced stakeholders interested in cutting-edge changes. However, they are not guaranteed to be as stable or as thoroughly maintained and tested as the stable releases. Nightly releases do not include maintenance commitments and are therefore not recommended for production environments. Up to once per quarter, a nightly release is promoted to a stable release and is then supported with nine months of standard maintenance.
+
+### Next Release
+
+The "next" release is a projection of the upcoming stable release. It exists primarily to give users an idea of when the next stable version **might** be released and how long it **might** be supported.
```
fwilhe, line 31: that's new to me, what is that? build from an experimental repo?

docs/00_introduction/release.md
```diff
@@ -6,17 +6,29 @@ On this page you will learn about the Garden Linux releases and their lifecycle.
...
+Every night, a new build is triggered via the GitHub Action [Nightly Build](https://github.com/gardenlinux/gardenlinux/actions/workflows/nightly.yml) (e.g., "nightly-1660.0"). You can monitor their status publicly on the action page. The maintainers aim to have a successful build every 24 hours. Occasionally, there might be nightly patches (like "nightly-1660.1") released if the initial nightly build requires urgent fixes. Nightly releases contain the latest software packages and are geared toward rapid iterations for advanced stakeholders interested in cutting-edge changes. However, they are not guaranteed to be as stable or as thoroughly maintained and tested as the stable releases. Nightly releases do not include maintenance commitments and are therefore not recommended for production environments. Up to once per quarter, a nightly release is promoted to a stable release and is then supported with nine months of standard maintenance.
```
fwilhe, line 27: > Occasionally, there might be nightly patches (like "nightly-1660.1") released if the initial nightly build requires urgent fixes.

really? not aware of this so far.

docs/00_introduction/release.md
yeoldegrove: This did exist in here already ...
  - https://github.com/gardenlinux/gardenlinux/blob/f27744b247fb05b858fbfcfa272196a871094168/docs/00_introduction/release.md?plain=1#L35-L37
  - https://github.com/gardenlinux/gardenlinux/blob/f27744b247fb05b858fbfcfa272196a871094168/docs/00_introduction/release.md?plain=1#L55

... and I basically just defined it.

Is "a projection of the upcoming stable release" not clear enough? We could name this a "virtual release" to make it clearer that it is not something that you can use.

docs/00_introduction/release.md
yeoldegrove: This is some feature I want to add with https://github.com/gardenlinux/glrd?tab=readme-ov-file#nightly-releases.

This makes it possible to refer to a nightly that failed as e.g. `nightly-1660.0`... fix it ... re-run the nightly pipeline ... automatically name it `nightly-1660.1` if it succeeded.

We might decide not to release a failed nightly (which is fine, but needs to be aligned on). Anyways, as nightlies share the same schema as patches (I thought this makes sense), having patch releases would be possible in theory and I wanted to clarify this in the doc.

docs/00_introduction/release.md
fwilhe: I understood it as 'there are artifacts with next-xy names' but that's not the case, right? Maybe something @pnpavlov can clarify? I'm fine with having that in there but it seems like something that does not really exist to me so maybe we can just get rid of it?

If it's useful then I'm also fine with keeping it in, I'm just not aware of the value.

docs/00_introduction/release.md
fwilhe: hm.. this might give the false impression that those nightlies are maintained..? not sure. @pnpavlov what's your position on this?

yeoldegrove:
@pnpavlov As discussed, I will merge this "as is", please open a new issue if there need to further adoptions.

PR #2411, opened 2024-10-16
Selectable region to clean_orphans_aws.py

MrBatschner:

Allows to select the AWS region for clean_orphans_aws.py from the command line.

**Release note**:

```feature user
Allows to select the AWS region for clean_orphans_aws.py from the command line.
```


Akendo (approved):
LGTM

PR #2421, opened 2024-10-23
Bootstrap PR template

fwilhe:
Add a definition of done to remind about tests, docs and potentially needed backports.

I want to keep this lean, but for sure that dod is not 'done'. Feedback is welcome.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,6 +5,7 @@ Fixes #
 
 **Definition of Done:**
 - [ ] The code is sufficiently documented
+- [ ] Shared the changes with the Team so everyone is aware
 - [ ] The code is appropriately tested
 - [ ] If needed, the code is backported to the release branches of all currently maintained releases
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,25 @@
+**What this PR does / why we need it**:
+
+**Which issue(s) this PR fixes**:
+Fixes #
+
+**Definition of Done:**
+- [ ] The code is sufficiently documented
+- [ ] The code is appropriately tested
+- [ ] If needed, the code is backported to the release branches of all currently maintained releases
+
+**Special notes for your reviewer**:
+
+**Release note**:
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,25 @@
+**What this PR does / why we need it**:
+
+**Which issue(s) this PR fixes**:
+Fixes #
+
+**Definition of Done:**
+- [ ] The code is sufficiently documented
+- [ ] The code is appropriately tested
```
Akendo, line 8: We maybe should add that the code was reviewed/merge or has a different person was made aware of. Currently, most of the time this knowledge is earlier lost since even when things getting documented, it's challenging to find it.

fwilhe: I don't understand this suggestion. We have a pr workflow. If people actually take the time to review is another question.

Can you suggest into an item for this list that reflects the issue you're describing?

Akendo: We have the problem, as you have observed before, that we're often not doing a proper review. This item on the list should be remaining on not only having a 'LGTM' or '+1', but sharing the knowledge behind this with someone else. So it could be:

- [ ] Shared the changes with the Team. 

So we want to prevent having a low bus factor.

fwilhe: thanks, added to the list

.github/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
+**What this PR does / why we need it**:
+
+**Which issue(s) this PR fixes**:
+Fixes #
+
+**Definition of Done:**
+- [ ] The code is sufficiently documented
+- [ ] Shared the changes with the Team so everyone is aware
+- [ ] The code is appropriately tested
+- [ ] If needed, the code is backported to the release branches of all currently maintained releases
```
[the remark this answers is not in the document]
fwilhe: yeah, agree the phrasing is not good. good catch.

5kt (approved):

PR #2446, opened 2024-11-04
Add dpkg-dev package to _dev feature

fwilhe:
dpkg-dev is needed for downloading the sources via `apt-get source` (at least for some packages), thus it should be available in dev images.


Akendo (approved):
LGTM

PR #2451, opened 2024-11-05
Update kubernetes repository GPG key

damyan:
The key was renewed after its expiry on 2024-11-02, so update it. S. https://github.com/kubernetes/release/issues/3818 for details

Note: the `khost` build fails currently without the change


fwilhe (approved):
thanks for taking care

PR #2456, opened 2024-11-06
feat: start-vm: add ability to create copy-on-write ephemeral disks

nkraetzschmar:

Adds an optional parameter qcow=size to the start-vm image arguments. Providing this parameter will cause start-vm to create a copy-on-write file as an overlay of the input image and use that as a disk instead.

This has two uses:

1. Allows for testing images without modifying the build artifacts
2. Allows for simpler resizing of disk images during tests

**Definition of Done:**

### DIFF
--- a/bin/start-vm
+++ b/bin/start-vm
@@ -180,7 +180,8 @@ source "${CURR_DIR}/.constants.sh" \
 --skipkp    <bool>        Skip the keypress to the verify status before execute.
 --vnc       <bool>        Sitches from serial console to vnc graphics console / enables vnc in --daemonize.
 --tpm2      <bool>        Enable TPM2 emulation and QEMU passthrough using swtpm.
-<image file>              A file containing the image to boot. Format is determined by the extension (raw, vdi, vpc, vhd, vhdx, vmdk, qcow2, iso)."
+<image file>              A file containing the image to boot. Format is determined by the extension (raw, vdi, vpc, vhd, vhdx, vmdk, qcow2, iso).
+                          An optional suffix of the form <image file>,qcow=<size> can be specified to write changes to a copy-on-write device instead"
 
 eval "$dgetopt"
 while true; do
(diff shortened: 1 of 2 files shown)


bin/start-vm
```diff
@@ -241,6 +241,7 @@ set_binaries () {
     bin_stat="stat"
     bin_sed="sed"
     bin_uuid="cat /proc/sys/kernel/random/uuid"
+    bin_realpath="grealpath"
```
fwilhe, line 244: the `g` prefix is macos/homebrew specific, on linux the binary is called `realpath`

bin/start-vm
```diff
@@ -255,6 +256,7 @@ set_binaries () {
         bin_stat="gstat"
         bin_sed="gsed"
         bin_uuid="/usr/bin/uuidgen"
+        bin_realpath="grealpath"
```
fwilhe, line 259: the `g` prefix is macos/homebrew specific, on linux the binary is called `realpath`

fwilhe:
Thanks for providing a fix @nkraetzschmar.

The syntax was a bit non-obvious to me, but it seems to work like this

```bash
./bin/start-vm .build/kvm_dev_usi-arm64-today-a9cc5c03.raw,qcow=3G
```

Seems to work fine on mac, please look into the grealpath issue and please amend the help text to make it easier to use.

bin/start-vm
nkraetzschmar: good catch, yeah should be realpath for linux and grealpath on macos, having grealpath in both places was a typo

fwilhe (approved):
LGTM

fwilhe (changes requested):
tiny thing: just noticed that we dont have qcow2 in the gitignore, leading to this git status

```
On branch feat/start-vm/qcow2
Your branch is up to date with 'origin/feat/start-vm/qcow2'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	bin/028f950e83f1.qcow2
```

Please add qcow2 to the gitignore file and I'll approve again

fwilhe:
btw, why is this qcow2 file in `bin`? I see this is the pwd of the `start-vm` script and also the efivars are in there, but it seems wrong to me. would it not make sense to have it in `.build`?

fwilhe (approved):

PR #2465, opened 2024-11-12
Reduce noise on serial console for _dev feature

fwilhe:
This reverses parts of #2167 because that makes working with the serial console too painful

### DIFF
--- a/features/kvm/file.include/etc/kernel/cmdline.d/10-console.cfg
+++ b/features/kvm/file.include/etc/kernel/cmdline.d/10-console.cfg
@@ -1,8 +1,13 @@
 serial_dev=ttyS0
+earlycon="earlycon=ttyS0"
 
+if [ "$(arch)" = x86_64 ]; then
+	earlycon="earlyprintk=$serial_dev,115200n8"
+fi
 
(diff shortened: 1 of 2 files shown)


features/kvm/file.include/etc/kernel/cmdline.d/10-console.cfg
```diff
@@ -1,13 +1,8 @@
...
-if [ "$(arch)" = x86_64 ]; then
-	earlycon="earlyprintk=$serial_dev,115200n8"
-fi
 
 if [ "$(arch)" = aarch64 ]; then
 	serial_dev=ttyAMA0
-	earlycon="earlycon=pl011,mmio,0x09000000"
 fi
 
-CMDLINE_LINUX="$earlycon console=tty0 console=$serial_dev $CMDLINE_LINUX"
```
nkraetzschmar, line 13: what's the advantage of removing earlycon?

nkraetzschmar:
removing the `systemd.journald.forward_to_console=1` is fine by me, but I'd keep the earlycon unless there's a good reason to remove it

nkraetzschmar (approved):

PR #2475, opened 2024-11-14
Bump cli, include container and change OCI repo

mxmxchere:
- #2466 is a prerequisite.
- KMS key should be changed before running on gardenlinux/gardenlinux


mxmxchere:
closed in favor of #2484

PR #2489, opened 2024-11-20
ci: release-page - fix platform test upload error code

yeoldegrove:

Currently an `exit 1` is thrown when a non existing platform test log file is uploaded in the release-page workflow.
This tries to fix it.

* https://github.com/gardenlinux/gardenlinux/actions/runs/11930076105/job/33250190914#step:6:121

**Definition of Done:**

* Could not test this. Test release will be done after merge.
* needs not backporting


fwilhe (approved):
I have no way to test it but it seems plausible, assuming it's actually ok that we have no logs in those cases

PR #2493, opened 2024-11-25
Firstboot mechanisms and rw sysroot

### LINKED ISSUE #2474: systemd firstboot preset mechanism seems broken
According to the documentation : 

```The systemd manager itself will initialize machine-id and preset all units, enabling or disabling them according to the systemd-preset settings.```

In certain situations, the preset mechanism is not working because the rootfs hasn't yet been remounted read/write.
This can also conflict with units being enabled/disabled via a feature's _exec.config_ ending up with an unwanted setup.

### PR DESCRIPTION
**Pull request by 5kt**

Fixes #2474

### CODE REVIEW

File: features/gardener/exec.config [hunk 1]
```diff
@@ -5,8 +5,12 @@ set -Eeuo pipefail
 chmod u-s /sbin/mount.nfs /sbin/mount.cifs
 
 # Disable containerd, Gardener will have to enable it
-systemctl disable containerd
-systemctl disable ssh
+# TODO: for the time being keep these here as it's more obvious, but they should be moved to file.include
+mkdir -p /etc/systemd/system-preset
+echo "disable containerd.service" > /etc/systemd/system-preset/00-containerd-disable.preset
```
nkraetzschmar, line 10: what's the point of using preset instead of directly disable?

nkraetzschmar (approved):

features/gardener/exec.config
5kt: In this case it's going to have the same effect, it's just to make it more readable.

PR #2503, opened 2024-11-27
hack: add script to list package releases

fwilhe:
This is useful for finding outdated pinned versions in our package repo.


nkraetzschmar (approved):
very nice :+1:

PR #2513, opened 2024-12-02
feat: usi: use bind mounts for /opt /root /home instead of symlinks

### LINKED ISSUE #2508: switch to using bind mounts in _usi
- use bind mounts instead of symlinks for `/opt`, `/home`, etc in `_usi` feature
- make the set of bind mounts runtime configurable
  - maybe have `/persistence` instead of `/var` and make the latter a bind mount as well

nkraetzschmar:

Using bind mounts instead of symlinks allows changing them dynamically without having to modify the immutable EROFS.
This will also allow making the set of persistent locations run-time configurable in the future.

With this the new mounts look like:

```
/efi /dev/vda1 vfat rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,utf8,errors=remount-ro
/var /dev/mapper/var ext4 rw,relatime,seclabel
/etc overlay overlay rw,relatime,seclabel,lowerdir=/etc,upperdir=/var/etc.overlay,workdir=/var/etc.overlay.workdir,uuid=on
/opt /dev/mapper/var[/opt] ext4 rw,relatime,seclabel
/root /dev/mapper/var[/root] ext4 rw,relatime,seclabel
/home /dev/mapper/var[/home] ext4 rw,relatime,seclabel
```

Fixes #2508 

**Definition of Done:**

### DIFF
--- a/features/_usi/file.exclude
+++ b/features/_usi/file.exclude
@@ -3,3 +3,6 @@
 /etc/systemd/system/ignition-disable.service
 /etc/repart.d/root.conf
 /usr/lib/tmpfiles.d/home.conf
+/opt/{,.}*
+/home/{,.}*
+/root/{,.}*
--- a/features/_usi/exec.config
+++ b/features/_usi/exec.config
@@ -8,6 +8,9 @@ mv /etc/systemd/system/default.target.wants /etc/systemd/system/post-reload.targ
 
 systemctl enable systemd-bless-boot.service
 systemctl enable var.mount
+systemctl enable opt.mount
+systemctl enable root.mount
+systemctl enable home.mount
 systemctl enable etc-overlay.service
 
 update-kernel-cmdline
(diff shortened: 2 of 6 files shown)


Akendo (approved):
This is a very firm and brief change. LGTM!

features/_usi/file.exclude
```diff
@@ -3,3 +3,6 @@
 /etc/systemd/system/ignition-disable.service
 /etc/repart.d/root.conf
 /usr/lib/tmpfiles.d/home.conf
+/opt/{,.}*
```
fwilhe, line 6: 2 questions:

what is the meaning of `{,.}*`? I see it is a regex but the comma dot is unclear to me. What is the intention?

What does excluding those files here do? What would happen if we did not do this?

nkraetzschmar: not a regex, just bash shorthand for `/opt/*` and `/opt/.*` since in bash `*` by default omits files starting with a dot

This excludes all files in `/opt`, `/home`, and `/root` but not the directories themselves. Since we are bind mounting over them there is no point in keeping any content inside them in the image

Issue #2526, opened 2024-12-06
Inconsistent entries in `/etc/passwd` file


Akendo:

Akendo:
My solution for now is to remove all but `www-data`.

Akendo:
Okay, here comes the solution to the problem at hand:

We need to ensure that we do not have any users presented on the system without a proper `/home` folder or missing shell. We test this with the usage of `pwck` that's a built-in tool from [`shadow-utils`](https://github.com/shadow-maint/shadow).
The tool complains as follows:
```
root@830652dcfff3:~# pwck
user 'games': directory '/usr/games' does not exist
user 'man': directory '/var/cache/man' does not exist
user 'lp': directory '/var/spool/lpd' does not exist
user 'news': directory '/var/spool/news' does not exist
user 'uucp': directory '/var/spool/uucp' does not exist
user 'www-data': directory '/var/www' does not exist
user 'list': directory '/var/list' does not exist
user 'irc': directory '/run/ircd' does not exist
user 'uuidd': directory '/run/uuidd' does not exist
pwck: no changes
```
The initial draft was to remove the given user from the password files. However, this breaks some assumptions Debian has of the base system. It's not possible to install any web server any longer, since without the `www-data` users presented, the web server has no system users, and Debian does not add it.

This is because Debian follows the [Linux Standard Base Core Specification 4.1](https://refspecs.linuxfoundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/book1.html) recommendation for [User & Group Names](https://refspecs.linuxfoundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/usernames.html). This implies that some of the optional users are present, as shown up there.

In short, removing unnecessary users from the system will break our system. Instead, we didn't find any requirement that these uses need a valid home directory. `pwck` utilizing the `/etc/login.defs` for [configuration](https://www.man7.org/linux/man-pages/man8/pwck.8.html#CONFIGURATION). By default, Debian configures the `NONEXISTENT` parameter for system users to the following home folder: `/nonexistent`. So we set for every user to expect `www-data` the default home folder to be `/nonexistent`

PR #2527, opened 2024-12-07
Feature/complianc mapping v1

Akendo:
We need to provide evidence that Garden Linux fulfills compliance requirements. In this PR, we introduce the first set of `security_marks` and it's corresponding tests that allow the identification of evidence that Garden Linux is up to standards. 

The first set of tests focus around User Management on a Linux system, and most of they are located in the base feature set. Only when we have users enabled for the cloud profiles, we alternate here.

Most of the tests passing with the exemption of the `passwd` check. By default, Debian adds a list of accounts to the system and makes some assumptions about them. This needs to be decided with the [#2526](https://github.com/gardenlinux/gardenlinux/issues/2526). 

Fixes: https://github.com/gardenlinux/security/issues/59


Akendo:
Will not close the PR here, but have an extra issue created for this: https://github.com/gardenlinux/security/issues/79

PR #2530, opened 2024-12-09
Remove assertion for capabilities on 'ping'

Akendo:
`iputils` was updated in debian testing with a newer version of ping that does not require setting `cap_net_raw`.
As the requirement of the package has changed, the test now does not test for the right thing. Now we want to test that the capability is not set.

For reference: [This is the related Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008281)

[this is the change log entry](https://metadata.ftp-master.debian.org/changelogs/main/i/iputils/testing_changelog)

```
* ping: Add Recommends on linux-sysctl-defaults and stop setting additional
     capabilities (Closes: #[1008281](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008281))
```

Also fixes a few typos that came up while working on it.


features/server/test/capabilities.d/capabilities.list
```diff
@@ -1 +1,2 @@
-/usr/bin/ping cap_net_raw=ep
\ No newline at end of file
+# Issue 2530
+#/usr/bin/ping cap_net_raw=ep
```
fwilhe, line 2: any specific reason to not delete the file instead? does it add value to keep this around in this form?

fwilhe:
For reference: [This is the related Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008281)

this is the changelog entry

```
* ping: Add Recommends on linux-sysctl-defaults and stop setting additional
     capabilities (Closes: #[1008281](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008281))
```

https://metadata.ftp-master.debian.org/changelogs/main/i/iputils/testing_changelog

fwilhe (changes requested):
We have multiple mentions of capabilities in ping. I'm not 100% sure if this relates to that, but I assume so. Please make sure the documentation is still accurate after your change is applied.

For reference, those are the findings:

```
bin/start-vm:164:    --help  "Starts a virtual machine with the most basic environment, needed. Perfect use for test cases and running samples of GardenLinux. This script can run unprivileged but it should have the possibility to use kvm (group kvm) otherwise it will be super slow. If not being root, the network will be slow and no ICMP (ping) capabilities are with the VM.
features/server/README.md:12:This features support unit testing and validates extended capabilities on `ping`, the installed packages, sshd configuration, systemd units etc.
features/vhost/README.md:11:To be fully complaint these unit tests validate the extended capabilities on `ping` and `gstreamer`, the installed packages, correctly defined suids and sgids as well as the systemd unit files.
features/chost/README.md:11:To be fully complaint these unit tests validate the extended capabilities on `ping` and `gstreamer`, the installed packages, correctly defined suids and sgids as well as the systemd unit files.
```

Also, commenting out is not appropriate in this case, imo. I if I understood it correctly, it's the new desired behaviour that ping does not need any capabilities, and as it is the only thing in the server feature that needed it, the feature should not have the test at all.

Please correct me if my understanding is wrong.

MrBatschner:
How about reversing the test (checking that ping no longer has the capability) instead of disabling it?

fwilhe:
> How about reversing the test (checking that ping no longer has the capability if it does not need it) instead of disabling it?

Agree, that would be the best approach. Thanks Thomas.

Akendo:
I looked into the code, and while I agree with @ThormaehlenFred approach, I lack the time to proper reengineering the underlying test to allow the negative test case properly. Instead, I just removed the given test. This will fix the nightly for now.

fwilhe:
Not sure about those findings, but I think the same that applies to `server` applies to them too.

Screenshot 2024-12-10 at 09 38 51

fwilhe:
So if I understand `tests/helper/tests/capabilities.py` correctly, you can test for a non-set capability by removing that line. An emtpy file in `features/$FEATURE/test/capabilities.d/capabilities.list` means 'no capabilities set', so leaving this file empty is better than deleting the test as far as I can tell.

fwilhe (approved):

features/_bfpxe/test/capabilities.d/capabilities.list
```diff
@@ -1,2 +1 @@
-/usr/bin/arping cap_net_raw=ep
-/usr/bin/ping cap_net_raw=ep
\ No newline at end of file
+
```
MrBatschner, line 1: If this file is empty anyway, we can just rm it (along with the `capabilities.d` directory). The test itself (`test_capabilities.py`) must remain however, as it also checks capabilities on `/usr`, `/var`, etc...

features/_iso/test/capabilities.d/capabilities.list
```diff
@@ -1,2 +0,0 @@
-/usr/bin/arping cap_net_raw=ep
```
MrBatschner, line 1: Same here...

features/_pxe/test/capabilities.d/capabilities.list
```diff
@@ -1,2 +0,0 @@
-/usr/bin/arping cap_net_raw=ep
```
MrBatschner, line 1: ... and here...

features/server/test/capabilities.d/capabilities.list
```diff
@@ -1 +0,0 @@
-/usr/bin/ping cap_net_raw=ep
```
MrBatschner, line 1: ... and here.

MrBatschner (changes requested):
see above

fwilhe:
> see inline

thanks, i was not sure how it would behave. updated.

MrBatschner (approved):
/lgtm

Issue #2525, opened 2024-12-05
Delete gl-oci manifest github artifacts after workflow run of upload_oci.yml


paulphys:
We should delete them as they're no longer needed after the workflow run and otherwise clog up the list of build artifacts

paulphys:
We should delete them as they're no longer needed after the workflow run and otherwise clog up the list of build artifacts

PR #2531, opened 2024-12-09
Delete temporary OCI manifest entry files after run

paulphys:
Addresses https://github.com/gardenlinux/gardenlinux/issues/2525

### DIFF
--- a/.github/workflows/upload_oci.yml
+++ b/.github/workflows/upload_oci.yml
@@ -121,3 +121,7 @@ jobs:
           do
             GLOCI_REGISTRY_TOKEN=${{ secrets.GITHUB_TOKEN }} GLOCI_REGISTRY_USERNAME=${{ github.repository_owner }} python /opt/glcli/src/glcli.py update-index --container ghcr.io/${{ github.repository }} --version ${{ inputs.version }} --manifest_file $manifest
           done
+      - name: Delete temporary OCI manifest entry files
+        uses: geekyeggo/delete-artifact@v5
+        with:
+          name: oci_manifest_entry_*


NotTheEvilOne (approved):
The code changes look good. You are deleting all artifacts with an defined prefix. A safeguard to not delete unexpected content would be to have another ID or build number to check but that would be a different PR anyway.

The overall handling could be further streamlined as well. Again this is topic of another PR.

paulphys:
Thanks for the fast review.

> The code changes look good. You are deleting all artifacts with an defined prefix. A safeguard to not delete unexpected content would be to have another ID or build number to check but that would be a different PR anyway.

How would it be able to delete other artifacts if I'm only selecting artifacts that begin with `oci_manifest_entry_`? Can you elaborate on that?

> The overall handling could be further streamlined as well. Again this is topic of another PR.

What improvements do you have in mind?

NotTheEvilOne:
> The only potential issue I see is that when two nightlys are running at the same time, one could delete the artifacts needed by the other (in the step between uploading the manifests and updating the index).

That would be one variant, the other variant an expected use of a name with prefix "oci_manifest_entry_". The action "actions/download-artifact" used earlier on workflow is scoped to the workflow run while the added script goes over all artifacts for a repository. As mentioned it might be more elegant to take an additional information like the exact file names into account as a future improvement.

> > The overall handling could be further streamlined as well. Again this is topic of another PR.
> 
> What improvements do you have in mind?

OCI uploads could happen directly in the action with help of `podman` negating the requirement to cache it. Further uploads could reuse the uploaded registry data. Another variant would be to take advantage of a custom `retention-period` in [actions/upload-artifact](https://github.com/actions/upload-artifact?tab=readme-ov-file#retention-period).

PR #2541, opened 2024-12-11
feat: add OCI artifact signing keys to cert build

nkraetzschmar:
Create an additional key to be used for signing OCI artifacts.

https://github.com/gardenlinux/sap-internal/issues/63


fwilhe (approved):
lgtm

Issue #2551, opened 2024-12-12
Enhance OCI Manifest with Add the mediaType


pnpavlov:
- Add the mediaType as described in spec https://github.com/opencontainers/image-spec/blob/main/image-index.md#image-index-property-descriptions:

> - mediaType string
> This property SHOULD be used and [remain compatible](https://github.com/opencontainers/image-spec/blob/main/media-types.md#compatibility-matrix) with earlier versions of this specification and with other similar external formats. When used, this field MUST contain the media type application/vnd.oci.image.index.v1+json. This field usage differs from the [descriptor](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#properties) use of mediaType.

The top level manifest should include

```
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [...]
}
```

paulphys:
Ready for review https://github.com/gardenlinux/python-gardenlinux-cli/pull/37

PR #2564, opened 2024-12-19
Bump glcli version

paulphys:
Closes https://github.com/gardenlinux/gardenlinux/issues/2551

The previous version of glcli was only able to apply the `mediaType` to new indexes, not update existing ones, which we need as [workflows/publish_container.yml](https://github.com/gardenlinux/gardenlinux/blob/main/.github/workflows/publish_container.yml) is actually writing to the OCI registry first before glcli uploads the rest.


NotTheEvilOne (approved):
LGTM

PR #2566, opened 2024-12-20
Ensure systemd detects drop-in files in `chronyd-restricted` for azure

### LINKED ISSUE #2562: Fix nightly 19th Dec 2024

### LINKED ISSUE #2565: fix nightly 2024-12-20
- nightly broken for feature `metal` due to changes in package `frr`
- new version sets `/etc/passwd` entry to: `frr:x:999:995::/nonexistent:/bin/sh`
  + this fails the `assert shell in ['/usr/sbin/nologin', '/bin/false']` check

NotTheEvilOne:
After another appearance of the error `Starting chronyd-restricted.service - NTP client (restricted)... Could not open /dev/ptp_hyperv : No such file or directory` it was found that `chronyd-restricted.service.d` was never applied because it is a symlink. systemd expect drop-in files to be placed in directories. Linking individual drop-in files is supported while analysis shows that symlinking directories does not (for systemd 257).

If directories are symlinked `systemctl cat chronyd-restricted` only outputs the original unit. Symlinking the drop-in file results in the expected output of the drop-in file as well.

Fixes #2562
Fixes #2565

### DIFF
--- a/features/azure/file.include/etc/systemd/system/chronyd-restricted.service.d
+++ b/features/azure/file.include/etc/systemd/system/chronyd-restricted.service.d
@@ -1,1 +1,0 @@
-chronyd.service.d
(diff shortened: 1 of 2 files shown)


fwilhe (approved):
failed CI seems unrelated to this pr

Issue #2567, opened 2024-12-23
Fix nightly #1852 (23rd Dec 2024)


pnpavlov:
Nightly is failing https://github.com/gardenlinux/gardenlinux/actions/runs/12451755097. It looks like the platform tests are executed in wrong container (1443?) and packages are not found.

Goal: Fix the nightly.

pnpavlov:
It appears that nightly tag was changed by [nightly #1850](https://github.com/gardenlinux/gardenlinux/actions/runs/12434796470) which was executed as part of releasing 1443 patch 16.
This breaks the nightly as the required packages are not available.

**Short term workaround:** @5kt tagged :nightly with the :1724.0 (last sucessfull main branch nighly job run [nightly #1836](https://github.com/gardenlinux/gardenlinux/actions/runs/12387301109))

**Long term solution:** https://github.com/gardenlinux/sap-internal/issues/69

PR #2570, opened 2024-12-26
Allow access to `ptp_hyperv` for `chronyd-restricted.service`

NotTheEvilOne:
This PR is an additional attempt to fix the issue `Starting chronyd-restricted.service - NTP client (restricted)... Could not open /dev/ptp_hyperv : No such file or directory`. Beside the fixes introduced in #2563 and #2566 the following setting is used in the default systemd unit: `PrivateDevices=yes`

`PrivateDevices=yes` means:
```
If true, sets up a new /dev/ mount for the executed processes and only adds API pseudo devices such as /dev/null, /dev/zero or /dev/random (as well as the pseudo TTY subsystem) to it, but no physical devices such as /dev/sda, system memory /dev/mem, system ports /dev/port and others. This is useful to turn off physical device access by the executed process.
```

Instead we would like to restrict access to pseudo devices and add `/dev/ptp_hyperv` to the devices allowed to be accessed by `chronyd-restricted.service`.

Fixes: #2562
Fixes: #2565
Fixes: #2567

### DIFF
--- a/features/azure/file.include/etc/systemd/system/chronyd-restricted.service.d/05-private-devices-access.conf
+++ b/features/azure/file.include/etc/systemd/system/chronyd-restricted.service.d/05-private-devices-access.conf
@@ -1,0 +1,4 @@
+[Service]
+DeviceAllow=/dev/ptp_hyperv
+DevicePolicy=closed
+PrivateDevices=no


features/azure/file.include/etc/systemd/system/chronyd-restricted.service.d/05-private-devices-access.conf
```diff
@@ -0,0 +1,4 @@
+[Service]
+DeviceAllow=/dev/ptp_hyperv
+DevicePolicy=closed
```
nkraetzschmar, line 3: setting this to `closed` might not even be necessary, as according to the docs[^1] `auto` (the default value) will behave the same, if any `DeviceAllow` value is present.

But this makes it a bit clearer, so probably good.

[^1]: https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#auto

nkraetzschmar (approved):
lgtm

PR #2575, opened 2024-12-31
Allow access for group `_chrony` to access symlinked `/dev/ptp_hyperv`

NotTheEvilOne:
This PR tries to fix `Starting chronyd-restricted.service - NTP client (restricted)... Could not open /dev/ptp_hyperv : Permission denied` by granting access to `/dev/ptp_hyperv` to the group `_chrony`. PTP devices usually are accessible by `root` only and use mode `0600`. As stated in the original `chronyd-restricted.service` file it `... cannot use reference clocks, HW timestamping, RTC tracking, and other features.`. As we are interested in allowing access to a device this requires changes to the udev rule as well.


nkraetzschmar (approved):
lgtm

nkraetzschmar:
maybe we should create a branch to test this on with all the permissions to do full nightly builds to speed up the dev cycle; so we don't need to always, PR, merge, test, repeat.

If this change still does not result in a green nightly I'll setup a branch with the necessary access for you tomorrow.

PR #2599, opened 2025-01-09
move core logic of _usi and related features into the initrd

### LINKED ISSUE #2591: move _usi logic into initrd
to support in-place updates on ironcore and cc we need to move the logic of `_usi` and related features directly into the initrd

nkraetzschmar:

Move all the repart / mount / etc overlay setup logic of `_usi` and related features into the initrd stage.

Resolves #2591


NotTheEvilOne (commented):
```
[aws-gardener_prod_trustedboot-amd64-today-local.esp.tar] Couldn't find EFI system partition. It is recommended to mount it to /boot or /efi.
[aws-gardener_prod_trustedboot-amd64-today-local.esp.tar] Alternatively, use --esp-path= to specify path to mount point.
[aws-gardener_prod_trustedboot-amd64-today-local.esp.tar] Wrote unsigned /tmp/tmp.gW0xs7h4ew
```
As the resulting `*.esp.tar` contains the expected `EFI` directory it may be possible to prevent this warning?

NotTheEvilOne (commented):
I like the integration of another possibility to configure `initrds` dynamically if needed. I'm not sure if the implementation in a separate feature flag would be more suitable for re-usability.

`_usi` is used by `_trustedboot` (only) as far as I know. Would it maybe make more sense to move the TPM related stuff to `_trustedboot` to keep `_usi` reasonable small and readable?

nkraetzschmar:
`_usi` can also be used without `_trustedboot`, in that case it wil auto pull in `_nocrypt`. This mode of operation is needed to support in-place updates for gardener on nodes without TPM support.

NotTheEvilOne:
> `_usi` can also be used without `_trustedboot`, in that case it wil auto pull in `_nocrypt`. This mode of operation is needed to support in-place updates for gardener on nodes without TPM support.

I meant to move the TPM related stuff to `_trustedboot` as you mentioned that `_nocrypt` doesn't use it anyway :)

5kt (approved):

PR #2617, opened 2025-01-22
Fix GCP platform-test resource cleanup

yeoldegrove:

- Currently GCP subnets are not deleted before the actual network is deleted.
- Add option `A` to interactive mode to delete ALL resources that are found.


NotTheEvilOne (approved):
LGTM (actually not code-wise as the `google-cloud-compute` SDK is not really useful for linked resources like `Subnetwork` but it does it's job)

PR #2628, opened 2025-01-30
Split `_onmetal` and `_oci` feature flags

NotTheEvilOne:
This PR suggests renaming the existing feature flag `_oci` to `_onmetal` and creating a new `_oci` feature flag that creates OCI archives.

This can be reused for `gardenlinux-ccloud` to create OCI archives for non-container use like orabos.


sramapu:
no activity for more than 500d

PR #2630, opened 2025-01-31
Add utility script to find the correct build container for Garden Linux releases

fwilhe:
This script might make it a bit more easy to maintain backports of packages, when needed.

### DIFF
--- a/bin/find-build-container-for
+++ b/bin/find-build-container-for
@@ -1,0 +1,33 @@
+#!/usr/bin/env python3
+
+import requests
+import sys
+
+def download_file(version):
+    # Ensure the version follows the format "1234.5"
+    if '.' not in version:
+        version = f"{version}.0"
+
+    url = f"https://raw.githubusercontent.com/gardenlinux/repo/refs/tags/{version}/.container"
+
+    try:
+        response = requests.get(url)
+        
+        if response.status_code == 200:
+            print(response.text)
+        else:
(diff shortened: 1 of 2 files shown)


Vincinator:
Maybe this functionality fits good in https://github.com/gardenlinux/glrd?

bin/find-build-container-for
NotTheEvilOne, line 1: LGTM

In general I wonder if it would help to reuse existing dependencies e.g. to `click` to use advanced print capabilities like colored output (see https://click.palletsprojects.com/en/stable/utils/#ansi-colors) or args parsing. In general for this specific use case we only want the URL, so there is no really need for it, but it would improve our code base in reusing what we have and maybe even add some library if needed later on.

NotTheEvilOne (approved):

bin/find-build-container-for
fwilhe: I try to limit external dependencies in python scripts because I'm not a fan of the tooling, pip, virtualenv etc.. And I think in this case requests is the only real dependency.

fwilhe:
@Vincinator 

> Maybe this functionality fits good in https://github.com/gardenlinux/glrd?

I don't know, I think its a different use case, this is about maintainer tools, similar to what @gehoern proposed for the 'need backport?' tool. I'm open to putting this somewhere else.

nkraetzschmar (approved):
lgtm

PR #2656, opened 2025-02-05
Feature/add read file remote func

Akendo:

In our tests, we often read the content of a file from the remote file system. This PR adds this as a more generic function to the code.


tests/helper/utils.py
```diff
@@ -5,6 +5,21 @@
...
+    """
+    status, output = execute_remote_command(client, f"cat {file}", skip_error=True)
+    assert status == 0, f"Error reading {file}" 
+    to_return = output.split("\n")
+    # Check if we might have an emtory entry.
+    if "" in to_return:
+        to_return.remove("")
+    if remove_comments: 
+        return [line for line in to_return if '#' not in line]
+    return to_return
```
moritzkirchnersap, line 20:
```suggestion
   return [line for line in output.split("\n") if line and if not line.startswith("#")]
```
Your version would also remove lines like:

`woah this is code # and this is an inline comment!`
 
Also I would love to see what this function does for the file "/dev/urandom" or "/dev/zero" :D

Akendo: Good catch. I'm going to check your solution.

tests/helper/utils.py
```diff
@@ -5,6 +5,22 @@
...
+    status, output = execute_remote_command(client, f"cat {file}", skip_error=True)
+    #assert status == 0, f"Error reading {file}"
+    to_return = output.split("\n")
+    ## Check if we might have an emtory entry.
+    if "" in to_return:
+        to_return.remove("")
+    if remove_comments:
+        # Ensure that we only skip lines when the comment start with.
+        return [line for line in output.split("\n") if line and not line.startswith("#")]
+    return to_return
```
moritzkirchnersap, line 21:
```suggestion
    #assert status == 0, f"Error reading {file}"
    to_return = []
    For line in output.split("\n"):
        ## Check if we might have an emtory entry.
        if not line:
            continue
        if remove_comments and line.startswith('#'):
            continue
       to_return.append(line) 
       
    return to_return
```

tests/helper/utils.py
```diff
@@ -5,6 +5,22 @@
 import subprocess
 
 
+def read_file_remote(client, file, remove_comments=False)-> list:
+    """ Read a file from the remote file system and returns it for
+        proper processing trunecated.
+    """
+    status, output = execute_remote_command(client, f"cat {file}", skip_error=True)
```
NotTheEvilOne, line 12: I would recommend to use `client.download_file()` here instead:
- We already implemented "reading remote files"
- It's more file type agnostic
- It doesn't copy memory in splitting and recombining its return value

The method named `read_file_remote()` should than read the file from e.g. a temporary location instead. `download_file()` may need to be extended to support defining target locations as the underlying `SCP.get()` does support that functionality.

Wrap it in a `with` context to delete the temporary file after download an reading if required.

tests/helper/utils.py
```diff
@@ -5,6 +5,22 @@
...
+    #assert status == 0, f"Error reading {file}"
```
NotTheEvilOne, line 13: If functionality of removing comments or empty files is needed this could be a separate function.

NotTheEvilOne (changes requested):

tests/helper/utils.py
Akendo: > We already implemented "reading remote files"

I'm not aware of any function that does this. Can you point me to it via a link? 

> I would recommend to use client.download_file() here instead:

Since we look at password databases at times, I consider it a bad idea to download the password db from the other system down to the local one, even for testing. Likewise, I operate sometime on the proc file system, something that you can't download.

NotTheEvilOne: > Since we look at password databases at times, I consider it a bad idea to download the password db from the other system down to the local one, even for testing. Likewise, I operate sometime on the proc file system, something that you can't download.

I can only partly agree with you here. If your use case is to "look at password databases" it's a minor difference if you save it as temporary file to (hopefully) a memory file system or leave the data in plain text in memory anyway. Both are more or less a security issue if not used for testing only. Best way would be to run the test required locally on the machine containing the sensitive data.

If proc file system should be read SCP won't work.

Still I would appreciate to split the functionality to read data from remote and do any kind of parsing on it. Is it really useful to return a `list` of lines?

NotTheEvilOne: > > We already implemented "reading remote files"
> 
> I'm not aware of any function that does this. Can you point me to it via a link?

`file_content()` in [tests/helpers/tests/file_content.py](https://github.com/gardenlinux/gardenlinux/blob/cdfcd2c7943be87acc9eb935aad3f08bc494da39/tests/helper/tests/file_content.py#L4). IMHO this function should be splitted as it does too much.

Akendo: I took the time to read over the file and tend to agree that the function does a lot. It's also designed to be more a test than a standalone function. The `_get_content_remote_file` from it does, from the looks of it, do quite the same as the function I implemented here. Hence, it might worth refactor here the code as well

tests/helper/utils.py
```diff
@@ -5,6 +5,25 @@
...
 
+def read_file_remote(client, file, remove_comments=False)-> list:
+    """ Read a file from the remote file system and returns it for
+        proper processing trunecated.
+    """
+    status, output = execute_remote_command(client, f"cat {file}", skip_error=True)
+    assert status == 0, f"Error reading {file}"
+    to_return = []
+    for line in output.split("\n"):
+        ## Check if we might have an empty entry.
```
Vincinator, line 16: Silently removes empty lines, this could be misleading for developer writing tests using this helper.

Suggest to make this optional as with `remove_comments`

Akendo: Good point. I've updated my PR to consider this as well.

Vincinator (approved):
Thanks! Looks good to go. 
@NotTheEvilOne your feedback is basis for a bigger refactor task.

Issue #2661, opened 2025-02-07
Fix Nightly #1979 2025-02-07 (tbd)


pnpavlov:
Nightly failed [#1979](https://github.com/gardenlinux/gardenlinux/actions/runs/13194161702)

nkraetzschmar:
Looks like two issues at once:

1. Booting `_trustedboot` is flaky in the platform tests, root cause yet unknown, can't reproduce locally
2. On GCP `systemd-analyze` reports the kernel boot time as over 30s despite logs showing a time of 23s till init is started

For 2 we can temporarily bump the allowed kernel boot time.
1 still requires further investigation

nkraetzschmar:
Root cause for the GCP error: the kernel boot time in the test is actually the kernel + initrd time. since `_usi` is now doing a significant amount of first boot logic in the initrd it is clear that this takes longer now.

nkraetzschmar:
Update on the flaky issue with `_trustedboot` on aws, managed to capture the following serial output during the test:

```
No bootable device found.
Rebooting in 1 minute.
```

nkraetzschmar:
Root cause identified:

`Requires` dependecies on the mount units cause them to get killed when udev is stopped

PR #2665, opened 2025-02-07
fix: bump kernel time to work with _usi

nkraetzschmar:
The `tolerated_kernel_time` in this test is actually the sum of kernel time + initrd time. Since #2591 the initrd is doing a lot more work on first boot, so this needs to be bumped up

#2661


Vincinator (approved):

Vincinator:
sounds reasonable

PR #2658, opened 2025-02-06
Feature/clean up uncessary user accounts

Akendo:
**What this PR does / why we need it**:

During our process integrating the hardening rules, we noticed that we have home directories with the `passwd` that are not existing with the file system. We clean up this with this PR.

**Which issue(s) this PR fixes**:
Fixes #2526

PR #2668, opened 2025-02-07
Feature/test shadow permission

Akendo:

We need to ensure that the shadow files are well protected and have the right permissions to them. With this PR, we have tested that all files have set to the correct file permissions.


NotTheEvilOne (changes requested):
I would appreciate that all changes and comments related to `password_shadow` would be proposed in one PR.

As mentioned in #2669 I would recommend to move these tests to the feature test file as it is not used in multiple places. Therefore it does not qualify for a library function.

As mentioned in #2658 it would increase readability to split the tests into multiple test cases instead of encapsulating all checks in one test function.

PR #2669, opened 2025-02-07
Feature/test password hashes

Akendo:

Based on CIS [5.3.4](https://www.tenable.com/audits/items/CIS_Distribution_Independent_Linux_Workstation_L1_v2.0.0.audit:2ecda5602069de22f53cf45659f91103) Benchmark, we need to ensure that we have a strong password hash present. Since we deviate from this assumption, we need to provide some additional documentation regarding this deviation. This PR adds some doc and the `security_mark`.


features/base/test/test_password_hashes.py
```diff
@@ -1 +1,10 @@
...
+from helper.tests.password_hashes import password_hashes
+import pytest
+
+
+@pytest.mark.security_id(325)
+def test_password_hashes(client):
+    """
+    Ensure that we have strong hash function enabled.
+    """
+    password_hashes(client)
```
NotTheEvilOne, line 10: Please be aware of the changes proposed at #2658 and feedback given there. Furthermore the underlying "password_hashes()" is used exactly one time. I'm not sure if it would be better readable to integrate the intended changes into then feature test file of `base`.

NotTheEvilOne (changes requested):

features/base/test/test_password_hashes.py
Akendo: I provided feedback about this in the other PR as well, and I think that the current change here the way to go. Since otherwise the tests won't be properly marked.

NotTheEvilOne (approved):
I'll approve this PR only for correct labeling, not based on quality of the suggested code change.

features/base/test/test_password_hashes.py
Akendo: Also, I've created an issue for this: https://github.com/gardenlinux/gardenlinux/issues/2781

PR #2672, opened 2025-02-10
Upgrade python to 3.13 in platform tests

fwilhe:
Python 3.13 is the current version in debian trixie and the maintained version of python in Garden Linux today.


NotTheEvilOne (approved):
LGTM

Vincinator (approved):
Would it make sense to add in the commit message when merging a reason for upgrading to 3.13? WDYT? 

Nothing fancy, just something along the lines of: 
- To match current debian testing
- Fix tests that depend on 3.13
- other 

LGTM

PR #2685, opened 2025-02-11
Fix/test mount trustedboot

Akendo:
Our tests do not take the trustedboot feature into account, in this PR we update the test for mount to respect that.


moritzkirchnersap (commented):
I would have expected to see something like this: 

https://docs.pytest.org/en/7.1.x/example/markers.html

and then:

pytest -m "trustedboot" ...
pytest -m "not trustedboot" ...
pytest.mark.skipif(something something)

tests/conftest.py
```diff
@@ -418,6 +418,26 @@ def ccee(platform):
     if platform != 'openstack-ccee' and platform != 'openstack-baremetal-ccee':
         pytest.skip(f"test only supported on ccee")
 
+@pytest.fixture
+def ephemeral(platform):
+    if platform != 'ephemeral':
+        pytest.skip(f"test only supported on ephemeral")
+
+@pytest.fixture
+def non_ephemeral(platform):
```
nkraetzschmar, line 427: Instead of filtering on `_trustedboot` and `_ephemeral` we should filter on `_usi` as that's the core feature and all the others are just specifications on which optional security measures are enabled.

nkraetzschmar (dismissed):

tests/conftest.py
Akendo: I've removed this for now. Since this PR fixed the mount test. I'll add this later as well.

moritzkirchnersap (approved):

PR #2691, opened 2025-02-11
Always upload test artifacts

moritzkirchnersap:
Right now if some of the tests have the result "failure" the job fails (which is ok) but the artifacts as in the pytest test log is not being uploaded. For the compliance reporting it would be good to keep those artifacts to then do reporting on them.

This PR is supposed to upload the test result log files independently of the test results themselves.

Fixes #

**Definition of Done:**

**Release note**:

```feature user

```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -139,7 +139,7 @@ jobs:
     - uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # pin@v4.4.3
       with:
         name: platform-test-${{ env.cname }}
-        if: {{ success() || failure() }} 
+        if: success() || failure()
         path: |
           ${{ env.artifact_dir }}/${{ env.cname }}.platform.test.log
           ${{ env.artifact_dir }}/${{ env.cname }}.platform.test.xml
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -139,6 +139,7 @@ jobs:
     - uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # pin@v4.4.3
       with:
         name: platform-test-${{ env.cname }}
+        if: success() || failure()
```
Akendo, line 142: This `if` seems a bit unnecessary now. Can we not remove it?

NotTheEvilOne: In fact the `if` statement should remain but the value may change to `always()`. See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#context-availability

Please be aware that this means that the workflow is triggered under any circumstances and artifacts might not be there to be uploaded.

moritzkirchnersap: Thanks for the feedback! The idea was not to upload the artifacts in case the job is manually cancelled, hence the check for success() || failure(). And yes: it might be that the artifacts might not be produced, but I would assume that then the tests producing the artifacts also failed in a separate step, so I would also assume that the resulting state of the workflow run wouldn't be worse.

WDYT?

PR #2695, opened 2025-02-11
Feature/configuration for tests uci

Akendo:
As discussed, this commit adds a single flavor for the trustedbooted and ephemera named usi. Also, we disable the test for mount, since our rework wasn't working as intended.


moritzkirchnersap (approved):
Better than failing tests but 😐

PR #2697, opened 2025-02-11
Add generic support to retry failing tests

NotTheEvilOne:
This PR adds support to add a wrapper around tests to retry them for a specified amount of times if a given list of exceptions occurs. This is used for `base/test/test_time_config.py::test_ntp` and may be added for other tests as required.

### DIFF
--- a/features/base/test/test_time_config.py
+++ b/features/base/test/test_time_config.py
@@ -35,6 +35,7 @@ def test_clock(client):
     ), "clock skew should be less than 5 seconds. Local time is %s and remote time is %s" % (local_seconds, remote_seconds)
 
 
+@pytest.mark.flaky(reruns=3, reruns_delay=10, only_rerun="AssertionError")
 def test_ntp(client, non_azure, non_chroot):
     """ Azure does not use systemd-timesyncd """
     (exit_code, output, error) = client.execute_command("timedatectl show")
--- a/tests/Pipfile
+++ b/tests/Pipfile
@@ -14,6 +14,7 @@ scp = "*"
 dacite = "*"
 cffi = "*"
 pytest-json-report = "*"
+pytest-rerunfailures = "*"
 
 [dev-packages]
 pytest = "*"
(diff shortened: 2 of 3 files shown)


tests/helper/wrappers.py
```diff
@@ -0,0 +1,43 @@
+from asyncio import sleep as asyncio_sleep
+from functools import wraps
+from time import sleep
+
+async def async_retry(py_callable, retries = 3, wait_time = 3, exceptions_retried = ( AssertionError, )):
```
yeoldegrove, line 5: To make it even more robust initially, we could wait for 10s before retrying again. What do you think?
Otherwise, I like this feature and that it can be chosen on a per test basis.

yeoldegrove (changes requested):

tests/helper/wrappers.py
moritzkirchnersap: While I like the functionality I'm wondering if there is a pytest package with similar functionality. If there are packages then I'm always a proponent of using them because my naive assumption is they work.

What do we think about pytest-rerunfailures https://pypi.org/project/pytest-rerunfailures/#re-run-all-failures-matching-certain-expressions

which should allow something like:
`@pytest.mark.flaky(rerun_except="AssertionError", reruns=5, reruns_delay=2)`

NotTheEvilOne: We can increase the default value to e.g. 5 seconds, 10 might be a bit high. You can always customize it if needed per test basis @yeoldegrove.

`@pytest.mark.flaky` seems to contain all features implemented here as well, you are right @moritzkirchnersap :)

yeoldegrove (approved):
Thanks for this feature. Approved.

Few minor things:
- Does it make sense to document somewhere that we can re-run tests this way? I am not sure what we have documented around pytest, yet ...
- The version bump is because we use a newer python version since a few nighties, right? Did you test that this works?

NotTheEvilOne:
[...]
>     * The version bump is because we use a newer python version since a few nighties, right? Did you test that this works?

Yes, as I touched Pipfile anyway I updated Python as well. Most tests are done locally. As we have it in `platform-test-image`s I assume that no corner cases hide there.

PR #2543, opened 2024-12-11
cloud resource provisioning via OpenTofu

yeoldegrove:
**What this PR does / why we need it**:

In https://github.com/gardenlinux/gardenlinux/issues/2335 the proposal a new approach for cloud resource provisioning for the platform tests is requested and outlined.

Please have a look at the [README](https://github.com/gardenlinux/gardenlinux/blob/adad10b24d5683068cc95d68d9941b8704b3b0fd/tests/platformSetup/tofu/README.md) for an introduction.

This PR adds:
- [add ECDSAKey and Ed25519Key support to tests/helper/sshclient.py](https://github.com/gardenlinux/gardenlinux/commit/c6617976593379077c3dd0c147046f18d0c93914)
  - Newer key types are good from a security best practice.
  - Enables developers to use their existing ECDSA and ED25519 ssh keys from their workstation.
- [enable "manual" pytest runs to define a platform](https://github.com/gardenlinux/gardenlinux/commit/ea84efe150d530343db77bbdacae17a11e5c5be9)
  - adds a new fixture/setting `platform` to the manual pytest configs and maps the old `iaas` flag to it (if not provided) to be backward compatible. Every occurance of `iaas` to skip a test is replaced by the new `platform` setting.
- [use openTofu for platform-test provisioning](https://github.com/gardenlinux/gardenlinux/commit/0693d30763f857a35a8035b5306ed1ed54f2fb41)
  - adds all the needed OpenTofu code and modules to deploy the platforms and flavors we need in gardenlinux
  - see [`tests/README.md`](https://github.com/gardenlinux/gardenlinux/blob/03e62b955f8ae342adf5348c438a8a28bb0dca40/tests/README.md#public-cloud-platforms) on how to use the new OpenTofu code for public cloud platforms.
- [use Makefile to build images and use flavours.yaml](https://github.com/gardenlinux/gardenlinux/commit/d48734404c6df613aad10528aa2872944a699b8c)
  - every step of the build/test/test-platform cycle is now callable via a make target
  - build targets are named after the flavors from `flavors.yaml` and are enabled dynamically if the `build`, `test´ or `test-platform` keys are set to `true` in a flavor.
    - `make` --> lists `build` and `test` (chroot) targets
    - `make --directory tests/platformSetup` lists (OpenTofu) targets to plan/apply/show/login/destroy `platform-test` resources
    - `make --directory tests` lists `platform-tests` targets

**Which issue(s) this PR fixes**:
Fixes #2335 

**Definition of Done:**
[...]

PR #2704, opened 2025-02-12
platform-test logs are not named with a cname but a combination of $FLAVOR-$ARCH

yeoldegrove:

platform-test logs are not named with a cname but a combination of $FLAVOR-$ARCH
This PR fixes this in the GithubActions.

This is a change introduced with https://github.com/gardenlinux/gardenlinux/pull/2543

https://github.com/gardenlinux/gardenlinux/blob/e033a4ea8aa4df50123e77960df32f7b049ba418/tests/Makefile#L56C215-L56C216

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -201,7 +201,7 @@ jobs:
       run: |
         make --directory=tests ${FLAVOR}-${ARCH}-test-platform 2>&1 | tee "${{ env.artifact_dir}}/${{ env.cname }}.platform.test.log"
 
-    - name: copy platform test log for ${{ matrix.flavor }} on ${{ matrix.arch }}
+    - name: copy platform test junit xml file for ${{ matrix.flavor }} on ${{ matrix.arch }}
       if: always()
       run: |
         cp "tests/${FLAVOR}-${ARCH}.platform.test.xml" "${{ env.artifact_dir }}/${{ env.cname }}.platform.test.xml" || true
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -201,6 +201,11 @@ jobs:
       run: |
         make --directory=tests ${FLAVOR}-${ARCH}-test-platform 2>&1 | tee "${{ env.artifact_dir}}/${{ env.cname }}.platform.test.log"
 
+    - name: copy platform test log for ${{ matrix.flavor }} on ${{ matrix.arch }}
```
moritzkirchnersap, line 204:
```suggestion
    - name: copy platform test junit xml file for ${{ matrix.flavor }} on ${{ matrix.arch }}
```

.github/workflows/tests.yml
```diff
@@ -201,6 +201,11 @@ jobs:
...
+      if: always()
+      run: |
+        cp "tests/${FLAVOR}-${ARCH}.platform.test.xml" "${{ env.artifact_dir }}/${{ env.cname }}.platform.test.xml" || true
```
moritzkirchnersap, line 207: This assumes that we always have a xml file to copy, right? While the || true wouldn't cause an error in case the xml file doesn't exist in this step, in case the file doesn't exist but the run should continue then the next step would fail as it does right now, correct?
Does such case exist? Or it is hypothetical?

yeoldegrove: https://github.com/gardenlinux/gardenlinux/blob/e033a4ea8aa4df50123e77960df32f7b049ba418/tests/Makefile#L56C215-L56C216 writes the file as soon as the make target reaches a certain stage. If anything fails before that, there might not be a file... This is why i added `if: always()`, in case the tests fail and/or there is no log file.  `|| true` is added as if there is no log file I want this step to not cause a failure of the workflow (even though this is the last step and might not actually cause any harm).

moritzkirchnersap (approved):

PR #2706, opened 2025-02-12
Feature/cleanup conftest

Akendo:
During my testing on the last day, I noticed some misconfigurations happening within the `conftest.py` file. For instance, functions to detect for `firecracker` or `local` were added more than once and with out-dated configuration settings.  Also, adding more entries was error prune, since details were earlier lost. In this PR, we split the configuration into smaller sized configuration files to allow better maintaining the code. 

The split is implemented by treating all the configuration settings as [`pytest_plugins`](https://docs.pytest.org/en/latest/how-to/fixtures.html#using-fixtures-from-other-projects) variable. Please note: This will only work since we're using a top-level `conftest.py`, we can't use this in any [lower level](https://docs.pytest.org/en/7.1.x/deprecations.html#pytest-plugins-in-non-top-level-conftest-files) of it.


tests/conftests/elements.py
```diff
@@ -0,0 +1,18 @@
+"""
+Features Elements.
```
Vincinator, line 2: spelling

tests/conftests/features.py
```diff
@@ -0,0 +1,32 @@
+"""
+Features Flags.
```
Vincinator, line 2: Feature
```suggestion
Feature Flags.  
```

tests/conftests/features.py
```diff
@@ -0,0 +1,32 @@
+"""
+Features Flags.  
+Please note: Feature flags have always a _ present.  When missed, you wont't able
+to find it.
```
Vincinator, line 4:
```suggestion
Please note: Feature flags must be prefixed with a "_", for example "_prod". 
```

tests/conftests/platforms.py
```diff
@@ -0,0 +1,166 @@
+"""
+In this we keep the platforms specify test configuration.
+The file has been strucutred as fllows:
```
Vincinator, line 3:
```suggestion
The file is structured as follows:
```

tests/conftests/platforms.py
```diff
@@ -0,0 +1,166 @@
+"""
+In this we keep the platforms specify test configuration.
```
Vincinator, line 2:
```suggestion
In this we keep the platform specific test configuration.
```

tests/conftests/platforms.py
```diff
@@ -0,0 +1,166 @@
...
+The file has been strucutred as fllows:
+    - platform
+    - non_platform
+    - pool platform
+    - pool non_platform
```
Vincinator, line 7: only nit pick: 
I did have to look up in the code what you mean with the term "pool". 

Idea is good, I'm fine with the term now, but maybe `platform set`  or `platform exclusion set` is more self explanatory

Vincinator (commented):
Makes sense to me. 
added some minor comments/suggestions.

tests/conftests/platforms.py
```diff
@@ -0,0 +1,166 @@
...
+    if platform != 'chroot':
+        pytest.skip('test only supported on chroot platform')
+
+@pytest.fixture
+def container(platform):
+    """
+    This fixture is an alias of "chroot" but does not use the "chroot" env.  However, it only needs
+    the underlying container for its tests.
+    """
+    if platform != 'container' or platform != 'chroot':
```
Vincinator, line 42: We have no **Test** Platform called "container".

`platform` here is referring to the test platform (see available test setups here [tests/platformSetup](https://github.com/gardenlinux/gardenlinux/tree/3347229d6e09cf786a39fed0d60b32319e92f8ba/tests/platformSetup))
e.g.

- chroot - chroot into the Image, start ssh server and execute tests via ssh 
- kvm - start kvm virtual machine, start ssh server and execute tests via ssh 
- some clouds ☁️  - spawn cloud infra via opentofu and execute tests
- manual  - attach via ssh to some instance you spawned yourself, could be anything that answers to your ssh (e.g. a container 😄)
- openstackccee  

more info about the test platforms can be found [here](https://github.com/gardenlinux/gardenlinux/tree/3347229d6e09cf786a39fed0d60b32319e92f8ba/tests#using-the-tests-on-supported-platforms)

Vincinator: We do have a Garden Linux Feature [container](https://github.com/gardenlinux/gardenlinux/tree/main/features/container), which has nothing to do with the test platform

tests/conftests/platforms.py
```diff
@@ -0,0 +1,166 @@
...
+    if platform == 'azure':
+        pytest.skip('test not supported on azure')
+
+@pytest.fixture
+def non_chroot(platform):
+    if platform == 'chroot':
+        pytest.skip('test not supported on chroot')
+
+@pytest.fixture
+def non_container(testconfig):
```
Vincinator, line 100: In contrast, this is in fact referring to images having the feature [container](https://github.com/gardenlinux/gardenlinux/tree/main/features/container) enabled. 

This is not the opposite of what the current implementation of the `container` fixture above does.

Vincinator (changes requested):
As discussed, here are some comments based on our discussion.


In my opinion it should behave like this

- container fixture => allows tests on flavours that have the container feature enabled 
- non_container fixture => allows tests on flavours that have NOT the container feature enabled 

While the non_container already does what I would expect, the current container fixture is misleading. 

This was not introduce by your PR, but is present already for some years.

tests/conftests/platforms.py
Akendo: Thanks, I'm going to re-work this. I'm thinking about what we can call it `container_image` instead of `container`? I guess we could also call it `linux_container`. This should reduce changes of misunderstandings. Also, I think we shouldn't have the ambiguity with the term container.

Akendo: > We do have a Garden Linux Feature [container](https://github.com/gardenlinux/gardenlinux/tree/main/features/container), which has nothing to do with the test platform. 

Technically, it's considered a [platform](https://github.com/gardenlinux/gardenlinux/blob/main/features/container/info.yaml#L2). 

Maybe we have to reconsider the term platform as well?

Vincinator: > Technically, it's considered a [platform](https://github.com/gardenlinux/gardenlinux/blob/main/features/container/info.yaml#L2).

 yes, but not **test** platform

yeoldegrove:
@Akendo As just discussed, maybe wait before renaming `kvm->qemu` for the restructuring that I am currently doing.
I will link my PR here as soon as it is done.

Akendo:
@gehoern and @yeoldegrove agree that we rename the `iaas` keyword to `provisioner`. However, I see that this would be much more work to do. Instead, I'm going to create an issue for this task. I started with some small changes, but we will finish this later.

Akendo:
issue created: https://github.com/gardenlinux/gardenlinux/issues/2783

Vincinator (approved):

tests/conftests/provisioner.py
```diff
@@ -0,0 +1,57 @@
+"""We define the fixtures here, that we can use to run the tests.
+
+Do not confuse it with the platform, since we're talking here more about
+Container Runtime Environment like podman/docker or Virtualization Environment
+we want to conduct our tests on.
```
Vincinator, line 5: Do we have a container runtime environment that we use to run our test on? 

`pytest --iaas=X` where X is one of the following:

```
chroot
kvm 
manual
openstack-ccee
aws
ali
gcp
azure 
local
firecracker
```
(based on [tests/README.md](https://github.com/gardenlinux/gardenlinux/tree/main/tests#platform-tests) and [tests/conftest.py:testconfig](https://github.com/gardenlinux/gardenlinux/blob/f171b52b0051006b8628e084141c7f513dbbcbff/tests/conftest.py#L102))

tests/conftests/platforms.py
```diff
@@ -0,0 +1,146 @@
...
+
+@pytest.fixture
+def azure(platform):
+    if platform != 'azure':
+        pytest.skip('test only supported on azure platform')
+
+@pytest.fixture
+def container(platform):
+    """
+    This fixture is for the container platform and require a Container Runtime Environment.
```
Vincinator, line 34: previous definition of this fixture did something else:

```
def container(iaas):
     if iaas == 'chroot' or iaas == 'manual':
         pytest.skip('test only supported on containers')
```

it made sure that a test is not executed with `pytest --iaas=chroot` or `pytest --iaas=manual`

Are we sure we want this change?

Vincinator (approved):
works for me, we will revisit this anyway.
just added some comments as food for thought. 

dont want to block

tests/conftests/platforms.py
Akendo: I think the underlying assumes have changed. Before, we did only perform tests within the environments.

Our original code is this one:
```
@pytest.fixture
def container(iaas):
    if iaas != 'chroot':
        pytest.skip('test only supported on containers')
```
It was introduced by this [commit][0] and updated three weeks ago by this [commit from @yeoldegrove ][1]. I think this emphasizes the problem we have already, that things mix up. The initial code didn't different between `chroot` and `container` pre-Se. And that makes sense, since at the time of writing we didn't have the container features. 

This assumption is incorrect now because we have introduced [containers][2], a year later. But from what I get out of the code, the tests were never refactored in our code-base to take this into account. Instead, in the PR, the necessary tests were simply disabled by creating `TEST-NAME.disable` files within the container features testes folder. 

However, we might need to check with @yeoldegrove about the intention of his change. As far as I can tell, this change was done to make testing with OpenToFo working, since it's using the `manual` provider. So far, it seems like that it does not affect the platform tests. 


[0]: https://github.com/gardenlinux/gardenlinux/commit/5444eda458cf5337c6231a58c442b0427c1177e6
[1]: https://github.com/gardenlinux/gardenlinux/commit/e33221e52b212620fe1e8256c81669f30d98f39a      
[2]: https://github.com/gardenlinux/gardenlinux/pull/1725

tests/conftests/provisioner.py
```diff
@@ -0,0 +1,57 @@
...
+
+The parameter will be not be define by the feauter, but by a execution flag
```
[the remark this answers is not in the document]
Akendo: Thanks. I think, I need to update the term to use provisioner here instead.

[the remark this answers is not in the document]
Akendo: Updated

tests/conftests/provisioner.py
Akendo: I think we do not test our container images on any CRE as of now. Also, the container feature is fairly empty on that manner. We might, should add an issue for this. What do you think?

tests/conftests/platforms.py
Akendo: I've let the platform test run through: https://github.com/gardenlinux/gardenlinux/actions/runs/13589257417 they finished without a problem. Hence, this PR is good to go!

Vincinator: You want tests to run only if feature `container` is enable? 
```python3
@pytest.fixture
def container(testconfig):
    features = testconfig.get("features", [])
        if "container" not in features:
            pytest.skip('test only supported on container')
```
You want tests to run only if they these tests are executed with `pytest --iaas=container` then do: 
```python3
@pytest.fixture
def container(platform):
    if platform != 'container':
        pytest.skip('test only supported on container platform')
```

however, as you can see here: https://github.com/gardenlinux/gardenlinux/blob/7a5191bf4becd5899cc8bc9470fcae23226b3c44/tests/conftest.py#L105

there does not exist a `container` option.

Issue #2712, opened 2025-02-13
Run chroot tests as unprivileged user


nkraetzschmar:
Currently chroot tests run as root while platform tests run as a regular user.
This causes inconsistencies and results in failures during platform tests that could otherwise already be caught at unit test time.

To provide consistency there the chroot tests should also be run as a regular user.

PR #2714, opened 2025-02-13
Fix broken nightly by fixing the test shadow password test

moritzkirchnersap:
In some cases the tests are being run by an unprivileged user. In those cases we need to run the commands with sudo. In other cases, sudo doesn't exist on the system. In those cases we must run the command with sudo.

Nightly on this branch looks promising: https://github.com/gardenlinux/gardenlinux/actions/runs/13313736886

Minor sneak in changes:
 - the pytest.ini now contains -vvv as additional argument which shouldn't make a huge difference if the tests pass but in case they don't it can help to see the errors more clearly.
 - [tests/helper/tests/capabilities.py](https://github.com/gardenlinux/gardenlinux/pull/2714/files#diff-4c8e94db6da1afedbf8da96c36123aff6d6f8c93275b6c30f9012c614aa7b0d0) now doesn't fail any longer in some cases in which "match" wasn't defined

Broken nightly

**Definition of Done:**

**Release note**:

```feature user

```

### DIFF
--- a/features/base/test/test_password_shadow.py
+++ b/features/base/test/test_password_shadow.py
@@ -21,6 +21,12 @@ def test_password_shadow(client, command, expected_exit_code, expected_output):
     imply we execute read-only.
     """
     password_shadow(client)
+    outputs = []
+    exit_code, output = execute_remote_command(client, "whoami", skip_error=True)
+    outputs.append(output)
+    exit_code, output = execute_remote_command(client, "ls -la /etc/shadow", skip_error=True)
+    outputs.append(output)
+    assert outputs == []
     exit_code, output = execute_remote_command(client, command, skip_error=True)
     assert output == expected_output
     assert exit_code == expected_exit_code
--- a/features/base/test/test_password_shadow.py
+++ b/features/base/test/test_password_shadow.py
@@ -10,7 +10,7 @@ import pytest
 @pytest.mark.security_id(324)
 @pytest.mark.parametrize(
     "command, expected_exit_code, expected_output",
-    [("sudo -S /usr/sbin/pwck -r", 0, ""), ("sudo -S /usr/sbin/grpck -r", 0, "")],
+    [("pwck -r", 0, ""), ("grpck -r", 0, "")],
 )
 def test_password_shadow(client, command, expected_exit_code, expected_output):
     """This ensure that not only the passwd and shadow is as expected,
@@ -21,6 +21,11 @@ def test_password_shadow(client, command, expected_exit_code, expected_output):
     imply we execute read-only.
     """
     password_shadow(client)
-    exit_code, output = execute_remote_command(client, command, skip_error=True)
+    exit_code, _ = execute_remote_command(client, "ls /usr/bin/sudo", skip_error=True)
+    if exit_code == 0:
+        exit_code, output = execute_remote_command(client, f"sudo -S {command}", skip_error=True)
+    else:
+        exit_code, output = execute_remote_command(client, command, skip_error=True)
+
     assert output == "" 
     assert exit_code == expected_exit_code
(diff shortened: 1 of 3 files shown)


features/base/test/test_password_shadow.py
```diff
@@ -21,6 +21,11 @@ def test_password_shadow(client, command, expected_exit_code, expected_output):
     imply we execute read-only.
     """
     password_shadow(client)
-    exit_code, output = execute_remote_command(client, command, skip_error=True)
+    exit_code, _ = execute_remote_command(client, "ls /usr/bin/sudo", skip_error=True)
+    if exit_code == 0:
```
nkraetzschmar, line 25: won't the `sudo` binary always be there, regardless if the current user is root or not? so some check like `[ $(id -u) = 0 ]` or similar might be more useful.

nkraetzschmar (changes requested):
also relevant: https://github.com/gardenlinux/gardenlinux/issues/2712

features/base/test/test_password_shadow.py
moritzkirchnersap: > won't the sudo binary always be there

Apparently not. That's why this run: https://github.com/gardenlinux/gardenlinux/actions/runs/13312752314/job/37179387595#step:14:3568 / container build failed, at least exit code 127 can mean "command not found" and when I manually build the container I could't find sudo either.

The other solution I thought of would be to make sure sudo is always installed with something like feature/sudo or something of that sort.

nkraetzschmar: Ah, ok that makes sense, `sudo` shouldn't be installed in a pure container image.
I'd still say the check if we should use sudo should be based on if the current user is root or not. If the user is not root and there is no sudo, it's perfectly fine for the test to fail.

This just complicates things with running the chroot test as an unprivileged user, as this would imply that the container feature needs a different way to run the tests then any other features 🤔 

Or yes, we just always install sudo, probably don't even need a feature for that, I can't see any obvious issue with putting sudo into base (in general we need to re-consider what parts of server should actually belong into base).

features/base/test/test_password_shadow.py
```diff
@@ -21,6 +21,11 @@ def test_password_shadow(client, command, expected_exit_code, expected_output):
     imply we execute read-only.
     """
     password_shadow(client)
-    exit_code, output = execute_remote_command(client, command, skip_error=True)
+    exit_code, _ = execute_remote_command(client, "ls /usr/bin/sudo", skip_error=True)
```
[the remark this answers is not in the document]
moritzkirchnersap: > irrelevant nitpick, but I think which is the right tool to see if a program is available (and in path).
Ah, yes, that would have been also an option. I was thinking about touch without create, basically just anything that checks for the existence of a file and gives a proper exit code. Gone now anyway...

@nkraetzschmar : changes included, what do you think? Nightly to check looks good:

https://github.com/gardenlinux/gardenlinux/actions/runs/13326183549/job/37220009790

features/base/test/test_password_shadow.py
```diff
@@ -21,6 +21,10 @@ def test_password_shadow(client, command, expected_exit_code, expected_output):
     imply we execute read-only.
     """
     password_shadow(client)
-    exit_code, output = execute_remote_command(client, command, skip_error=True)
+    exit_code, output = execute_remote_command(
+        client,
+        f"if [ $(id -u) = 0 ]; then {command}; else sudo -S {command}; fi",
```
fwilhe, line 26: I am not a fan of short options in scripts, i always have to look up what `-S` means

```
  -S, --stdin                   read password from standard input
```

1. why do we need this? don't we have passwordless sudo?
2. if possible please use the long option

fwilhe (approved):
lgtm, please address my comment before merge

features/base/test/test_password_shadow.py
moritzkirchnersap: > why do we need this? don't we have passwordless sudo?

I saw some tests where the command failed because there was no password input for sudo and the error hint was to either use -S or --stdin. I tried to prevent such failures, but can be tried without the option if that's preferrable.

Vincinator (approved):
LGTM

Issue #2711, opened 2025-02-13
Rework GitHub build workflow


NotTheEvilOne:
# Epic: Rework GitHub build workflow

**Goal:** A fine-granular set of workflows to build and test PRs, create nightlies and releases. As seen lately in #2676 and #2700 issues occur with the current `nightly` build workflow. Images and tags in the GitHub OCI repository get overwritten by concurrent or later runs of the same day. Splitting and adding save-guard measures should reduce the risk at least significantly.

### Mar 
- https://github.com/gardenlinux/gardenlinux/issues/2751
- https://github.com/gardenlinux/gardenlinux/issues/2752
- https://github.com/gardenlinux/gardenlinux/issues/2753
- https://github.com/gardenlinux/gardenlinux/issues/2754

moritzkirchnersap:
🙏

PR #2705, opened 2025-02-12
feat: support running platform tests without KMS backed keys

nkraetzschmar:

PR #2716, opened 2025-02-14
feat: add skip_upload option to nightly workflow

nkraetzschmar:

Add option to skip upload. Useful to test only platform tests from branches without permissions to write to S3.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -17,6 +17,9 @@ on:
         type: string
         required: true
         default: "nightly"
+      skip_upload:
+        type: boolean
+        default: false
 jobs:
   build:
     uses: ./.github/workflows/build.yml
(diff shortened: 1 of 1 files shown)


Vincinator (approved):
This as an option is good idea. LGTM

NotTheEvilOne (approved):
LGTM

Referencing in #2711

moritzkirchnersap:
Should this then be a mandatory merge requirement so that no merge breaks the nightly?

nkraetzschmar:
> Should this then be a mandatory merge requirement so that no merge breaks the nightly?

Yes, at least in combination with #2705, both together allow to run the nightly without having special permissions on the branch except for the permissions to create the platform test resources. We should consider if we can create some automation to add the branch to the platform test OIDC allow list and run the nightly. Obviously not fully automatic for any PR but for any we mark in some special way.

PR #2727, opened 2025-02-18
Add auto rerun on GHCR upload

paulphys:
Implements retry logic for **502 Bad Gateway** errors on GHCR upload. After 3 unsuccessful retries the action should then terminate.


nkraetzschmar (approved):
lgtm

PR #2749, opened 2025-02-20
Fix/test supported distro

Akendo:
In this PR we re-enable the `gl_supported_test` that was reverted.


features/base/test/test_basics.py
```diff
@@ -1,8 +1,33 @@
...
 
 
+@pytest.mark.security_id(1)
+def test_gl_is_support_distro(client, container, non_container):
+    """
+    This tests ensures that the vendor field is set to 'gardenlinux'.
+    """
+
+    # We have to enable sudo to allow.
+    if non_container:
```
Vincinator: if you add `non_container` as parameter to the function, the fixture [non_container](https://github.com/gardenlinux/gardenlinux/blob/3347229d6e09cf786a39fed0d60b32319e92f8ba/tests/conftest.py#L495) makes sure that the test is not even executed for any image with the feature `container` enabled. So this if statement here should never be true.

Unfortunately, the fixture [container](https://github.com/gardenlinux/gardenlinux/blob/3347229d6e09cf786a39fed0d60b32319e92f8ba/tests/conftest.py#L430) is not the negation of `non_container`. This is misleading, we need to fix this.

My suggestion is:
1.  split test_gl_is_support_distro in two test functions, one for container case (not using sudo) and one for non_container case (using sudo)
2. refactor the current [container](https://github.com/gardenlinux/gardenlinux/blob/3347229d6e09cf786a39fed0d60b32319e92f8ba/tests/conftest.py#L430) fixture to a different name 
3. add an fixture named container that does actually the oposite of non_container 


I know, the container fixture was not part of your PR, but I don't know a way around it.

Vincinator (changes requested):

features/base/test/test_basics.py
Akendo: You're right. I wanted to be a smart ass, and that turned out to be a really dumb decision. Going to refactor this a bit.

features/base/test/test_basics.py
```diff
@@ -1,17 +1,28 @@
...
 from helper.sshclient import RemoteClient
 
 
+@pytest.mark.security_id(1)
+def test_gl_is_support_distro(client):
+    """
+    This tests ensures that the vendor field is set to 'gardenlinux'.
+    """
+    os_releases = read(client, file="/etc/os-release")
+    assert os_releases[0] == "ID=gardenlinux"
```
moritzkirchnersap, line 14: Not having checked yet how read_file_remote.read() works, isn't it a strong assumption that this is always in the first place?

Vincinator: Garden Linux base feature writes it here:
https://github.com/gardenlinux/gardenlinux/blob/61f50ec4e7ccddc494379e7e4cead86ebb0ea6ab/features/base/exec.config#L5

I am not aware of any reason why we must have `ID=gardenlinux` in first line of /etc/os-release

fwilhe: after reading the [man page](https://www.freedesktop.org/software/systemd/man/latest/os-release.html) I think that it should not be assumed that ID is in a given place.

I think

```
assert "ID=gardenlinux" in os_releases
``` 

should do the right thing in python..

Akendo: Thanks for the suggestion. I'm going to test for it correctly.

moritzkirchnersap (approved):

Vincinator (approved):

PR #2759, opened 2025-02-21
Fix/dockerignore symlink for testcontainer buildcontext

Vincinator:

This commit tells podman to ignore the `tests/platformSetup/tofu/cert` symlink, which points outside the build context used in [./test](https://github.com/gardenlinux/gardenlinux/blob/3347229d6e09cf786a39fed0d60b32319e92f8ba/test#L63).

When executing `./test <target>` podman build will bail out before starting to build tests/Containerfile, since a symlink in the provided build context is pointing outside. 

```
error: invalid symlink "/var/tmp/libpod.../build/platformSetup/tofu/cert" -> "../../../cert"
```

**Notes for Reviewer** 
Need to check if we need to rename the .dockerignore file and tell the ./test script to explicitly ignore the renamed file, so we avoid any other podman build ignoring that symlink where we actually need it.


nkraetzschmar (approved):
lgtm

PR #2761, opened 2025-02-23
Use different path pattern for caching certificates

NotTheEvilOne:
This PR adds support for symlinked `cert` directories like used for [`gardenlinux-ccloud`](https://github.com/gardenlinux/gardenlinux-ccloud).

Fixes https://github.com/gardenlinux/gardenlinux-ccloud/issues/42

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,9 +77,9 @@ jobs:
           for f in secureboot.{{pk,null.pk,kek,db}.{crt,der,auth},db.arn,aws-efivars}; do
             ln -sr "cert/gardenlinux-$f" "cert/$f"
           done
-      - uses: actions/cache/save@d4323d4df104b026a6aa633fdb11d772146be0bf # pin@v3
+      - uses: actions/cache/save@1bd1e32a3bdc45362d1e726936510720a7c30a57 # pin@v4.2.0
         with:
-          path: cert
+          path: cert/*.*
           key: cert-${{ github.run_id }}
   base:
     name: bootstrap stage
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -73,13 +73,13 @@ jobs:
       - name: use kms backed certificates
         if: ${{ inputs.use_kms }}
         run: |
-          touch cert/gardenlinux-secureboot.db.arn
+          touch ./cert/gardenlinux-secureboot.db.arn
```
nkraetzschmar, line 76: does this change actually make a difference? this should be functionally identical, no?

NotTheEvilOne: No, there's no functional difference here. Intended to improve readability here but after all it doesn't change much anyway.

.github/workflows/build.yml
```diff
@@ -77,9 +77,9 @@ jobs:
           for f in secureboot.{{pk,null.pk,kek,db}.{crt,der,auth},db.arn,aws-efivars}; do
             ln -sr "cert/gardenlinux-$f" "cert/$f"
           done
-      - uses: actions/cache/save@1bd1e32a3bdc45362d1e726936510720a7c30a57 # pin@v3
+      - uses: actions/cache/save@1bd1e32a3bdc45362d1e726936510720a7c30a57 # pin@v4.2.0
         with:
-          path: cert
+          path: cert/*.*
```
Vincinator, line 82: thanks for taking care of this.

Interesting that this action references actions/toolkit but do not seem to expose the `follow-symbolic-links` feature, at least I could not find it in actions/cache repo. 

Do you have a test run that you can link by any chance? 

Otherwise looks good to me

Vincinator: I would assume that cert/*.* would force the action to follow the symlinks via wildcard (implemented by mentioned https://github.com/actions/toolkit/tree/main/packages/glob)

Vincinator (approved):

PR #2779, opened 2025-02-26
Create package diff using glvd for release notes

fwilhe:

Add new Release to glvd before creating release page, and use the glvd api to get a list of upgraded packages and fixed CVE.

Fixes https://github.com/gardenlinux/glvd/issues/147

Note to reviewer: This can't be properly tested on a pr branch, we'll have to see how well it works in main

### DIFF
--- a/.github/workflows/manual_gh_release_page.yml
+++ b/.github/workflows/manual_gh_release_page.yml
@@ -20,9 +20,55 @@ on:
         type: string
         description: "Full commitish"
 jobs:
+  # Create new version in GLVD so it has the package list of the new release
+  # This is needed for the automatic changelog generation
+  glvd:
+    if: ${{ inputs.type }} == 'patch'
+    name: Update GLVD distro list
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      # This is required for requesting the JWT
+      id-token: write
+    needs:
+      - glrd
+    steps:
+      - name: Get OIDC token
+        id: get-token
+        run: |
+          IDTOKEN=$(curl -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=glvd" | jq -r '.value')
+          echo "idToken=${IDTOKEN}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


.github/workflows/release-page.yml
```diff
@@ -136,3 +139,28 @@ jobs:
...
+    steps:
+      - name: Get OIDC token
+        id: get-token
+        run: |
+          IDTOKEN=$(curl -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=glvd" | jq -r '.value')
+          echo "idToken=${IDTOKEN}" >> $GITHUB_OUTPUT
+      - uses: azure/k8s-set-context@v4
+        with:
+           method: kubeconfig
+           kubeconfig: "${{ secrets.KUBECONFIG }}"
```
fwilhe, line 164: This secret is not yet created, and the repo also needs to have permission for the OIDC token in the cluster

Vincinator:
would it make sense to test stuff like this in a fork?

fwilhe:
> would it make sense to test stuff like this in a fork?

Maybe, not sure how much effort that would be because this also has dependencies on GLRD and the gardener cluster where glvd is deployed. I was thinking about this, but it sounded like a forbidden amount of effort to me.

Maybe we can evaluate this together, you might have ideas for how to simplify this that I don't think of.

.github/workflows/release-page.yml
```diff
@@ -23,6 +23,8 @@ jobs:
   create_github_release:
     if: ${{ inputs.type }} == 'patch'
     environment: oidc_aws_s3_upload
+    needs:
+      - glvd
```
Vincinator, line 27: For readability, I would suggest to place the definitions of the jobs in the order they run:
1. glrd
2. glvd 
3. create_github_release

yeoldegrove (changes requested):
@fwilhe Could we put all the steps in the `glvd` job into a seperate workflow?

Vincinator (approved):

moritzkirchnersap (approved):
Looks reasonable to me, and for me it's always hard to review github action workflows. If you want to test it: go ahead!

Issue #2786, opened 2025-02-26
Merge test for security_id 165


Akendo:
Ensure that the test for security_id 165 will be merged.

Akendo:
The PR is open https://github.com/gardenlinux/gardenlinux/pull/2820

Akendo:
Replaced by DISA STIG.

PR #2820, opened 2025-02-27
feature/test sudo

Akendo:
In this PR, we add a basic test for `sudo` that ensures that we do have the correct configuration for `sudo` in place. 
It will check that the `%sudo` group is present and document as well that we do not use the `wheel` group for this. 

Also, we have added the security id 165 mark to it. 

Fixes: #2786


features/base/test/test_sudo.py
```diff
@@ -0,0 +1,24 @@
...
+    assert output.split(":")[0] == 'sudo', "Group 'sudo' is missing!"
+
+    sudoers = read_file_remote(client,
+                               file="/etc/sudoers",
+                               remove_comments=True,
+                               remove_newlines=True)
+    # Drop sudo permissions.
+    client._default_to_sudo = False
+    _ = [config for config in sudoers if '%sudo' in config]
+    assert len(_) == 1, "Missing sudo configuration in '/etc/sudoers'!"
```
moritzkirchnersap, line 24:
```suggestion
    assert len(
        [config for config in sudoers if '%sudo' in config]
    ) == 1, "Missing sudo configuration in '/etc/sudoers'!"
```

features/base/test/test_sudo.py
```diff
@@ -0,0 +1,26 @@
...
+    Validate that we have the necessary %sudo configration entry in sudoers.
+    Note: we do not use wheel, but instead the sudo group.
+    """
+    # Ensure we have sudo permissions.
+    client._default_to_sudo = True
+    output = run(client, cmd="getent group sudo")
+    assert output.split(":")[0] == 'sudo', "Group 'sudo' is missing!"
+
+    sudoers = read(client,
+                   file="/etc/sudoers",
```
Vincinator, line 18: Do we want to check /etc/sudoers.d/* as well?

Vincinator: may check if we have `@includedir`, `#include` or `#includedir` (see [man of sudoers](https://www.man7.org/linux/man-pages/man5/sudoers.5.html) chapter Including other files from within sudoers)

features/base/test/test_sudo.py
```diff
@@ -0,0 +1,26 @@
...
+                   remove_comments=True,
+                   remove_newlines=True
+                   )
+    # Drop sudo permissions.
+    client._default_to_sudo = False
+    assert len(
+        [config for config in sudoers if '%sudo' in config]
+    ) == 1, "Missing sudo configuration in '/etc/sudoers'!"
```
Vincinator, line 26: I dont know what we actually want to test here.

two remarks:
1. position of %sudo in line is not considered, commented out #%sudo would also count as valid
2. multiple occurrences of %sudo not considered by this test 

```python
if '%sudo' in config
```
does not care about where in line `%sudo` is located. 

particularly, a commented out %sudo would pass this test
```
# %sudo ALL=(ALL) ALL 
```
=> test_sudo would pass

features/base/test/test_sudo.py
```diff
@@ -0,0 +1,26 @@
+from helper.utils import read_file_remote as read
+from helper.utils import execute_remote_command as run
+import pytest
+
+
+@pytest.mark.security_id(165)
+def test_sudoers(client, non_container):
+    """
+    Validate that we have the necessary %sudo configration entry in sudoers.
+    Note: we do not use wheel, but instead the sudo group.
```
Vincinator, line 10: Server feature includes wheel in `/etc/sudoers.d/wheel` 
https://github.com/gardenlinux/gardenlinux/blob/main/features/server/file.include/etc/sudoers.d/wheel

Would this be an issue?

Akendo: Thanks for point this out, wasn't aware of this.

PR #2822, opened 2025-02-28
[DRAFT]feature/test for root

Akendo:

Fixes #

**Definition of Done:**

**Release note**:

```feature user

```


features/base/test/test_users.py
```diff
@@ -1,6 +1,38 @@
...
+    login_name = 0
+    numeric_user_id = 2
+
+    # test if we have only a single user with root and id 0.
+    entries_with_root_name = [entry for entry in passwd if 'root' in
+                              entry.split(":")[login_name]]
+    entries_with_root_id = [entry for entry in passwd
+                            if '0' == entry.split(":")[numeric_user_id]]
+    assert len(entries_with_root_name), 1
+    assert len(entries_with_root_id), 1
```
moritzkirchnersap, line 38:
```suggestion
    number_of_users = execute_remote_command('''python -c 'import pwd; print(len([entry for entry in pwd.getpwall() if entry.pw_name == "root" and entry.pw_uid == 0]))'''')
    assert int(number_of_users) == 1, "There should only be one user with login_name 'root' and user_id 0"
```

moritzkirchnersap: I'm not very happy with this suggestion but I'm a bit proponent of using standard python modules where they exist. If you have an idea how to elegantly use this python module remotely - I'd be open to it, but I doubt we have to re-invent the wheel...

moritzkirchnersap (changes requested):

features/base/test/test_users.py
Akendo: This won't work, since `python` isn't on every system present. Instead, we're using `bash` the way, since that's what we have always at our disposal. 

I think what we need is something like `sshfs`. There is a python module allowing us to use this with a context manager. The problem is that we not use `root` by default for the platform test (and that's good). But the module can't deal with `sudo`. There are some methods to elevate a `sshfs` connection to be root, but I fear that this would be another hack.

Instead, we have to refactor our `RemoteClient` or have a class on top, that implement's the necessary context manager. I can't afford the time to do it as of now.

Issue #2787, opened 2025-02-26
Merge test for security_id 166


Akendo:
Ensure that the test for security_id 166 will be merged.

pnpavlov:
@Akendo , re-assigning as example for the proposal shared [here](https://github.com/gardenlinux/security/issues/79#issuecomment-2689373035).

Akendo:
opened the PR for it https://github.com/gardenlinux/gardenlinux/pull/2822

PR #2823, opened 2025-02-28
feature/test su

Akendo:
In this PR, we added a test to ensure that access to `su` is restricted by the OS. 

The test will read the file from `/etc/pam.d/su` and check if it has the configuration parameter we expect.
A user needs to be configured to be admitted to use `su`. By default, this is realized by a specify group, the `wheel` group. We have to ensure that this behavior is enabled. 

This way, the restricting of what target identities are allowed to elevate is done via the group `wheel` and by not fizzing around with the PAM configuration files. 

Closes: #2787


features/base/test/test_su.py
```diff
@@ -0,0 +1,23 @@
...
+
+    pam_configuration = read_file_remote(client,
+                                         file="/etc/pam.d/su",
+                                         remove_comments=True)
+
+    auth_config = [config for config in pam_configuration if 'auth' in config]
+    for entry in auth_config:
+        if 'pam_wheel.so' in entry:
+            assert 'required pam_wheel.so' in entry, \
+              "Error configuration for the pam_wheel.so"
```
Vincinator, line 19: May I suggest 
```suggestion
              "control value of PAM Module pam_wheel.so must be set to required"
```

or something that explains what went wrong

features/base/test/test_su.py
```diff
@@ -0,0 +1,23 @@
...
+            test_successfully = True
+
+    if not test_successfully:
+        assert False, "Error pam_wheel.so not found!"
```
Vincinator, line 23:
```suggestion
            test_successfully = True

    assert test_successfully, "pam_wheel.so not set as PAM Module"
```

moritzkirchnersap (changes requested):
Please use black for formatting

moritzkirchnersap (approved):

PR #2862, opened 2025-03-17
fix login to tofu platform test targets

yeoldegrove:

Fix path to generated login script for tofu platform test targets.


NotTheEvilOne (approved):
LGTM

Is the output of e.g.
```
?? tests/platformSetup/tofu/login.<platform>-amd64.sh
INFO: Pytest configfile 'tests/config/pytest.manual.<platform>-amd64.yaml' created.
```
intended?

yeoldegrove:
@NotTheEvilOne The output was intended to remind you that you have an unclean git dir. In `tests/platformSetup/tofu` there should not be any login scripts anymore as they are moved to `tests/platformSetup` now.

Issue #2868, opened 2025-03-18
Investigate rate limiting issue uploading to OCI


paulphys:
We're occasionally hitting the [2000 requests per minute rate limit](https://github.com/gardenlinux/gardenlinux/actions/runs/13333857947/job/37248900661) when uploading to GHCR. I'm not sure how we are even getting close to this limit as we don't have that many concurrent uploads to GHCR.

```
ERROR: oras.logger:retry-after: 1.889377ms, allowed: 2000/minute
Error: Issue with https://ghcr.io/v2/gardenlinux/gardenlinux/blobs/upload/... : Too Many Requests
Error: Process completed with exit code 1.
```

This is also the reason the [nightly currently fails](https://github.com/gardenlinux/gardenlinux/actions/runs/13916470153/job/38944455969) to complete the OCI uploads.

Resolving this might be as easy as just sleeping a few seconds in between uploading each artifact or limiting the amount of concurrent jobs

PR #2871, opened 2025-03-18
Limit concurrent jobs uploading to OCI to prevent rate limiting

paulphys:
Addresses https://github.com/gardenlinux/gardenlinux/issues/2868

Limiting the amount of concurrent jobs to 5 in order to prevent rate limiting.
If this issue still persists after that we could also try adding sleep statements either in the [upload_oci.yml](https://github.com/gardenlinux/gardenlinux/blob/main/.github/workflows/upload_oci.yml) or directly to [python-gardenlinux-cli](https://github.com/gardenlinux/python-gardenlinux-cli/blob/main/src/registry.py )

### DIFF
--- a/.github/workflows/upload_oci.yml
+++ b/.github/workflows/upload_oci.yml
@@ -31,6 +31,7 @@ jobs:
       packages: write
     environment: oidc_aws_s3_upload
     strategy:
+      max-parallel: 5
       fail-fast: false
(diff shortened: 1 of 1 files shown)


yeoldegrove (approved):
Limiting this makes sense to me.
Let's test out how high we can go and how long the upload takes.

PR #2878, opened 2025-03-20
feat: remove rootlesskit from chost

5kt:
After discussing with @gehoern it was decided to remove _rootlesskit_ from the _chost_ feature.
At the moment nobody using this feature really needs the functionality provided by _rootlesskit_.


gehoern (approved):
i am against this :-D

PR #2881, opened 2025-03-21
fix: extra space on runner

5kt:
Get extra space on runner so the nightly platform tests can correctly run.

Fixes #


5kt:
#2880 already took care of this - closing.

PR #2894, opened 2025-03-26
feat(capi): use sulogin --force when root is locked

5kt:
When using _metal-capi_ in CC, any failure that would trigger the emergency service (initrd or regular system) a usable shell won't be started if the root account is locked.
As this is needed for debug purposes, setting SYSTEMD_SULOGIN_FORCE=1 is needed[^1]

[^1]: https://systemd.io/ENVIRONMENT/


nkraetzschmar:
I don't think having this in any production feature is a good idea. Maybe this should be in the _dev feature instead?

5kt:
Good point by @nkraetzschmar 
This was discussed in the team and it's not needed, closing.

PR #2896, opened 2025-03-27
test rp_filter in Gardener feature not set to 1

MrBatschner:

For the Gardener feature, the sysctls `net.ipv4.conf.all.rp_filter` and `net.ipv4.conf.default.rp_filter` must not be set to `1` as it will interfere with CNI plugins like Calico or Cilium.

This test tests if those sysctls are set to 1 and will fail if so.

For this test to work, I created an new test helper `check_sysctl()`. Even though we already had a test helper `test_kernel_parameter()` (this name is utterly misleading btw), it is too inflexible as it only allows to test for a certain sysctl to be set to a certain value. With `check_sysctl()`, I introduced an operator with which it is possible to test wether a sysctl `is` a certain value, `isnot` a certain value, `isnotorisnotset` a certain value, `isset` but the value does not matter, `isnotset` at all.

**Definition of Done:**

Should be backported to `rel-1592` as 1592.7 introduced a regression that makes this test necessary.

### DIFF
--- a/tests/helper/tests/sysctl.py
+++ b/tests/helper/tests/sysctl.py
@@ -1,0 +1,39 @@
+import logging
+import pytest
+
+logger = logging.getLogger(__name__)
+
+
+def check_sysctl(client, key, operator, value):
+    (exit_code, output, error) = client.execute_command(
+        "sudo -u root /usr/sbin/sysctl -a", quiet=True
+    )
+    assert exit_code == 0, f"no {error=} expected"
+
+    sysctls = {}
+    for line in output:
+        if len(line) < 1 or line.startswith("#"):
+            continue
+        s = line.split("=", 1)
+        if len(s) < 2:
+            continue
+        sysctls[s[0].strip()] = s[1].strip()
+
+    match operator:
+        case "is":
+            assert key in sysctls
+            assert sysctls[key] == value, f"{key} is not {value} but {sysctls[key]}"
+        case "isnot":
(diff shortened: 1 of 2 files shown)


fwilhe (approved):

tests/helper/tests/sysctl.py
```diff
@@ -0,0 +1,35 @@
+import logging
+import pytest
+
+logger = logging.getLogger(__name__)
+
+
+def check_sysctl(client, key, operator, value):
+    (exit_code, output, error) = client.execute_command("sysctl -a", quiet=True)
```
Vincinator, line 8: sysctl binary is in `usr/sbin/sysctl`

From platform test run 
```
FAILED gardener/test/test_sysctls.py::test_sysctls[net.ipv4.conf.all.rp_filter-isnotornotset-1] - AssertionError: no error='bash: line 1: sysctl: command not found\n' expected
FAILED gardener/test/test_sysctls.py::test_sysctls[net.ipv4.conf.default.rp_filter-isnotornotset-1] - AssertionError: no error='bash: line 1: sysctl: command not found\n' expected
```

fwilhe: Thanks for checking, I approved without checking if those tests were actually running. Should have been more careful.

MrBatschner:
Aand it failed again, making this a draft for now, apparently there is something else wrong with it.

Vincinator (approved):

PR #2919, opened 2025-03-29
Add support to intersect matrices for use in GitHub workflows and tests

NotTheEvilOne:
This PR adds support to intersect flavor matrices in GitHub workflows to execute tests only for enabled flavors.


Vincinator (approved):
LGTM. let's try next nightly with it already - will merge now.

PR #2929, opened 2025-03-31
fix: support arm64 machine hardware type for platform-test images

Vincinator:
`arm64` or `aarch64` should both resolve to `arm64` in this case. 

`uname -m` on osx returns `arm64`, which would result in an empty `TARGETARCH` variable during building , since no existing conditions are met. 

**Reproduce**:  
```
make --directory=tests/images build-platform-test
# Check logs for linux-image (see that apt can not install package, because we only have linux-image-${TARGETARCH}
# Check logs for TARGETARCH= 
```

### DIFF
--- a/tests/images/Makefile
+++ b/tests/images/Makefile
@@ -15,6 +15,8 @@ endif
 ifeq ($(strip $(TARGETARCH)),)
 ifeq ($(shell uname -m), aarch64)	
 TARGETARCH := arm64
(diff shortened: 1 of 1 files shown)


yeoldegrove (approved):
Cannot test, but LGTM.

NotTheEvilOne (approved):
LGTM

PR #2932, opened 2025-03-31
Fix "ambiguous argument" in GitHub workflow executed `git` for `publish_s3.yml`

NotTheEvilOne:
This PR fixes an issue in GitHub workflow `publish_s3.yml`:
```
fatal: ambiguous argument '<short rev>': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
```

While not occured locally the Ubuntu variant seems to be more restrictive regarding command line parameters.

### DIFF
--- a/.github/workflows/publish_s3.yml
+++ b/.github/workflows/publish_s3.yml
@@ -189,7 +189,7 @@ jobs:
           aws-region: ${{ secrets.AWS_REGION }}
       - id: git_hash
         name: Get Git hash for "${{ needs.workflow_data.outputs.commit_id }}"
-        run: echo "hash=$(git log -1 --pretty=format:"%H" "${{ needs.workflow_data.outputs.commit_id }}")" | tee -a $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_s3.yml
```diff
@@ -189,7 +189,7 @@ jobs:
           aws-region: ${{ secrets.AWS_REGION }}
       - id: git_hash
         name: Get Git hash for "${{ needs.workflow_data.outputs.commit_id }}"
-        run: echo "hash=$(git log -1 --pretty=format:"%H" "${{ needs.workflow_data.outputs.commit_id }}")" | tee -a $GITHUB_OUTPUT
+        run: echo "hash=$(git log -1 --pretty=format:"%H" "${{ needs.workflow_data.outputs.commit_id }}")" -- . | tee -a $GITHUB_OUTPUT
```
fwilhe, line 192: should this not be an argument to git? I think then it should be within the parentheses, if I'm not mistaking

NotTheEvilOne: You're totally right.

PR #2940, opened 2025-04-01
Re-enable Create package diff using glvd for release notes

fwilhe:
In #2779, I forgot to pass the OIDC token to kubectl which is why the workflow failed. This should fix that.

Follow-up for #2779


Akendo (approved):
LGTM

PR #2941, opened 2025-04-02
fix test: disable test_clocksource on qemu without kvm support

Vincinator:

The test

	base/test/test_time_config.py::test_clocksource

will fail with

	AssertionError: unknown hypervisor qemu

This happens because systemd-detect-virt will output "qemu" in case we start tests via qemu target, e.g.:

	make --directory=tests aws-gardener_prod-amd64-qemu-test-platform

In order to exclude this test for this test target, we need to add new fixtures for "qemu" iaas. As defined in [tests/conftest.py::testconfig](https://github.com/gardenlinux/gardenlinux/blob/d659d9bc8a843ae1ba49bbff9fb2e14a826fd6b8/tests/conftest.py#L173), iaas can be set to qemu.

This commit adds:
- fixtures to exclude/include qemu iaas
- excludes test_time_config.py::test_clocksource for qemu platform

### DIFF
--- a/features/base/test/test_time_config.py
+++ b/features/base/test/test_time_config.py
@@ -100,7 +100,7 @@ __EOF
     assert "FAIL" not in output
 
 
-def test_clocksource(client, aws):
+def test_clocksource(client, aws, non_qemu):
     """ Test for clocksource """
     # refer to https://aws.amazon.com/premiumsupport/knowledge-center/manage-ec2-linux-clock-source/
     # detect hypervisor type kvm or xen
--- a/tests/conftests/platforms.py
+++ b/tests/conftests/platforms.py
@@ -53,6 +53,11 @@ def kvm(platform):
         pytest.skip('test only supported on kvm platform')
 
 @pytest.fixture
+def qemu(platform):
+    if platform != 'qemu':
+        pytest.skip('test only supported on qemu platform')
+
+@pytest.fixture
 def metal(platform):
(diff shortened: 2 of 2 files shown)


fwilhe:
Naive question: I assume aliasing qemu to kvm would not work (at least not in cases where qemu is run without kvm)?

fwilhe (approved):

Vincinator:
> Naive question: I assume aliasing qemu to kvm would not work (at least not in cases where qemu is run without kvm)?

Absolutely. This PR addresses two issues. 

1. The test_clocksource is not written for the case qemu VM without kvm support. It is not clear what the clock source should be on amd64 architectures without kvm support (tsc, hpet, ...) => disable it if this test is executed on qemu test platform 
2. introduce option to disable tests on qemu platform. The qemu target (defined [here](https://github.com/gardenlinux/gardenlinux/blob/d659d9bc8a843ae1ba49bbff9fb2e14a826fd6b8/tests/platformSetup/Makefile#L147)) passes on the `--provisioner qemu`, and in the pytest fixture the argument `platform` of fixtures
```python
@pytest.fixture
def blablahblah(platform)
```
 will evaluate to whatever the respectively named fixture [platform](https://github.com/gardenlinux/gardenlinux/blob/b8816c28be3f50b8896518fa75aac79f1cb9ec15/tests/conftest.py#L85) will return, in the case above it will be `qemu`

PR #2955, opened 2025-04-10
Support branch agnostic GitHub workflows

### LINKED ISSUE #2950: Support GitHub workflows for release branches
Analyze options to support GitHub workflows for release branches and older variants of `Makefile` etc.

**Why is this needed**:
GitHub UI supports only "actions" provided by the `main` branch to be triggered manually. Renaming and improvements to the workflows can not be used in current and future (if changed / enhanced again) release branches. A common source of workflows could help to unify these workflows.

NotTheEvilOne:
This PR adds support to run GitHub workflows from `main` in a branch agnostic way.

Closes #2950


.github/workflows/build_flavor.yml
```diff
@@ -78,7 +82,8 @@ jobs:
           t="$(date '+%s')"
           find .build -exec touch -d "@$t" {} +
       - name: Build
-        run: make ${{ inputs.flavor }}-${{ inputs.arch }}-build
+        # @TODO: run: make ${{ inputs.flavor }}-${{ inputs.arch }}-build
```
Akendo, line 85: What does this TODO imply here? DO we need to change the `Makefile`? Did we create a ticket for this already?

.github/workflows/build_kmodbuild_container.yml
```diff
@@ -67,7 +71,13 @@ jobs:
           podman login -u token -p "${GH_TOKEN}" ghcr.io
           podman pull --arch ${{ matrix.arch }} "$snapshot"
 
-          podman build --arch ${{ matrix.arch }} --build-arg base="$base" --build-arg snapshot="$snapshot" -t ghcr.io/${{ github.repository }}/kmodbuild:${{ matrix.arch }}-${version} images/kmodbuild
+          # @TODO: Changed between supported releases - replaced by images/kmodbuild
```
Akendo, line 74: Does this imply now, that we can't build kmodbuilds?

NotTheEvilOne: No, it's here to remove the `if`/`else` statement once releases using the old path are out of maintenance.

.github/workflows/build_flavor.yml
NotTheEvilOne: The `Makefile` is unfortunately only available for release 18**. Once older releases are out of maintenance we can switch to it. An discussion about using `./build` or `make` is ongoing in #3015.

nkraetzschmar (changes requested):
before we proceed with this, we'll need to discuss how this interacts with our security assumptions, seeing as github action environment rules rely on branches

NotTheEvilOne:
Progress is blocked until a solution has been found for branch specific environments / secrets. Adding top-level (triggerable) workflow files using files like `build.yml` etc. from main (branch agnostic) would mean that input parameter and top-level workflow files need to be copied over to each (release) branch.

We can take the learnings from the branch agnostic approach and move workflows to a separate repository like "Gardener" uses partly [1] [2]. This still would require interaction between integrating repository/ies (like "gardenlinux") and the workflow repository. @gehoern, @pnpavlov

[1] https://github.com/gardener/cc-utils/.github/workflows/
[2] https://github.com/gardener/gardener-extension-provider-aws/blob/master/.github/workflows/build.yaml

PR #2961, opened 2025-04-10
try first to upload artifacts via GitHub hosted runners, if that fails fallback to custom upload runner

Shegox:
This PR adjust the behavior of the upload to S3 GitHub action to leverage GitHub hosted runners

Currently all upload jobs are using custom paid GitHub-hosted runners. Trying to upload first with the free GitHub-hosted runners avoids paying for these. The fallback to the custom `upload` runners ensures that in case the GitHub-hosted runners get blocked (especially for china uploads) the custom runners are used. 
The hope is that this will not happen to often and thus cuts the cost of the runners a lot. 

Currently `publish_s3.yml` already consumed 200k paid minutes in the first 10 days of April: https://github.com/gardenlinux/gardenlinux/actions/metrics/usage

n/a

**Definition of Done:**

To allow for better retry behavior for individual uploads the matrix is now generated in the main `publish_s3.yml` and `upload_to_s3` is only called with one specific `flavor` & `arch` to upload that single thing.
If that upload fails using the GitHub hosted runner (`ubuntu-24.04`)  it is retried within `upload_to_s3` via the `upload` custom runner.

Rather than the ugly code duplication for both upload attempts the same could be archived using an external action, but as the duplicated code is rather small and anyway was mentioned that this should be removed in the future I think this is okay for now.

//cc @nkraetzschmar 

**Release note**:

```other developer
Upload artifacts to S3 using GitHub hosted runners; fallback to custom runners on failure
```

### DIFF
--- a/.github/workflows/publish_s3.yml
+++ b/.github/workflows/publish_s3.yml
@@ -147,11 +147,15 @@ jobs:
     name: Upload to S3
     permissions:
       id-token: write
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.workflow_data.outputs.flavors_matrix) }}
     uses: ./.github/workflows/upload_to_s3.yml
     with:
       commit_id: ${{ needs.workflow_data.outputs.commit_id }}
       version: ${{ needs.workflow_data.outputs.version }}
-      flavors_matrix: ${{ needs.workflow_data.outputs.flavors_matrix }}
+      flavor: ${{ matrix.flavor }}
+      arch: ${{ matrix.arch }}
       run_id: ${{ github.run_id }}
     secrets:
       aws_region: ${{ secrets.AWS_REGION }}
@@ -163,11 +167,15 @@ jobs:
     name: Upload to S3 (China)
     permissions:
       id-token: write
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.workflow_data.outputs.flavors_matrix) }}
     uses: ./.github/workflows/upload_to_s3.yml
     with:
       commit_id: ${{ needs.workflow_data.outputs.commit_id }}
       version: ${{ needs.workflow_data.outputs.version }}
-      flavors_matrix: ${{ needs.workflow_data.outputs.flavors_matrix }}
+      flavor: ${{ matrix.flavor }}
+      arch: ${{ matrix.arch }}
       run_id: ${{ github.run_id }}
     secrets:
       aws_region: ${{ secrets.AWS_CN_REGION }}
--- a/.github/workflows/upload_to_s3.yml
+++ b/.github/workflows/upload_to_s3.yml
@@ -8,7 +8,10 @@ on:
       version:
         type: string
         required: true
-      flavors_matrix:
+      flavor:
+        type: string
+        required: true
+      arch:
         type: string
         required: true
       run_id:
@@ -26,18 +29,15 @@ on:
 jobs:
   upload_to_s3:
     name: upload to S3
-    runs-on: upload
+    runs-on: "ubuntu-24.04"
     defaults:
       run:
         shell: bash
     env:
-      CNAME: ''
+      CNAME: ""
     permissions:
       id-token: write
     environment: oidc_aws_s3_upload
-    strategy:
-      fail-fast: false
-      matrix: ${{ fromJson(inputs.flavors_matrix) }}
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # pin@v4.1.1
         with:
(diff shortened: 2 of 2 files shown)


NotTheEvilOne:
Could you share some more insights why a different VM in the same host environment (Azure for GitHub) improves the success rate of uploads? Would it be an alternative solution to only use public GitHub runners for that purpose as we have improved and implemented a workflow retry logic lately?

Shegox:
The difference seems to be that there are generally two types of GitHub Action runners: 
- [free GitHub hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners) (e.g. label `ubuntu-24.04`)
- [custom GitHub hosted runners (paid)](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners) (e.g. label `upload`)

In theory they should behave similar, however it seems that the free runner according to @nkraetzschmar get blocked from time to time from uploading to AWS S3 in China. That behavior seems to be both sporadic and then it happens be persistent across retries on different free runners for a while. I assume that the free runner and paid runners use different IP subnets and the free subnet gets blocked from time to time by the China firewall.

Because of this Gardenlinux switched to using paid custom runners with the label `upload` for all S3 upload jobs (to normal S3 as well as China S3) and that works stable as well. This approach has the disadvantage that these are paid and currently consume the above mentioned 200k paid runner minutes every 10 days.

With this PR the upload is first tried using the free runner and if that fails is only then retried on the custom paid runner. Assuming that in most cases the free runners are sufficient, but having a stable fallback with the paid runners in case the free runners are currently blocked.

.github/workflows/publish_s3.yml
NotTheEvilOne, line 1:
```patch
diff --git a/.github/workflows/publish_s3.yml b/.github/workflows/publish_s3.yml
index ccb6d68e..227df494 100644
--- a/.github/workflows/publish_s3.yml
+++ b/.github/workflows/publish_s3.yml
@@ -160,6 +160,25 @@ jobs:
       aws_s3_bucket: ${{ secrets.AWS_S3_BUCKET }}
   upload_to_s3_cn:
     needs: [ copy_artifacts, prepare_certs, workflow_data ]
+    if: ${{ contains(fromJSON('["1", "2", "3"]'), github.attempt) }}
+    name: Upload to S3 (China)
+    permissions:
+      id-token: write
+    uses: ./.github/workflows/upload_to_s3.yml
+    with:
+      commit_id: ${{ needs.workflow_data.outputs.commit_id }}
+      version: ${{ needs.workflow_data.outputs.version }}
+      flavors_matrix: ${{ needs.workflow_data.outputs.flavors_matrix }}
+      run_id: ${{ github.run_id }}
+      runner: upload
+    secrets:
+      aws_region: ${{ secrets.AWS_CN_REGION }}
+      aws_role: ${{ secrets.AWS_CN_IAM_ROLE }}
+      aws_session: ${{ secrets.AWS_CN_OIDC_SESSION }}
+      aws_s3_bucket: ${{ secrets.AWS_CN_S3_BUCKET }}
+  upload_to_s3_cn_custom_runner:
+    needs: [ copy_artifacts, prepare_certs, workflow_data ]
+    if: ${{ contains(fromJSON('["4", "5"]'), github.attempt) }}
     name: Upload to S3 (China)
     permissions:
       id-token: write
```

.github/workflows/upload_to_s3.yml
NotTheEvilOne, line 1:
```patch
diff --git a/.github/workflows/upload_to_s3.yml b/.github/workflows/upload_to_s3.yml
index 90171dc7..3656ca23 100644
--- a/.github/workflows/upload_to_s3.yml
+++ b/.github/workflows/upload_to_s3.yml
@@ -14,6 +14,9 @@ on:
       run_id:
         type: string
         required: true
+      runner:
+        type: string
+        default: ubuntu-24.04
     secrets:
       aws_role:
         required: true
@@ -26,7 +29,7 @@ on:
 jobs:
   upload_to_s3:
     name: upload to S3
-    runs-on: upload
+    runs-on: ${{ inputs.runner }}
     defaults:
       run:
         shell: bash
```

NotTheEvilOne (changes requested):
> Because of this Gardenlinux switched to using paid custom runners with the label `upload` for all S3 upload jobs (to normal S3 as well as China S3) and that works stable as well. This approach has the disadvantage that these are paid and currently consume the above mentioned 200k paid runner minutes every 10 days.

Thanks for the clarification. While I would (personally) not see the custom runners as being more stable (they only get the job done currently because if they get blocked they are retried and that seems to be sufficient) I see the idea of adding a fallback. I think we could get the best of both worlds by improving the `publish_s3.yml` workflow in a slightly different way:

Instead of copying the underlying GitHub job code you could add logic to give the requested `runner` as an input parameter and use `github.attempt` to "calculate" (numbers are unfortunately not supported for conditions) when to switch from public to custom runners.

NotTheEvilOne:
> If that is sufficient that would be a minimal change and as well accomplish what my main goal is (not paying for the runners :D)

Agreed. We can get back to the suggested code change variant afterwards if issues arise with the GitHub hosted runners.

Shegox:
Created #2962 to switch to the GitHub hosted runners. I will close this PR and we can come back to it in case switching to the free GitHub hosted runners causes problems

PR #2962, opened 2025-04-11
switch to free github-hosted runners for S3 upload

Shegox:
This PR adjust the behavior of the upload to S3 GitHub action to leverage GitHub hosted runners

Based on the discussion in #2961 using the (free) GitHub hosted directly should work for upload to S3 for both normal AWS and AWS China. If this change should not be compatible we can try again with a more suffisticated fallback solution as discussed in #2961. 

Switching to the free GitHub hosted runners reduces the amount of paid GitHub Action minutes by roughly 200k minutes / 10 days (https://github.com/gardenlinux/gardenlinux/actions/metrics/usage). 

n/a

**Definition of Done:**

**Release note**:

```other developer
Upload artifacts to S3 using free GitHub hosted runners
```

### DIFF
--- a/.github/workflows/upload_to_s3.yml
+++ b/.github/workflows/upload_to_s3.yml
@@ -26,7 +26,7 @@ on:
 jobs:
   upload_to_s3:
     name: upload to S3
-    runs-on: upload
+    runs-on: ubuntu-24.04
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


NotTheEvilOne (approved):
LGTM

PR #2972, opened 2025-04-14
tutorial: how to run tests locally

Vincinator:
We need a tutorial for new users on how to run tests locally. This is an addition to the tests/README.md. This tutorial is aimed for beginners who want to learn. 

This is written as a lesson, rather than a checklist.

### DIFF
--- a/docs/tutorials/tutorial_run_tests_locally.md
+++ b/docs/tutorials/tutorial_run_tests_locally.md
@@ -3,7 +3,7 @@
 You want to test a specific Garden Linux flavour locally? This tutorial is for you. 
 We will walk through the local test process, and assume you are already familiar with building Garden Linux images.
 
-TODO: insert link to tutorial to build images here
+TODO: Write tutorial for building images and insert link here.
 
 First, we need to know the flavour of Garden Linux that we want to test. (for example `kvm-gardener_prod-amd64`) 
 In a later bonus lesson we will test a custom image not defined in flavour.yaml (yet).  
@@ -175,7 +175,8 @@ We have the full name of the Garden Linux image, for example `kvm-gardener_prod-
 
 But the make targets list a lot more than just the flavours? which one to pick? This is easy once you understand the different methods to run the Garden Linux VM we want to test. 
 
-TODO: insert link to tests/README.md
+[Go to test README for a technical reference](../../tests/README.md)
+
 
 Since we want to run tests locally, we focus here on the two local options:
 
--- a/docs/tutorials/tutorial_run_tests_locally.md
+++ b/docs/tutorials/tutorial_run_tests_locally.md
@@ -5,8 +5,8 @@ We will walk through the local test process, and assume you are already familiar
 
 TODO: insert link to tutorial to build images here
 
-First, we need to know the flavour of Garden Linux that we want to test.
-We will use `kvm-gardener_prod` flavour in this tutorial, but feel free to divert here and play with other flavours. In a later bonus lesson we will test a custom image not defined in flavour.yaml (yet).  
+First, we need to know the flavour of Garden Linux that we want to test. (for example `kvm-gardener_prod-amd64`) 
+In a later bonus lesson we will test a custom image not defined in flavour.yaml (yet).  
 
 Typically, when you have the task to test something locally you already know the target, but if you need to pick a target, you can get a list of all officially supported test targets by running
 
@@ -171,7 +171,7 @@ which will read the flavours.yaml and parse it to generate the list of Make targ
 
 Great. You see a lot of test potential test targets now. You know the name of the image you want to test(we continue with `kvm-gardener_prod`), and you know the architecture of the image you want to test (we continue with `amd64`).
 
-We have the full name of the Garden Linux image: `kvm-gardener_prod-amd64`
+We have the full name of the Garden Linux image, for example `kvm-gardener_prod-amd64`
 
 But the make targets list a lot more than just the flavours? which one to pick? This is easy once you understand the different methods to run the Garden Linux VM we want to test. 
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/tutorial_run_tests_locally.md
```diff
@@ -0,0 +1,248 @@
...
+
+You want to test a specific Garden Linux flavour locally? This tutorial is for you. 
+We will walk through the local test process, and assume you are already familiar with building Garden Linux images.
+
+TODO: Write tutorial for building images and insert link here.
+
+First, we need to know the flavour of Garden Linux that we want to test. (for example `kvm-gardener_prod-amd64`) 
+In a later bonus lesson we will test a custom image not defined in flavour.yaml (yet).  
+
+Typically, when you have the task to test something locally you already know the target, but if you need to pick a target, you can get a list of all officially supported test targets by running
```
Akendo, line 11: You should add the set-up process as well. Remember a person reading this has never done this before.

docs/tutorials/tutorial_run_tests_locally.md
```diff
@@ -0,0 +1,248 @@
+# Tutorial: Run tests locally 
+
+You want to test a specific Garden Linux flavour locally? This tutorial is for you. 
+We will walk through the local test process, and assume you are already familiar with building Garden Linux images.
+
+TODO: Write tutorial for building images and insert link here.
+
+First, we need to know the flavour of Garden Linux that we want to test. (for example `kvm-gardener_prod-amd64`)
```
Akendo, line 8: When you never run gl, you might guess what a flavor is. However, I would think it would be something like Ubuntu has. But you need to teach the reader about flavors first.

docs/tutorials/tutorial_run_tests_locally.md
```diff
@@ -0,0 +1,248 @@
...
+In a later bonus lesson we will test a custom image not defined in flavour.yaml (yet).  
+
+Typically, when you have the task to test something locally you already know the target, but if you need to pick a target, you can get a list of all officially supported test targets by running
+
+```bash
+make --directory=tests/
+```
+which will read the flavours.yaml and parse it to generate the list of Make targets you can execute. Try it!
+
+<details>
```
Akendo, line 18: Love the usage of detail here! <3

docs/tutorials/tutorial_run_tests_locally.md
```diff
@@ -0,0 +1,248 @@
+# Tutorial: Run tests locally 
+
+You want to test a specific Garden Linux flavour locally? This tutorial is for you. 
+We will walk through the local test process, and assume you are already familiar with building Garden Linux images.
+
+TODO: Write tutorial for building images and insert link here.
```
Akendo, line 6: I think you should start with an introduction, trying to explain a bit how our build system is intended to. Cloud-Init has some [tutorials](https://cloudinit.readthedocs.io/en/latest/tutorial/qemu.html) as well. They are not the best, but I think might be a good frame of reference.

docs/tutorials/tutorial_run_tests_locally.md
Vincinator: What would be a good scope for a single tutorial? 
As a target, I would like to have multiple tutorials: 
- build 
- test
- develop feature
- develop builder
- develop package

maybe a 
- getting started tutorial

would be a good place

docs/tutorials/tutorial_run_tests_locally.md
Vincinator: good point, will add this.

docs/tutorials/tutorial_run_tests_locally.md
Vincinator: on the other hand repeating does not hurt. We can be brief or link to a setup. but agree that we need the info 👍🏻

fwilhe: I would agree with @Vincinator on the point that tutorials should be small and focussed in scope. Not sure if our existing docs on how to build images are good enough, I'm open to the idea of also having a tutorial for that, but I'd prefer to keep those focussed on that topic and cross-reference prerequisites as needed/suitable.

docs/tutorials/tutorial_run_tests_locally.md
```diff
@@ -0,0 +1,257 @@
...
+In a later bonus lesson we will test a custom image not defined in [flavour.yaml](../../flavors.yaml) (yet).  
+
+Typically, when you have the task to test something locally you already know the target, but if you need to pick a target, you can get a list of all officially supported test targets by running
+
+```bash
+make --directory=tests/
+```
+which will read the flavours.yaml and parse it to generate the list of Make targets you can execute. Try it!
+
+<details>
```
fwilhe, line 27: just a suggestion, I think this listing could be abbreviated with dots, I think the reader gets the point. it is hidden by default, so this is not too big of a deal, but I think that would make it easier to read. keep some example lines that are important for the later text, and replace the oters with a line like

".. (abbreviated for readability)"

docs/tutorials/tutorial_run_tests_locally.md
```diff
@@ -0,0 +1,257 @@
...
+
+
+Great. You see a lot of potential test targets now. You know the name of the image you want to test (you can continue with `kvm-gardener_prod`), and you know the architecture of the image you want to test (you can continue with your local CPU arch, as long as it is arm64 or amd64).
+
+We have the full name of the Garden Linux image, for example `kvm-gardener_prod-amd64`
+
+But the make targets list a lot more than just the flavours, which one to pick? 
+This is easy once you understand the different methods to run the Garden Linux VM we want to test. 
+
+Since we want to run tests locally, we focus here on the two local options:
```
fwilhe, line 188: maybe here it would make sense to quickly elaborate on what the alternative to running test 'locally' means.. this might not be obvious to the reader

Issue #2964, opened 2025-04-11
Create new GL feature that is compatible with lima


pnpavlov:
Create a new feature in GL that enables the requirements for running a Lima VM 

- [ ] Create a new feature that includes cloudinit
- [ ] Publish the feature as qcow2
- [ ] Test the system can boot with sufficient disk space for basic evaluation (e.g. 1-2 small packages at least ~100mb).

fwilhe:
done with https://github.com/gardenlinux/gardenlinux/pull/2980

PR #2980, opened 2025-04-24
Allow building images suitable for use with lima-vm

fwilhe:
This PR adds a `lima` feature that enables the resulting image to be usable with https://lima-vm.io

This is useful for a few reasons:
- lima takes care of creating a user, generating an ssh key and some other aspects that Garden Linux does not have its own tooling for
- This allows developers and interested externals to create Garden Linux virtual machines for testing, evaluation and similar purposes

Part of https://github.com/gardenlinux/gardenlinux/issues/2964


fwilhe:
Few observations, known issues:

- [ ] It looks like mounts don't work despite we have the sshfs package available, reason is not clear
- [ ] We can't boot using `vz`, not sure why that is, it looks like the vm is stuck at the efi bootloader

fwilhe:
ci failed for unrelated reasons

Vincinator (approved):

PR #3010, opened 2025-05-06
Clean up GitHub workflow actions to authenticate to cloud providers

NotTheEvilOne:
This PR resorts the workflow actions to authenticate to cloud providers for easier readability


Akendo (changes requested):
Please break this change into a smaller PR's. Each change to the respected Hypervisor authentication. This way we can see better in case something breaks. Also, It's difficult to review since I feel like you mix the different Hyperscaler provides.

NotTheEvilOne:
Unfortunately this (separating to multiple PRs) won't work. This PR mainly resorts the existing steps by alphabet enhancing readability to find the authentication of a specific cloud provider more easily. More or less all logic is the same as before.

Akendo (approved):
LGTM.

PR #3012, opened 2025-05-09
Separate certs into two channels: stable and nightly

nkraetzschmar:

- Separate certs into two channels: *stable* and *nightly*
- Adjust cert lifetime to be 10 years for all prod certs, both stable and nightly, and 1 year for local dev certs (without `gardenlinux-*-` prefix)
- Re-generate all certs with new configs
- For now make everything default to nightly certs before we introduce cert channel selection in subsequent PR


.github/workflows/build_flavor.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           fail-on-cache-miss: true
       - name: Write secureboot db arn for kms backed certificates
         if: ${{ inputs.use_kms }}
-        run: echo "${{ secrets.secureboot_db_kms_arn }}" > cert/gardenlinux-secureboot.db.arn
+        run: echo "${{ secrets.secureboot_db_kms_arn }}" > cert/gardenlinux-nightly-secureboot.db.arn
```
NotTheEvilOne, line 67: This step is at the wrong place I think. It should be part of `.github/workflows/build_requirements.yml`. As this is not the case as of today we might:
- change `build_requirements.yml` to accept the required secret as optional.
- at least rename the file created here to `secureboot.db.arn`. In `build_requirements.yml` all files named `gardenlinux-nightly-*` are already linked and cached for later use e.g. in this step.

features/_usi/exec.post
```diff
@@ -6,7 +6,7 @@ rootfs="$1"
 
 mkdir -p "$rootfs/etc/gardenlinux"
 
-openssl x509 -in /builder/cert/gardenlinux-oci-sign.crt -pubkey -noout > "$rootfs/etc/gardenlinux/oci_signing_key.pem"
+openssl x509 -in /builder/cert/gardenlinux-nightly-oci-sign.crt -pubkey -noout > "$rootfs/etc/gardenlinux/oci_signing_key.pem"
```
NotTheEvilOne, line 9: "make build" is called at least on workflows after the certificate files relevant are already renamed to not contain the prefix `gardenlinux-nightly-`. For future extensibility this feature should account a file just named `oci-sign.crt`. If the file might not exist:
- we should either check for the file and fail with missing prerequisites
- or add the code line above if and only if the file `oci-sign.crt` does not exist.

.github/workflows/publish_s3.yml
```diff
@@ -133,7 +133,7 @@ jobs:
 
           # USE_KMS in build_requirements.yml
           for f in secureboot.{{pk,kek,db}.{crt,der,auth},aws-efivars}; do
-            cp "cert/gardenlinux-$f" "$CNAME/$CNAME.$f"
+            cp "cert/gardenlinux-nightly-$f" "$CNAME/$CNAME.$f"
```
NotTheEvilOne, line 136: This construction is currently a workaround as cached files are not accessible by subsequent workflows (which is based on the naming of the cache not really true but you can't guarantee that the cache still exists). If certificates are created in a single source of true way (in `.github/workflows/build_requirements.yml`) these files should be stored in an artifact and then we do not need these lines anymore.

NotTheEvilOne (commented):
Wasn't the idea to use `USE_KMS` as an enum containing `nightly` or `release` as it's value instead of being a `boolean`?

nkraetzschmar:
> Wasn't the idea to use `USE_KMS` as an enum containing `nightly` or `release` as it's value instead of being a `boolean`?

yes, but that will be step 2. for now this PR is **only** to actually rotate the certs and introduce the two cert channels, with the explicit intention that this just falls back to using nightly for now till the proper select cert mechanism is there.
Idea is to keep each PR small and reviewable, instead of one massive PR changing workflows and cert setup.

Akendo (approved):
For testing this is great!

Issue #2754, opened 2025-02-20
Github Actions - OCI artifact tagging


yeoldegrove:
**What would you like to be added**:

In #2711 we agreed to work on this. This is to track detailed work.

 - generate tags in python-gardenlinux-lib?
 - tags: to be discussed

* A clear strategy on how to call releases, images and tags has to be documented and implemented.
  * Make sure who is consuming our current releases and avoid breaking changes to the naming scheme.
  * alligned with: https://github.com/gardenlinux/gardenlinux/blob/main/docs/00_introduction/release.md
      * stable (virtual release)
      * patch
      * nightly
      * dev (to be discussed)
      * next (virtual release)
  * discuss/decide pre-/postfixes
      * would break or current versioning paradigm
      * would make releases much more distinguishable
  * discuss/decide tags on images
      * put `$commitShort` in a tags
      * a tag `1782.0-aws-gardener_prod_tpm2_trustedboot-arm64-1782.0-acf081e0-arm64` is no good
          * nightly tags
              * manifest `1782.0` + `nightly`
              * image `1782.0-aws-gardener_prod_tpm2_trustedboot-arm64`
          * example dev tags
              * manifest `1782.0-acf081e0`
              * image `1782.0-acf081e0-aws-gardener_prod_tpm2_trustedboot-arm64`
          * example release tags
              * manifest `1782.1`, `1782.1-release`, `latest`, `1782-latest`, `1782-release`
              * image `1782.1-aws-gardener_prod_tpm2_trustedboot-arm64`
          * in general `$version` including `release` or `$commitShort` (for dev) + `$version-$flavor-$arch` would be much clearer
  * discuss/decide if S3 should be handled similar to nightly tags

Vincinator:
Thanks! 👍 

Just a nit pick on the usage of the term "manifest" below: 
> a tag 1782.0-aws-gardener_prod_tpm2_trustedboot-arm64-1782.0-acf081e0-arm64 is no good
nightly tags
manifest 1782.0 + nightly
image 1782.0-aws-gardener_prod_tpm2_trustedboot-arm64

- oci index (media type `application/vnd.oci.image.index.v1+json`): this is the entry point, and should be tagged with version (+ optional suffixes according to your suggestions).
- oci manifest(s) (media type `application/vnd.oci.image.manifest.v1+json`): this should have some tag that can be directly referenced. The suggested rename is fine `1782.0-aws-gardener_prod_tpm2_trustedboot-arm64`

[...]

PR #3016, opened 2025-05-12
feat: add support for release and nightly channels to publishing

nkraetzschmar:

- Completes the work started in #3012
- Allows distinction between release and nightly channels when building and publishing

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,10 +9,12 @@ on:
         description: 'Garden Linux version'
         type: string
         default: today
-      use_kms:
-        description: 'Use Garden Linux KMS'
-        type: boolean
-        default: false
+      signing_env:
+        type: string
+        default: ""
+      channel:
+        type: string
+        default: nightly
       flavors_parse_params:
         description: 'Run bin/parse_flavors.py with these parameters'
         default: '--exclude "bare-*" --no-arch --json-by-arch --build --test'
@@ -54,7 +56,8 @@ jobs:
     uses: ./.github/workflows/build_requirements.yml
     with:
       version: ${{ inputs.version == '' && 'now' || inputs.version }}
-      use_kms: ${{ inputs.use_kms }}
+      use_kms: ${{ inputs.signing_env != '' }}
+      channel: ${{ inputs.channel }}
   bootstrap:
     needs: requirements
     name: Bootstrap build
@@ -87,8 +90,8 @@ jobs:
       flavor: ${{ matrix.flavor }}
       commit_id: ${{ needs.requirements.outputs.commit_id }}
       version: ${{ needs.requirements.outputs.version }}
-      # GitHub workflow workaround for bool
-      use_kms: ${{ needs.requirements.outputs.use_kms == 'true' }}
+      signing_env: ${{ inputs.signing_env }}
+      channel: ${{ inputs.channel }}
     uses: ./.github/workflows/build_flavor.yml
     secrets: inherit
   kmodbuild_container:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,12 +9,9 @@ on:
         description: 'Garden Linux version'
         type: string
         default: today
-      signing_env:
+      target:
         type: string
-        default: ""
-      channel:
-        type: string
-        default: nightly
+        default: dev
       flavors_parse_params:
         description: 'Run bin/parse_flavors.py with these parameters'
         default: '--exclude "bare-*" --no-arch --json-by-arch --build --test'
@@ -56,8 +53,7 @@ jobs:
     uses: ./.github/workflows/build_requirements.yml
     with:
       version: ${{ inputs.version == '' && 'now' || inputs.version }}
-      use_kms: ${{ inputs.signing_env != '' }}
-      channel: ${{ inputs.channel }}
+      target: ${{ inputs.target }}
   bootstrap:
     needs: requirements
     name: Bootstrap build
@@ -90,8 +86,7 @@ jobs:
       flavor: ${{ matrix.flavor }}
       commit_id: ${{ needs.requirements.outputs.commit_id }}
       version: ${{ needs.requirements.outputs.version }}
-      signing_env: ${{ inputs.signing_env }}
-      channel: ${{ inputs.channel }}
+      signing_env: ${{ needs.requirements.outputs.signing_env }}
     uses: ./.github/workflows/build_flavor.yml
     secrets: inherit
   kmodbuild_container:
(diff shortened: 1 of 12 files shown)


.github/workflows/build_flavor.yml
```diff
@@ -17,9 +17,12 @@ on:
       version:
         type: string
         required: true
-      use_kms:
-        type: boolean
-        default: false
+      signing_env:
+        type: string
+        default: ''
+      channel:
```
Akendo, line 23: We should set the nightly to default and only have release as an extra option?

.github/workflows/build_flavor.yml
```diff
@@ -40,8 +43,8 @@ jobs:
       id-token: write
     env:
       CNAME: ''
-      USE_KMS: ${{ inputs.use_kms }}
-    environment: ${{ inputs.use_kms && 'oidc_aws_kms_secureboot_signing' || '' }}
+      USE_KMS: ${{ inputs.signing_env != '' && 'true' || 'false' }}
```
Akendo, line 46: This seems a bit confusing to me.

yeoldegrove:
Could we use the same naming schemes as in https://github.com/gardenlinux/glrd?tab=readme-ov-file#release-schema and https://github.com/gardenlinux/gardenlinux/blob/main/docs/00_introduction/release.md, so calling this "release types" instead of "channels"?
I could imagine this gets confusing else wise, having "nightly release" and a "nightly channel" and distinguish between them.
Also it would be a  "patch" release, not a "stable" release, as "stable" releases are somewhat virtual.

.github/workflows/build.yml
```diff
@@ -9,10 +9,12 @@ on:
         description: 'Garden Linux version'
         type: string
         default: today
-      use_kms:
-        description: 'Use Garden Linux KMS'
-        type: boolean
-        default: false
+      signing_env:
```
NotTheEvilOne, line 12: I personally would love to merge `signing_env`and `channel` to a more generic `target` here (as "build target") should match the intended functionality pretty well. From a development perspective `environment` is also very common for both build and release publishing.

In both cases I think it's error prone to add two new variables for the intention to:
- switch between different certificate files / sources
- change possible tags and repositories later in release publishing processes

.github/workflows/build.yml
```diff
@@ -87,8 +90,8 @@ jobs:
       flavor: ${{ matrix.flavor }}
       commit_id: ${{ needs.requirements.outputs.commit_id }}
       version: ${{ needs.requirements.outputs.version }}
-      # GitHub workflow workaround for bool
-      use_kms: ${{ needs.requirements.outputs.use_kms == 'true' }}
+      signing_env: ${{ inputs.signing_env }}
```
NotTheEvilOne, line 93: This logic should move to `build_requirements.yml` as suggested in #3012. Therefore no additional information about KMS is required here I think. If the builder requires `USE_KMS` as environment we should leave the input as it is and base it on the value of `target` (or `environment` as stated above).

.github/workflows/build_flavor.yml
nkraetzschmar: there is also the dev.yml using build (run for every commit on any branch), which uses per build created local certs, so without any kms keys. we should keep support for this to e.g. allow running at least the build and chroot tests on PR from external contributors.

NotTheEvilOne: Should be removed or still named `USE_KMS` in case builder requires it. See feedback for `build.yml`.

.github/workflows/build_flavor.yml
```diff
@@ -63,10 +66,10 @@ jobs:
           key: cert-${{ github.run_id }}
           fail-on-cache-miss: true
       - name: Write secureboot db arn for kms backed certificates
-        if: ${{ inputs.use_kms }}
-        run: echo "${{ secrets.secureboot_db_kms_arn }}" > cert/gardenlinux-nightly-secureboot.db.arn
+        if: ${{ inputs.signing_env != '' }}
+        run: echo "${{ secrets.secureboot_db_kms_arn }}" > cert/gardenlinux-${{ inputs.channel }}-secureboot.db.arn
```
NotTheEvilOne, line 70: This should be moved to `build_requirements.yml`.

.github/workflows/build_flavor.yml
nkraetzschmar: basically the USE_KMS expects to be a boolean, now we moved to `signing_env` which may be empty to indicate no kms or some value to use kms. this just maps is not empty to true or false.

.github/workflows/build_requirements.yml
NotTheEvilOne, line 1: The whole creation of required certificates based on the intended `target` should happen here. Furthermore we should change the cert cache to an artifact for reusability in `publish_*.yml`.

.github/workflows/publish.yml
```diff
@@ -16,8 +16,24 @@ jobs:
     uses: ./.github/workflows/download_workflow_data.yml
     with:
       run_id: ${{ github.event_name == 'workflow_dispatch' && inputs.run_id || github.event.workflow_run.id }}
+  determine_repository:
```
NotTheEvilOne, line 19: I'm not sure if I like this job. Pushing images to gardenlinux/gardenlinux even if they are nightlies should work as long as we properly tag them as suggested in #2754. In case we need this side step I would love to move it to `publish.yml` as we reference / create repository URLs there anyway.

nkraetzschmar:
hmmmmm 🤔 not sure..... the issue is this really does not distinguish between stable, long term, patch, etc. it's just a matter of is this something to be truly released with release keys or something that is build automatically with less privileged certs.

so really there is no concept of anything other then release and nightly here

.github/workflows/publish_s3.yml
```diff
@@ -132,8 +137,8 @@ jobs:
           rm "$CNAME.tar.gz"
 
           # USE_KMS in build_requirements.yml
-          for f in secureboot.{{pk,kek,db}.{crt,der,auth},aws-efivars}; do
-            cp "cert/gardenlinux-nightly-$f" "$CNAME/$CNAME.$f"
+          for f in secureboot.{{pk,kek,db}.{crt,der,auth},aws-efivars} oci-sign.crt; do
```
NotTheEvilOne, line 140: This should be removed and resolved with changes in `build_requirements.yml`.

NotTheEvilOne (changes requested):

.github/workflows/publish.yml
nkraetzschmar: eventually we'll want to go for tagging, but that may have unintended consequences where various tools expect tags, or parts of tags, to be properly formed versions, so introducing e.g. a `~nightly` suffix after the version might cause unexpected side effects. this should be postponed till after the major release.

PR #3024, opened 2025-05-14
Feature/cert/update production keys

Akendo:
In this PR, we add the keys that are registered with the necessary production KMS.


nkraetzschmar (approved):
lgtm

PR #3027, opened 2025-05-14
Add reusable code for GitHub environment selection

NotTheEvilOne:
This PR adds reusable code for GitHub environment selection to `.github/workflows/github.mjs`.

### DIFF
--- a/.github/workflows/build_requirements.yml
+++ b/.github/workflows/build_requirements.yml
@@ -33,16 +33,16 @@ jobs:
     outputs:
       environment: ${{ steps.signing_environment.outputs.environment }}
     steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # pin@v4.1.1
+        with:
+          submodules: true
       - name: Set signing environment
         id: signing_environment
-        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/build_requirements.yml
```diff
@@ -33,16 +33,16 @@ jobs:
...
         id: signing_environment
-        run: |
-          if [[ "${{ inputs.target }}" == "release" ]]; then
-            echo "environment=oidc_aws_kms_release" | tee -a "$GITHUB_OUTPUT"
-          elif [[ "${{ inputs.target }}" == "nightly" ]]; then
-            echo "environment=oidc_aws_kms_nightly" | tee -a "$GITHUB_OUTPUT"
-          else
-            echo "environment=" | tee -a "$GITHUB_OUTPUT"
-          fi
+        uses: actions/github-script@v7
```
nkraetzschmar, line 41: this should probably be a pinned version

NotTheEvilOne: As we have the `v7` use in multiple places I'll address this in a separate PR. Thanks :)

PR #3043, opened 2025-05-21
fix: multipath ignore everything except iscsi and nvme/tcp

nkraetzschmar:

Fixes issue on gardener, where `multipathd` accessing regular SCSI devices is interfering with `mke2fs` during persistent volume creation.

### DIFF
--- a/features/multipath/file.include/etc/multipath.conf
+++ b/features/multipath/file.include/etc/multipath.conf
@@ -2,3 +2,12 @@ defaults {
     user_friendly_names yes
     find_multipaths no
 }
+
+blacklist {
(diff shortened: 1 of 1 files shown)


Akendo:
Can you explain the fix a bit more here?

Gchbg (approved):

nkraetzschmar:
> Can you explain the fix a bit more here?

This fix configures `multipathd` to ignore *all* devices except those that are connected via iSCSI or nvme over TCP. That way it does not interfere with anything trying to access such devices directly, i.e. not through a multipath device mapper.

PR #3062, opened 2025-05-30
fix: make image_variants optional in release page action

nkraetzschmar:
should fix failing workflow run for rel-1592:
https://github.com/gardenlinux/gardenlinux/actions/runs/15341910973/job/43169784446

by allowing to run it with the new image_variants arg set to an empty string.


Vincinator (approved):
lgtm 

Empty string is extended by input, therefore previous behaviour with default values from input still works.

PR #3100, opened 2025-06-11
add parent-workflow.json artifact to publish jobs

yeoldegrove:

This makes it possible to identify which "parent workflow" started the publishing jobs. This will be useful for historic analyses e.g. in the dashboard.


.github/workflows/publish.yml
NotTheEvilOne, line 1: While this change would work I would like to recommend to reuse the existing workflow-data artifact (see `download_workflow_data.yml`). This would allow you to use data added later on for `manual_release.yml` and `nightly.yml` for the dashboard as well. To "copy" it just have a look at `copy_artifacts` step of `publish_s3.yml`.

The steps would be:
1. `workflow_data`
2. (in parallel to existing steps) `copy_workflow_data` to the current workflow run

There is a chance to get a workflow execution without the `copy_workflow_data` step to be completed so please be aware that the logic expecting this artifact may encounter a 404.

NotTheEvilOne (changes requested):

.github/workflows/publish.yml
```diff
@@ -16,6 +16,22 @@ jobs:
     uses: ./.github/workflows/download_workflow_data.yml
     with:
       run_id: ${{ github.event_name == 'workflow_dispatch' && inputs.run_id || github.event.workflow_run.id }}
+  workflow_data_artifact:
+    name: Publish parent workflow-data artifact
+    needs: [ workflow_data ]
```
NotTheEvilOne, line 21: Note: For a single entry you may use a string instead of an array here. See https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

.github/workflows/publish_s3.yml
```diff
@@ -16,6 +16,22 @@ jobs:
     uses: ./.github/workflows/download_workflow_data.yml
     with:
       run_id: ${{ github.event_name == 'workflow_dispatch' && inputs.run_id || github.event.workflow_run.id }}
+  workflow_data_artifact:
+    name: Publish parent workflow-data artifact
+    needs: [ workflow_data ]
```
NotTheEvilOne, line 21: Same note about the array :)

NotTheEvilOne (approved):
LGTM

Issue #2965, opened 2025-04-11
Publish images for lima consumption


pnpavlov:
As newcomer to GL I'd like to have an easy to start with pre-build file that I can run locally or in the cloud.

**Tasks:**
- [ ] Automate an image build including the features required and file format usable bi lima
- [ ] Prepare easy to use URL on gardenlinux.com or another easy to follow source location

saubhikdattagithub:
As a POC, the tasks are appeared to be working in a test GCP environment:

**Requisites:**
- GCP Account (Project ID) / GCP Storage Bucket
- GCP Storage Bucket Key exported as json with "Storage Admin" permission
- Add credentials in Github Actions "Repository Secret"

**GH Actions file to:**
- Authenticate GCP
- Build Image with feature: lima
- Fetch the generated QCOW2 image file
- Upload the QCOW2 the GCP Storage Bucket using "gsutil"
- Copy the randomly generated QCOW2 by build to static named upload for persistent download (garden-lima-today.qcow2)
- Generate a static Public Accessible URL to fetch the image to run lima VM 
   - (https://storage.googleapis.com/lima-images-qcow2/garden-lima-today.qcow2)
- Workflow file: https://github.com/saubhikdattagithub/gardenlinuxfork/actions/runs/16080958380/workflow 

### lima creation using GardenLinux image created by above GH Actions
```
limactl create --name=gardenlinux-today https://storage.googleapis.c
om/lima-images-qcow2/garden-lima-today.qcow2
```

<img width="1473" height="428" alt="Image" src="https://github.com/user-attachments/assets/c5818937-d960-40cd-9c17-79f68edd546e" />

`limactl start gardenlinux-today`

<img width="1471" height="837" alt="Image" src="https://github.com/user-attachments/assets/9d3495ad-851d-44db-a4a3-ebe45ac52bbb" />

`limactl shell gardenlinux-today`

<img width="1298" height="142" alt="Image" src="https://github.com/user-attachments/assets/a0b86ccc-702a-4b8b-996f-f530cd767f58" />

### Summary:
Automate an image build including the features required and file format usable bi lima 

-   --> https://github.com/saubhikdattagithub/gardenlinuxfork/actions/runs/16080958380/workflow 

Prepare easy to use URL on gardenlinux.com or another easy to follow source location

-  --> https://storage.googleapis.com/lima-images-qcow2/garden-lima-today.qcow2

fwilhe:
[...]

PR #3118, opened 2025-06-19
Generate lima manifest automatically when building

fwilhe:
This is a convenience feature to make it more simple/easy to spin up lima vms based on custom built images. Having the yaml manifest with the correct file name available is helpful to create vm easily using the `limactl` tool.

This PR is related to https://github.com/gardenlinux/gardenlinux/issues/2965


saubhikdattagithub (approved):
approved as discussed

PR #3140, opened 2025-06-30
Add script to browse glvd data on the terminal with fuzzy search

fwilhe:
As discussed with @Vincinator it would be nice to have a `gl-search` like tool for glvd. This is a very first and rough draft for such a tool to browse vulnerabilities by Garden Linux versions.

TODO:

- [x] Add API to get available GL version from GLVD api to get rid of the hardcoded versions
- [ ] Add more modes, such as 'search by cve id', 'search by package name' etc

### DIFF
--- a/hack/glvd-search.sh
+++ b/hack/glvd-search.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+# Requirements: fzf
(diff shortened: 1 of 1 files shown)


fwilhe:
Thanks for the review @Vincinator 

It is not done or perfect yet but I think it is useful already, so going to merge now, can work on improvements as time permits

PR #3148, opened 2025-07-02
fix: release page action

### LINKED ISSUE #3063: re-design release page action
At the moment the release page action fails while having actually ran correctly due to some missing alignment between the flavours during build and for the release page. A similar issue is causing actual errors for creating a release page from rel-1592, currently fixed via workaround #3062.

We should eventually re-design the release page workflow to properly align with the `flavor.yml` based system used elsewhere.

Leon-hk:
- Reverts workaround from #3062 
- Scans `flavors.yaml` and checks if the selected flavor is available and published before downloading the manifest
- removes **Changes** section if there is no package upgrade, i.e. for new releases

Fixes #3063

The version has the same output when tested for the latest release with 
```bash
.github/workflows/release_note.py create --tag 1877.0 --commit "$(echo "ddd04088fe8ea9acadf95614af7d163d9f97a753")"
```

Testing patch 1592.10 with
```bash
.github/workflows/release_note.py create --tag 1592.10 --commit "$(echo "848e339adf8f49a89067b0a5d536c2f82d8fd450")"
```
where the error occurred had some slight changes, which are probably caused by some database updates in the meantime or the time of execution and have nothing to do with this fix.

<details>
<summary>Changes when testing with patch 1592.10</summary>

```diff
  - CVE-2025-37915
  - CVE-2025-37938
  - CVE-2025-23150
+ - CVE-2025-38005
  - CVE-2025-23151
  - CVE-2025-37824
  - CVE-2025-37773
...
  - CVE-2025-37879
  - CVE-2025-37829
  - CVE-2025-37995
+ - CVE-2025-38024
  - CVE-2025-37776
  - CVE-2025-37940
  - CVE-2025-37786
...
  - CVE-2025-37789
  - CVE-2025-37884
  - CVE-2024-49569
+ - CVE-2025-38008
  - CVE-2025-23163
  - CVE-2025-23149
  - CVE-2025-39989
...
  - CVE-2025-23146
  - CVE-2025-37757
  - CVE-2025-37892
+ - CVE-2025-38014
  - CVE-2025-37989
+ - CVE-2025-38020
  - CVE-2025-37804
  - CVE-2025-37923
  - CVE-2025-37778
+ - CVE-2025-38027
+ - CVE-2025-38023
  - CVE-2025-37836
  - CVE-2023-53130
  - CVE-2025-37933
...
  - CVE-2025-37992
  - CVE-2025-37954
  - CVE-2025-37800
+ - CVE-2025-38018
  - CVE-2024-46733
  - CVE-2025-37901
  - CVE-2025-37936
...
  - CVE-2025-37928
  - CVE-2025-37959
  - CVE-2025-37921
+ - CVE-2025-38015
  - CVE-2025-22120
  - CVE-2025-37819
  - CVE-2025-37947
...

  ## Changes in Package Versions Compared to 1592.9
  ```diff
- --- 1592.9	2025-05-30 10:37:45.255003770 +0000
+ --- 1592.9	2025-07-02 11:12:47
- +++ 1592.10	2025-05-30 10:37:45.315007567 +0000
+ +++ 1592.10	2025-07-02 11:12:47
  -bpftool 7.3.0+6.6.87-0gl0~bp1592
  -bpftool-dbgsym 7.3.0+6.6.87-0gl0~bp1592
  +bpftool 7.3.0+6.6.92-0gl0~bp1592
```
</details>


.github/workflows/release_note.py
```diff
@@ -35,7 +35,7 @@
 
 # https://github.com/gardenlinux/gardenlinux/issues/3044
 # Empty string is the 'legacy' variant with traditional root fs and still needed/supported
-image_variants = ['']
+image_variants = ['', '_usi', '_tpm2_trustedboot']
```
fwilhe, line 38: I think `image_variants` is not used anymore. Am I missing something?

The variable, the comment above and its usage can be deleted if I'm not mistaken.

.github/workflows/release_note.py
```diff
@@ -147,6 +160,8 @@ def is_unsupported_ali_combination(platform, architecture, variant):
 def download_all_singles(version, commitish):
     if commitish == None:
         raise Exception("Commitish is not set")
+    available = parse_published_flavors(version, commitish)
```
fwilhe: maybe a nitpick but I'd prefer a more speaking name such as `available_flavors`

.github/workflows/release_note.py
Leon-hk: It is used in line 170 to loop through the different flavor variants. I tried to use all variations with `public: true` in `flavors.yaml` (e.g. the combinations generated by parse_published_flavors())but that would've created more variants than before.

.github/workflows/release_note.py
Leon-hk: good point, makes the code more readable

.github/workflows/release_note.py
```diff
@@ -137,6 +137,23 @@ def download_meta_single_manifest(bucket, bucket_path, image_name, dest_path):
     return f"{dest_path}/{image_name}"
 
 
+def parse_published_flavors(commitish):
+    response = requests.get(f"https://raw.githubusercontent.com/gardenlinux/gardenlinux/{commitish}/flavors.yaml")
```
Leon-hk: I just realized that reading `flavors.yaml` from the current checkout will just be the newest version of the file and not the version from the actual commit to release. I have therefore added this fetch to get the correct version.
Please correct me if I'm wrong or if there is a cleaner way to do this

fwilhe: I think we might run into rate limit issues with that as github got super aggressive about rate limits for unauthenticated api calls, this also includes git clone or file downloads from github action runners. Might work 9 times and fail the 10th time.

fwilhe: Maybe using actions/checkout is the more safe way? https://github.com/actions/checkout?tab=readme-ov-file#Fetch-only-a-single-file

I did not check if that is authenticated automatically, but I would assume so.

Leon-hk: This is definitely a way cleaner implementation, thank you for the tip!
I wasn't able to completely test the changes on the gardenlinux repo, but I made a similar checkout setup on a small repo and tested it there.

nkraetzschmar (approved):

Issue #3149, opened 2025-07-02
Fix nightly 2025-07-02


yeoldegrove:
**What happened:**

Our nightly fails with ssh not being able to connect to GCP instances:
https://github.com/gardenlinux/gardenlinux/actions/runs/16017335669

yeoldegrove:
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#metadata_startup_script-1 this feature is working as the sshd start is working:
- https://github.com/gardenlinux/gardenlinux/blob/d9625e7e1dfbeb69952dd9efee7e7f21235f25c6/tests/platformSetup/tofu/modules/gcp/main.tf#L161
- https://github.com/gardenlinux/gardenlinux/blob/d9625e7e1dfbeb69952dd9efee7e7f21235f25c6/tests/platformSetup/tofu/vm-startup-script.sh#L3

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#metadata-1 seems broken at the moment as keys are not passed to the instance (but are visible in the cloud console):
- https://github.com/gardenlinux/gardenlinux/blob/d9625e7e1dfbeb69952dd9efee7e7f21235f25c6/tests/platformSetup/tofu/modules/gcp/main.tf#L157-L159

This cannot be an error in the tf code as we did not change anything or bumped versions.
I suspect the google-guest-agent as we have issues with this one in our current release, too https://github.com/gardenlinux/gardenlinux/issues/2982.

Akendo:
The VM didn't went up with the fixed version. Need to check what went wrong.

PR #3150, opened 2025-07-02
Workaround broken google-guest-agent

yeoldegrove:

Workaround broken google-guest-agent. See issue for details and assumptions.

Fixes https://github.com/gardenlinux/gardenlinux/issues/3149


yeoldegrove:
Most likely the root cause is that the service is not enabled, so #3149 provides other possibilities for fixes.

PR #3172, opened 2025-07-09
build and publish images for use with lima-vm

saubhikdattagithub:
Build image, then upload it to GCP Storage bucket to create "LIMA" VMs directly by URL "https://images.gardenlinux.io/garden-lima-today.qcow2"

Fixes [#2965 ](https://github.com/gardenlinux/gardenlinux/issues/2965)

**Testings done**
 - POC performed first independently in own setup and working recorded in [here](https://github.com/gardenlinux/gardenlinux/issues/2965#issuecomment-3037296396)
 - Separate workload-identity-pool, service-account, policy-binding created for OIDC to not impact current setup
 - GH Actions repository secrets added with name LIMA (GCP_LIMA_WORKLOAD_IDENTITY_PROVIDER, GCP_LIMA_SERVICE_ACCOUNT, GCP_LIMA_BUCKET and LIMA_IMAGE_NAME) for use in this workflow
 - Authentication worked fine in all the runs, URL name corrected [build-job](https://github.com/gardenlinux/gardenlinux/actions/runs/16166277003/job/45628833524)


.github/workflows/lima_upload_oidc.yml
```diff
@@ -0,0 +1,49 @@
+name: LIMA Qcow2
```
fwilhe:
```suggestion
name: build and publish images for use with lima-vm
```

please take care to pick sensible names/lables that are descriptive of what is happening if possible

.github/workflows/lima_upload_oidc.yml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  push:
+    branches:
+      - lima-upload
```
fwilhe: this is only for development purposes, this could run on the main branch, but maybe that's a bit too much.. some day it would also be nice to publish release images, but for the time being I would suggest using a schedule to do this once per week for example

relevant docs:
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule

.github/workflows/lima_upload_oidc.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+    
+    # Step 1: Authenticate to GCP
```
fwilhe: I feel those comments don't add value, they only repeat what the name says, please remove them

.github/workflows/lima_upload_oidc.yml
```diff
@@ -0,0 +1,49 @@
...
+
+    ## Step 2: Set up gcloud SDK  
+    - name: Set up gcloud SDK
+      uses: google-github-actions/setup-gcloud@v1
+      with:
+        project_id: '${{ secrets.GCP_PROJECT }}'
+        install_components: 'gsutil'
+
+    # Step 3: Confirm authenticated identity
+    - name: Check myself once
```
fwilhe: is this useful? the log output does not look like it is

.github/workflows/lima_upload_oidc.yml
```diff
@@ -0,0 +1,49 @@
...
+      run: gcloud auth list
+
+    # Step 4: Upload QCOW2 after image creation
+    - name: Run QCOW2 creation script
```
fwilhe: please split build and upload into 2 steps

fwilhe: please build both amd and arm images and upload both of them

.github/workflows/lima_upload_oidc.yml
```diff
@@ -0,0 +1,49 @@
...
+      run: |
+        ./build kvm-lima-amd64
+        cd ./.build
+        IMAGE_QCOW2=$(find . -mmin -20 -maxdepth 1 -type f -name "*.qcow2" -exec basename {} \;)
```
fwilhe: the image name should be predictable (cname), why do we need find here?

.github/workflows/lima_upload_oidc.yml
```diff
@@ -0,0 +1,49 @@
...
+        gsutil cp "${IMAGE_QCOW2}" gs://${{ secrets.GCP_LIMA_BUCKET }}/${{ secrets.LIMA_IMAGE_NAME }}
```
fwilhe: why is the image name a secret? this is public information and storing that as a secret is not helpful imo

fwilhe: Also, for the name, please don't abbreviate with 'garden'.

The product is called Garden Linux, 'gardenlinux' as an identifier.

`gardenlinux-amd64-today.qcow2` for example, the same than also for arm64.

fwilhe: Also, we should adapt and merge this pr https://github.com/gardenlinux/gardenlinux/pull/3118 to generate the yaml manifest automatically. It should include the right url and be also uploaded to the bucket. This will allow much easier creation of lima-based vms with Garden Linux

fwilhe: Maybe we should just keep the image name as `$BUILDER_CNAME-$BUILDER_COMMIT.qcow2`, and give the manifest a generic name like `gardenlinux-lima.yaml`. Then the manifest can be updated with new image versions, and the user gets a stable url. The image contains the commit id from which this image was built which is good.

.github/workflows/lima_upload_oidc.yml
```diff
@@ -1,9 +1,8 @@
-name: LIMA Qcow2
+name: build and publish images for use with lima-vm
 
 on:
-  push:
-    branches:
-      - lima-upload
+  schedule:
+    - cron: '* * * * 2'
```
fwilhe: use something like `0 0 * * 2` which translates to “At 00:00 on Tuesday.”

cf
https://crontab.guru/#0_0_*_*_2

your expression translates to “At every minute on Tuesday.”, which is not what we want

fwilhe: also, for development purposes feel free to keep you branch trigger as before

.github/workflows/lima_upload_oidc.yml
saubhikdattagithub: @fwilhe Thanks for taking the time for detailed review. Pls review the updates and improvements as suggested. 
The image would now be uploaded for amd64 and arm64

.github/workflows/lima_upload_oidc.yml
```diff
@@ -31,19 +31,24 @@ jobs:
...
+    - name: Build image with feature:lima for amd64
+      run: |
+        ./build kvm-lima-amd64
+        cd ./.build
+        IMAGE_AMD=$(find . -mmin -20 -maxdepth 1 -type f -name "*amd*.qcow2" -exec basename {} \;)
+        gsutil cp "${IMAGE_AMD}" gs://${{ secrets.GCP_LIMA_BUCKET }}/gardenlinux-amd64-today.qcow2
+
+    - name: Build image with feature:lima for arm64
       run: |
         ./build kvm-lima-amd64
```
fwilhe: this says it builds for arm but actually builds for amd

fwilhe: oh, outdated

.github/workflows/lima_upload_oidc.yml
```diff
@@ -0,0 +1,54 @@
...
+        sudo update-binfmts --enable qemu-aarch64
+        ./build kvm-lima-arm64      
+        cd ./.build
+        IMAGE_ARM=$(find . -mmin -20 -maxdepth 1 -type f -name "*arm*.qcow2" -exec basename {} \;)
+        gsutil cp "${IMAGE_ARM}" gs://${{ secrets.GCP_LIMA_BUCKET }}/gardenlinux-arm64-today.qcow2
+ 
+    - name: Output public URL
+      run: | 
+        echo "amd64 image uploaded to https://images.gardenlinux.io/gardenlinux-amd64today.qcow2"
+        echo "arm64 image uploaded to https://images.gardenlinux.io/gardenlinux-arm64today.qcow2"
```
fwilhe:
```suggestion
        echo "amd64 image uploaded to https://images.gardenlinux.io/gardenlinux-amd64-today.qcow2"
        echo "arm64 image uploaded to https://images.gardenlinux.io/gardenlinux-arm64-today.qcow2"
```

.github/workflows/lima_upload_oidc.yml
```diff
@@ -0,0 +1,54 @@
+name: build and publish images for use with lima-vm
+
+on:
+  push:
+    branches:
+      - lima-upload  
+  schedule:
+    - cron: '0 11 * * 2'
```
fwilhe:
```suggestion
on:
  push:
    branches:
      - lima-upload  
  schedule:
    - cron: '0 11 * * 2'
  workflow_dispatch:
```

Add `workflow_dispatch` to allow manual rebuild/deployment of the images

yeoldegrove:
Is the `lima` this supposed to be "supported for development" in any kind of way? If yes, why not build/upload it via our usual mechanisms?

If we would put it into `flavors.yml` we would automatically get it built/uploaded/(published) and get the needed make targets to also do this locally.

Configuration of the lima yml file could even be added here https://github.com/gardenlinux/gardenlinux/blob/a2e3293a611e150362a83928b343749eff311c4e/tests/platformSetup/platformSetup.py#L30-L33.

fwilhe:
> If yes, why not build/upload it via our usual mechanisms?

I have not considered this so far. I'm not sure if acceptance for maintaining it is high enough in the team to integrate it into the normal publishing process. I am not against it, but also I don't see a very strong need to do it.

fwilhe:
@saubhikdattagithub are you still working on this?

Looks good mostly, what I see so far, before merge:
- We should remove the push trigger for your feature branch that was used for development 
- We should amend the README of the lima feature to document the new way a gl vm can be created, for example using this command `limactl create --name gardenlinux https://images.gardenlinux.io/gardenlinux.yaml`
  - No extra setup/download needed, if you have lima set up you can just use it, I think that's so cool, love it

saubhikdattagithub:
@fwilhe : Kindly review the updates

> * We should remove the push trigger for your feature branch that was used for development
-- Okay, removed workflow_dispatch from the actions workflow

> * We should amend the README of the lima feature to document the new way a gl vm can be created, for example using this command `limactl create --name gardenlinux https://images.gardenlinux.io/gardenlinux.yaml`
-- Okay, amended the new way of LIMA vm creation with GardenLinux

 - +restrictive OIDC for gcloud/attributes:  https://github.tools.sap/I550515/garden/blob/main/limaOIDCrestricted.md 

 - succesful test run: https://github.com/gardenlinux/gardenlinux/actions/runs/16578530596/job/46888723372

.github/workflows/lima_upload_oidc.yml
```diff
@@ -1,12 +1,10 @@
 name: build and publish images for use with lima-vm
 
 on:
-  push:
-    branches:
-      - lima-upload  
   schedule:
     - cron: '0 11 * * 2'
-    
+  workflow_on_dispatch:
```
fwilhe:
```suggestion
  workflow_dispatch:
```

are you testing how well I review?

for reference: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch

fwilhe: ok I see this is outdated, sorry did only see the commit

.github/workflows/lima_upload_oidc.yml
```diff
@@ -0,0 +1,90 @@
...
+        SHA_ARM=$(ls *arm*.sha512)
+        gsutil cp "gardenlinux.yaml" gs://${{ secrets.GCP_LIMA_BUCKET }}/gardenlinux.yaml
+        gsutil cp "$IMAGE_AMD" gs://${{ secrets.GCP_LIMA_BUCKET }}/gardenlinux-amd64-today.qcow2
+        gsutil cp "$SHA_AMD" gs://${{ secrets.GCP_LIMA_BUCKET }}/gardenlinux-amd64-today.qcow2.sha512
+        gsutil cp "$IMAGE_ARM" gs://${{ secrets.GCP_LIMA_BUCKET }}/gardenlinux-arm64-today.qcow2
+        gsutil cp "$SHA_ARM" gs://${{ secrets.GCP_LIMA_BUCKET }}/gardenlinux-arm64-today.qcow2.sha512
+ 
+    - name: Output public URL
+      run: | 
+        echo "amd64 image uploaded to https://images.gardenlinux.io/gardenlinux-amd64-today.qcow2"
```
fwilhe: we should remove this as the echo is useless because of github's masking

Screenshot 2025-07-29 at 18 09 26

PR #3177, opened 2025-07-10
Reduce number of features included in 'lima' image

fwilhe:
The lima-vm image included a bunch of features that are probably not useful or needed in the lima use-case. This PR excludes some features to build smaller and more focussed images for the lima use-case.

Related to https://github.com/gardenlinux/gardenlinux/issues/2965


saubhikdattagithub (approved):
approved as discussed (ssh and other features)

PR #3183, opened 2025-07-11
test-ng: arch independent

yeoldegrove:

Be arch independent in test-ng.

### DIFF
--- a/tests-ng/util/build_runtime.sh
+++ b/tests-ng/util/build_runtime.sh
@@ -48,7 +48,10 @@ export PYTHONPATH="$tmpdir/runtime/lib/python3.14/site-packages"
 # Create site-packages directory in target runtime
 mkdir -p "$tmpdir/runtime/lib/python3.14/site-packages"
 
-pip install --target "$tmpdir/runtime/lib/python3.14/site-packages" -r "$requirements"
+pip install \
+    --only-binary=:all: \
+    --target "$tmpdir/runtime/lib/python3.14/site-packages" \
+    -r "$requirements"
 
 # Create the final runtime archive
 tar -c -C "$tmpdir/runtime" . | gzip >"$output"
--- a/tests-ng/util/build_runtime.sh
+++ b/tests-ng/util/build_runtime.sh
@@ -28,14 +28,27 @@ if [ "$0" != /init ]; then
 	exec podman run --rm -v "$(realpath -- "${BASH_SOURCE[0]}"):/init:ro" -v "$PWD:/mnt" -w /mnt "$image_id" /init "$@"
 fi
 
-arch="$1"
+arch_host="$(uname -m)"
+arch_target="$1"
 requirements="$2"
 output="$3"
 
-mkdir "$tmpdir/runtime"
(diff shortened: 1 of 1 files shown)


tests-ng/util/build_runtime.sh
```diff
@@ -28,14 +28,27 @@ if [ "$0" != /init ]; then
 	exec podman run --rm -v "$(realpath -- "${BASH_SOURCE[0]}"):/init:ro" -v "$PWD:/mnt" -w /mnt "$image_id" /init "$@"
 fi
 
-arch="$1"
+arch_host="$(uname -m)"
+arch_target="$1"
 requirements="$2"
 output="$3"
 
+# Download host architecture Python runtime for package installation
```
nkraetzschmar: maybe make this conditional on `$arch_host != $arch_target`

tests-ng/util/build_runtime.sh
```diff
@@ -28,14 +28,27 @@ if [ "$0" != /init ]; then
...
+curl -sSLf "https://github.com/astral-sh/python-build-standalone/releases/download/20250626/cpython-3.14.0b3%2B20250626-$arch_target-unknown-linux-gnu-install_only.tar.gz" | gzip -d | tar -x -C "$tmpdir/runtime" --strip-components 1
+
+# Use host Python to install packages into target runtime
+export PATH="$tmpdir/host_runtime/bin:$PATH"
+export PYTHONPATH="$tmpdir/runtime/lib/python3.14/site-packages"
+
+# Create site-packages directory in target runtime
+mkdir -p "$tmpdir/runtime/lib/python3.14/site-packages"
+
+pip install --target "$tmpdir/runtime/lib/python3.14/site-packages" -r "$requirements"
```
nkraetzschmar: if some pip packages install / use native libs, will this properly work to pick the right versions when the running python interpreter is native and not target arch?

nkraetzschmar (changes requested):

tests-ng/util/build_runtime.sh
yeoldegrove: Will add this :+1:
Should not matter that much though, as we're running this for multiple architectures.

tests-ng/util/build_runtime.sh
yeoldegrove: yep, good point. We must either "not use packages that need native libs" or could use [manylinux](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#manylinux) and call e.g.

```
❯ pip install \
  [--platform manylinux_2_28_x86_64] \ # optional or default value will be chosen
  --only-binary=:all: \
  --target /tmp/tmp.YTSZALiznI/runtime/lib/python3.13/site-packages \
  -r util/requirements.txt
```

Just tried this out with `numpy`:

```
❯ grep numpy util/requirements.txt
numpy
❯ tar tf .build/tests-ng-dist-c2ebd358157d3075295573e01bc977dff96e34bfa015be2183fe187d2932a2e8.tar.gz|grep numpy-config
./runtime/aarch64/lib/python3.13/site-packages/bin/numpy-config
./runtime/x86_64/lib/python3.13/site-packages/bin/numpy-config
```

Seems to work.
I cannot really judge right now how many packages with native lib dependencies we will have and if manylinux wheels exist for all of them (and how this looks from a security perspective).

yeoldegrove: We have to switch to a stable python version, e.g. `3.13`, though as for `3.14` no wheels exist.

yeoldegrove:
superseded by #3192 3192

PR #3188, opened 2025-07-15
fix: remove hardcoded iscsi initiator name from image

Leon-hk:
- Installing `open-iscsi` automatically generated a unique InitiatorName in the `/etc/iscsi/initiatorname.iscsi` file
- This name was released with the image, which could cause problems when running it in the same network on several machines

**Definition of Done:**

I was not able to test iscsi locally as `test_iscsi.py` is skipped

### DIFF
--- a/features/iscsi/exec.config
+++ b/features/iscsi/exec.config
@@ -3,3 +3,6 @@
 sed -i 's/^\(node.session.scan\).*/\1 = manual/' /etc/iscsi/iscsid.conf
 sed -i 's/^\(node.session.auth.chap_algs\).*/\1 = MD5/' /etc/iscsi/iscsid.conf
 touch /var/lib/open-iscsi-configured0
+
+# Delete unique InitiatorName from image
+rm /etc/iscsi/initiatorname.iscsi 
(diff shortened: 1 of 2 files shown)


features/iscsi/exec.config
```diff
@@ -3,3 +3,7 @@
 sed -i 's/^\(node.session.scan\).*/\1 = manual/' /etc/iscsi/iscsid.conf
 sed -i 's/^\(node.session.auth.chap_algs\).*/\1 = MD5/' /etc/iscsi/iscsid.conf
 touch /var/lib/open-iscsi-configured0
+
+# Delete unique InitiatorName from initiatorname.iscsi
+sed -i '/^InitiatorName=/d' /etc/iscsi/initiatorname.iscsi
+mv /etc/iscsi/initiatorname.iscsi /etc/iscsi/initiatorname.iscsi.template
```
moritzkirchnersap: I think I would create the template somewhere else and explicitly and use it in file.include for it so that it's in the image and can be used by the systemd unit unterneath. Or is there a reason why we should use /etc/iscsi/initiatorname.iscsi and delete the initiator name from it?

The seds above were an explicit customer requirement.

moritzkirchnersap (changes requested):

features/iscsi/exec.config
Leon-hk: I have moved the template to files.include. Tests are passing

nkraetzschmar (approved):

PR #3208, opened 2025-07-23
Ensure build reproducibility 

Leon-hk:
- Guarantees the same behavior for testing
- Hardcoded unique identifiers can be found 

**Definition of Done:**
- [ ] Finish the workflow
- [ ] Create issue with the results

### DIFF
--- a/.github/workflows/reproducable_test.yml
+++ b/.github/workflows/reproducable_test.yml
@@ -1,0 +1,101 @@
+name: Check build reproducibility
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-reproducibility
+  cancel-in-progress: true
+
+# To be able to test actions on a different branch, change to workflow_dispatch after debugging is done
+on:
+  push:
+    branches:
+      - "feat/check-reproducibility"
+jobs:
+  build_a:
+    name: Build A
+    uses: ./.github/workflows/build.yml
+    permissions:
+      id-token: write
+    with:
+      version: now
+      target: nightly
+      platform_test_tag: latest
+      fail_fast: true
+      prefix: a-
+    secrets:
+      aws_region: ${{ secrets.AWS_REGION }}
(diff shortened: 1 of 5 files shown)


NotTheEvilOne:
Thanks for that interesting feature. Do you think it's feasable to reuse already created `nightly` artifacts instead of creating a new "a" prefixed set? You could use the input `workflow_id` similar to `.github/workflows/publish.yml` and `.github/workflows/publish_s3.yml`. This could even allow you to reuse already generated matrices to know what is available to generate for comparison instead of recreating everything from scratch.

Leon-hk:
I've tried comparing two nightly builds but had a lot of changes caused by the different build times, even tough they originated from the same commit.
I therefore changed my plan to build explicitly for the reproducibility test. But I may add a toggle so one can select if the builds are fresh or reused.
In order to do this, I first have to create a reliable whitelist for the nightly files.

Leon-hk:
End of the year state of this PR: Ready to merge (except resolving conflicts), Waiting for re-review of @gehoern 

(I'm also currently working on porting the python script to the python-gardenlinux-lib, but this can be done in a separate PR)

.github/workflows/reproducable_test.yml
```diff
@@ -0,0 +1,265 @@
...
+      if: always() && needs.run_number_b.result == 'success'
+      uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # pin@v4.3.0
+      with:
+        name: build-${{ matrix.flavor }}-${{ matrix.arch }}
+        path: B
+        run-id: ${{ needs.run_number_b.outputs.id }}
+        github-token: ${{ github.token }}
+    - name: Compare builds
+      run: |
+        podman unshare /bin/bash -c '\
```
nkraetzschmar, line 208: Instead of `/bin/bash -c` it might be cleaner to first create a script using here doc and have `podman unshare` execut that script that way you can avoid all the escape chars.

PR #3209, opened 2025-07-23
Configure serial console for GDCH Images

moritzkirchnersap:
Once Garden LInux gets booted in GDCH it's not quite clear what the image is doing - adding the serial console allows to at least see the login prompt to know that there's no error on boot.

Fixes #

**Definition of Done:**
- [x ] The code is sufficiently documented

Same as gcp

**Release note**:

```feature user
configure the serial console for GDCH images
```


yeoldegrove (approved):
Could not test this but looks reasonable.

features/gdch/file.include/etc/kernel/cmdline.d/10-console.cfg
```diff
@@ -0,0 +1 @@
+CMDLINE_LINUX="console=tty0 console=ttyS0 $CMDLINE_LINUX"
```
nkraetzschmar: Based on the kernel doc:

> You can specify multiple console= options on the kernel command line.
> 
> The behavior is well defined when each device type is mentioned only once. In this case, the output will appear on all requested consoles. And the last device will be used when you open /dev/console. So, for example:
> 
> console=ttyS1,9600 console=tty0
> defines that opening /dev/console will get you the current foreground virtual console, and kernel messages will appear on both the VGA console and the 2nd serial port (ttyS1 or COM2) at 9600 baud.
> 
> The behavior is more complicated when the same device type is defined more times. In this case, there are the following two rules:
> 
> The output will appear only on the first device of each defined type.
> 
> /dev/console will be associated with the first registered device. Where the registration order depends on how kernel initializes various subsystems.
> 
> This rule is used also when the last console= parameter is not used for other reasons. For example, because of a typo or because the hardware is not available.
> 
> The result might be surprising. For example, the following two command lines have the same result:
> 
> console=ttyS1,9600 console=tty0 console=tty1
> console=tty0 console=ttyS1,9600 console=tty1
> The kernel messages are printed only on tty0 and ttyS1. And /dev/console gets associated with tty0. It is because kernel tries to register graphical consoles before serial ones. It does it because of the default behavior when no console device is specified, see below.

So that would imply that setting `console=ttyS0 console=ttyS0,38400n8d` would only consider the first `console=ttyS0` arg.
Maybe instead this should be

```
CMDLINE_LINUX="$CMDLINE_LINUX console=tty0"
```

such that the serial console with explicit settings appears first.

nkraetzschmar (changes requested):

PR #3212, opened 2025-07-24
Add workflow for CPE

Akendo:

For some of our stakeholders, we need to provide a [CPE](https://nvd.nist.gov/products/cpe) within a release. This file needs to be encoded as a JSON file. In this PR, we add a workflow that allows us to 'predict' an entry CPE entry. This CPE file is incomplete but sufficient for our stakeholder. In a separate workflow, we will notify the NIST about a release; however, it takes several days or weeks until a proper CPE entry is updated by the NIST. With the prediction, we allow stakeholders to update their dataset and do not have to wait until the NIST has responded. 

The code responsible is placed in [here](https://github.com/gardenlinux/cpe/blob/main/predict-cpe.py). In this separate repository, which also builds the container.  

https://github.com/gardenlinux/security/issues/147

### DIFF
--- a/.github/workflows/cpe.yml
+++ b/.github/workflows/cpe.yml
@@ -16,10 +16,10 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Build local container 
-        run: docker pull ghcr.io/gardenlinux/cpe:latest
+        run: podman pull ghcr.io/gardenlinux/cpe:latest
 
       - name: Generate CPE 
-        run: docker run -i ghcr.io/gardenlinux/cpe:latest -p "${{ inputs.version }}" | tail -n1 > gardenlinux-cpe.json
+        run: podman run -i ghcr.io/gardenlinux/cpe:latest -p "${{ inputs.version }}" | tail -n1 > gardenlinux-cpe.json
 
       - name: Upload asset to release
         uses: softprops/action-gh-release@v2
--- a/.github/workflows/cpe.yml
+++ b/.github/workflows/cpe.yml
@@ -1,0 +1,31 @@
+name: Generate and upload CPE to a release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version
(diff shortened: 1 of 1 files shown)


Akendo:
> I like the idea of not having the "generate CPE" code itself in our main repo. Wouldn't this also fit into https://github.com/gardenlinux/python-gardenlinux-lib?

Thank you for your feedback. I do not know what the lib does, but I also want to prevent feature creep here. I rather like to have something lightweight. Furthermore, I will add one more feature to update the release with the proper CPE entry. But I feel like that this is enough.

yeoldegrove:
I won't block this but would also be interested in @NotTheEvilOne's opinion.

.github/workflows/cpe.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  build-and-upload:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Build local container 
+        run: docker pull ghcr.io/gardenlinux/cpe:latest
```
NotTheEvilOne: Please use `podman` as we use it for other workflow container executions as well.

.github/workflows/cpe.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      - name: Generate CPE 
+        run: docker run -i ghcr.io/gardenlinux/cpe:latest -p "${{ inputs.version }}" | tail -n1 > gardenlinux-cpe.json
```
NotTheEvilOne: I guess we could integrate the cpe to an GitHub action (see `python-gardenlinux-lib`). This is not needed to be done for that PR thought.

Akendo: I think this won't add any value as of now. We might need CPE later on, but as of now it's as good as it needs to be. Please focus more on other tasks.

PR #2875, opened 2025-03-19
Feat/tofu aws ami

yeoldegrove:
**What this PR does / why we need it**:

This scope of this feature is spinning up test instances of existing and released gardenlinux versions quickly on cloud providers.

#### Use existing published cloud images from the official gardenlinux releases
You can boot published cloud images from the official gardenlinux releases like e.g. [Release 1592.6](https://github.com/gardenlinux/gardenlinux/releases/tag/1592.6) by referencing the image path and image name in the `IMAGE_PATH` and `IMAGE_NAME` variables.
```bash
# use an existing image for ali
IMAGE_PATH=cloud:// IMAGE_NAME=m-d7oaobmhtxwjsw07e3as \
  make --directory=tests/platformSetup \
  ali-gardener_prod-amd64-tofu-config

# use an existing image for aws
IMAGE_PATH=cloud:// IMAGE_NAME=ami-073c0d4753f70877d \
  make --directory=tests/platformSetup \
  aws-gardener_prod-amd64-tofu-config

# use an existing image for azure
IMAGE_PATH=cloud:// IMAGE_NAME=/communityGalleries/gardenlinux-13e998fe-534d-4b0a-8a27-f16a73aef620/images/gardenlinux-gen2/versions/1592.6.0 \
  make --directory=tests/platformSetup \
  azure-gardener_prod-amd64-tofu-config

# use an existing image for gcp
IMAGE_PATH=cloud:// IMAGE_NAME=projects/sap-se-gcp-gardenlinux/global/images/gardenlinux-gcp-gardener-prod-amd64-1592-6-cb05e11f \
  make --directory=tests/platformSetup \
  gcp-gardener_prod-amd64-tofu-config
```
After creating the OpenTofu variables file, you can create the cloud resources with the `tofu-apply` make targets as usual.

**Definition of Done:**
- [x] The code is sufficiently documented
- [x] Shared the changes with the Team so everyone is aware
- [x] The code is appropriately tested
- [x] Checked if the code needs to be backportet to release branches of maintained versions (perform the actual backport after the merge to `main`)

**Special notes for your reviewer**:

This scope of this feature is spinning up test instances of existing and released gardenlinux versions quickly on cloud providers.
[...]

PR #3213, opened 2025-07-24
Fix renaming of cname to image_name

yeoldegrove:

Fix renaming of cname to image_name
This broke in https://github.com/gardenlinux/gardenlinux/pull/2875


ByteOtter (approved):
We have talked about it already. So, LGTM 👍🏼

Issue #3159, opened 2025-07-04
Better Test Framework: Remote Test Runner Tooling


nkraetzschmar:
Tooling to run test suite against remote targets:

- [x] aws, gcp, azure, ....
- [x] gardener nodes (#3693)
- [x] arbitrary remote machines via `ssh`

## Gardener

For gardener this should ideally work analogous to how `kubectl node-shell` works: some privileged container that uses `nsenter` to get into the host OS, but keep some directory fd of the test suit carried into the other namespace to then bind mount somewhere on `/tmp` during the test run.

Issue #3156, opened 2025-07-04
Redesign the Test Framework to run in-place on Target Systems


nkraetzschmar:
# Proposal: Redesign the Test Framework to run in-place on Target Systems

## Current Design and Its Limitations

The current  test framework was originally designed to validate build artifacts immediately after creation. It assumes a model where tests are executed remotely from the developer's local machine. Specifically, `pytest` runs locally and connects to the system under test via SSH. Even when the target is a container or chroot, the system is configured to run an `sshd` instance to support this remote execution model.

While this model is fine for freshly built, ephemeral targets, it leads to several architectural and operational limitations when applied to broader or longer-lived use cases:

1. **System Mutation**: The test framework modifies the target system to enable SSH and may also install packages or make other system-level changes during the test run. This makes the framework unsuitable for use on production or long-lived systems.

2. **Feature and Platform Skipping is Fragmented**:

   * There are two completely orthogonal mechanisms for making tests conditional on the enabled features. First, tests are selectively included based on the feature directories they live in—only the directories corresponding to enabled features are loaded, meaning some tests are never even discovered unless a given feature is enabled. Additionally, features can also explicitly disable tests from other features by placing a `.disable` file of the same name inside their own directory. Tests are often only discovered from enabled feature directories, creating an implicit and opaque mapping between features and test inclusion. Second, tests can include fixtures that implicitly cause them to be skipped if the corresponding feature is not enabled.
   * This dual mechanism leads to confusion and makes it very difficult to reason about whether a test will run by looking at the test file alone.
   * Conditional skipping based on fixtures is scattered across inconsistent and sometimes contradictory naming schemes. Examples include:

     * `aws`, `gcp`, `azure`, `container`, `kvm`, `metal`, `non_aws`, `non_gcp`, `non_azure`, `non_container`, `non_kvm`, `non_metal`, `non_trustedboot`, `non_feature_gardener` (note the inconsistent naming: some use a `feature_` prefix, some don't), `non_feature_githubActionRunner`
     * Many others across `features.py`, `platforms.py`, `miscellaneous.py`, `elements.py`, etc.

3. **Opaque Test Logic**: Most feature-specific tests are thin wrappers around global helper functions. These helpers often mix utility logic (e.g., parsing a file) with test assertions, and many are only used by a single test. This separation makes it difficult to understand what a test is actually verifying.

[...]

PR #3214, opened 2025-07-25
Document design decisions for new test framework

fwilhe:
Based on https://github.com/gardenlinux/gardenlinux/issues/3156 and https://github.com/gardenlinux/gardenlinux/issues/3159

This PR documents the design decisions for the new testing framework.


docs/architecture/decisions/0006-new-test-framework-in-place-self-contained-test-execution.md
```diff
@@ -0,0 +1,45 @@
...
+- **Reporting:** Output can be collected via stdout/stderr, persisted as JUnit XML, or exported in other formats.
+- **Migration:** Enables incremental migration; old and new frameworks can coexist during transition.
+
+## Alternatives Considered
+
+We evaluated packaging the test suite as an [OCI container](https://opencontainers.org) and leveraging [systemd system extensions](https://www.freedesktop.org/software/systemd/man/latest/systemd-sysext.html) for deployment.
+
+These approaches were not adopted due to the following limitations:
+- **Software Availability:** Container runtimes and systemd are not present in all target environments.
+- **Permission Requirements:** Both methods require elevated privileges, which may not be feasible or desirable in production or restricted systems.
```
nkraetzschmar: The permissions/privileges concern is not just, even not primarily, about production systems. One of the main reasons to aim for privilege less test setup and execution is to allow running tests as a developer (against a locally build artifact) without requiring any privileges beyond what's needed to build (so unprivileged user namespaces only).

docs/architecture/decisions/0007-non-invasive-read-only-testing.md
```diff
@@ -0,0 +1,31 @@
...
+
+Key limitations:
+- **System Mutation:** Tests may alter system configuration, install software, or change files, making the framework unsuitable for production validation.
+- **Operational Risk:** Running tests on production systems risks unintended changes, outages, or security issues.
+- **Limited Applicability:** The framework cannot be safely used on systems that must remain unchanged, such as production servers or persistent VMs.
+
+## Decision
+
+The redesigned test framework will treat the system under test as strictly read-only. Tests must not modify system state, install packages, enable services, or change configuration. The framework itself will not require SSH setup or any other mutation of the target system. All test logic must operate without side effects, ensuring that the system remains unchanged before, during, and after test execution.
```
nkraetzschmar: While in principle, yes, we do want to avoid having tests modify the system state; this cannot always be avoided for all tests (e.g. testing of loading kernel modules works). So instead we should build the tests such that tests that modify global system state **MUST** be clearly marked via a pytest marker and skipped unless the test framework is run with an explicit arg to allow modifications.
These tests would then only ever be run on ephemeral targets, such as local test only VMs or during the platform tests.

docs/architecture/decisions/0009-flexible-distribution-and-reporting.md
```diff
@@ -0,0 +1,35 @@
...
+- **Cloud Provider Diversity:** Different cloud platforms require different image formats and deployment strategies.
+- **Reporting Weaknesses:** Lack of standardized, exportable reporting formats makes it difficult to integrate test results with external systems or dashboards.
+
+## Decision
+
+We will support multiple mechanisms for distributing the test suite and collecting results, tailored to the needs of each environment:
+
+- **Distribution Mechanisms:** The test suite may be delivered via scp, cloud-init/user_data, OCI registry/artifact, image attach, or other platform-specific methods. The suite will be packaged as a relocatable tarball or directory, and may be built on demand or pulled as a build artifact.
+- **Cloud Provider Support:** Image formats and deployment workflows will be adapted for each provider (e.g., raw, vhd, qcow2), with research into automation and API integration for disk/image attachment.
+- **Reporting:** Test results will be exported in a [diki](https://github.com/gardener/diki)-compatible format as part of the MVP, enabling integration with external systems and dashboards. Additional formats (e.g., JUnit XML) may be supported as needed.
```
nkraetzschmar: Did we discuss/agree on the test result output already? My last state was that we would primarily aim for human readable test run logs with other formats as a side channel output, but I guess these details are still open for discussion.

fwilhe: Diki requirement is mentioned here https://github.com/gardenlinux/gardenlinux/issues/3156#issuecomment-3078947562

I mean, it could be designed in a way that new outputs formats can be extended without rewriting the tests/frameworks, so maybe this should really not be specified here

fwilhe:
@nkraetzschmar should be reviewable now where the initial framework is merged

docs/architecture/decisions/0009-flexible-distribution-and-reporting.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Decision
+
+We will support multiple mechanisms for distributing the test suite and collecting results, tailored to the needs of each environment:
+
+- **Distribution Mechanisms:** The test suite may be delivered via scp, cloud-init/user_data, OCI registry/artifact, image attach, or other platform-specific methods. The suite will be packaged as a relocatable tarball or directory, and may be built on demand or pulled as a build artifact.
+- **Cloud Provider Support:** Image formats and deployment workflows will be adapted for each provider (e.g., raw, vhd, qcow2), with research into automation and API integration for disk/image attachment.
+- **Reporting:** Test output will be flexible and allow custom formats in a plugin-based system, so that new formats are easy to add. The default output will be a human-readable text format, machine readable outputs such as a [diki](https://github.com/gardener/diki)-compatible format or JUnit xml output may be added later.
+- **Backchannel for Logs:** Mechanisms such as scp, custom APIs, or direct S3 uploads will be explored for retrieving logs and results from the system under test.
```
NotTheEvilOne: I think this part should be formulated more general. Such as: `As for reporting itself the delivery of the logs and results will be implemented in a plugin-based approach to suite execution scenarios best.`

docs/architecture/decisions/0006-new-test-framework-in-place-self-contained-test-execution.md
```diff
@@ -0,0 +1,45 @@
...
+
+Running tests on a developer machine will not require multiple invocations of `make`, building and testing an image are a single script invocation each.
+
+## Consequences
+
+- **Safety:** The system under test is not mutated by the test framework (no SSH setup, no package installs, no filesystem changes).
+- **Portability:** The test suite can run in containers, chroots, VMs, bare metal, and production systems.
+- **Flexibility:** Multiple deployment mechanisms are supported; the framework is not tied to a specific transport or runtime.
+- **Maintainability:** The framework is easier to reason about and maintain, as tests run in a predictable, local context.
+- **Reporting:** Output can be collected via stdout/stderr, persisted as JUnit XML, or exported in other formats.
```
Leon-hk, line 36: It is not clear to me how the output is collected via stdout/stderr. Sure, there must be a command dispatch to run the tests, but to me it seems like the new framework minimizes shell interaction. I'm probably missing something here, but maybe you could add some details to make things clear

fwilhe: IMO the important part here is that we agree both on the plain text human readable output as a default, and to allow other machine readable outputs via plugins. I think you are interpreting a bit too much into 'can be collected', but maybe the wording is also not idea. I'm open for better suggestions.

docs/architecture/decisions/0009-flexible-distribution-and-reporting.md
fwilhe: reworded

PR #3219, opened 2025-07-28
feat: _usi: append EROFS directly to initrd

nkraetzschmar:
For UKI based images use zstd compression inside the EROFS rather than wrapping the EROFS inside dracuts xz compression. The EROFS is then appended as a secondary uncompressed CPIO archive after dracuts.

This results in faster pre-init kernel boot time by reducing up-front decompression overhead and decompressing more on-the-fly.

In tests this improved the total boot time by a factor of 31% from 25.5 sec to 17.5 sec.

```
before: 
5.171s (firmware) + 364ms (loader) + 10.513s (kernel) + 6.020s (initrd) + 3.442s (userspace) = 25.511s

after:
5.140s (firmware) + 445ms (loader) + 2.421s (kernel) + 6.220s (initrd) + 3.301s (userspace) = 17.529s
```

Switching the EROFS compression to zstd also has a positive effect on the amount of `Shmem` used during idle, since the EROFS image is always resident in memory (backed by tmpfs). Cutting it by 33% from 285828 kB down to 191884 kB.

---

> [!IMPORTANT]
> This requires kernel support for zstd compressed EROFS via the `CONFIG_EROFS_FS_ZIP_ZSTD` so https://github.com/gardenlinux/package-linux/pull/144 must be merged first.

### DIFF
--- a/features/_usi/image.esp.tar
+++ b/features/_usi/image.esp.tar
@@ -11,14 +11,11 @@ rootfs="$(mktemp -d)"
 mount -t tmpfs -o size="$TEMPFS_SIZE" tmpfs "$rootfs"
 tar --extract --xattrs --xattrs-include '*' --directory "$rootfs" < "$input"
 
-# # shellcheck disable=SC2115
-# rm -rf "$rootfs/var"
-# mkdir "$rootfs/var"
-
 echo "creating root EROFS disk"
 
-erofs="$(mktemp -u)"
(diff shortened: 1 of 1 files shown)


fwilhe:
merged the kernel change, guess we should wait until tomorrow before we merge this change to have the new kernel in the nightly build

5kt (approved):

PR #3249, opened 2025-08-04
Initial Version of New Test Framework

nkraetzschmar:

Adds the basic framework of the proposed new test framework. This includes:

- Tooling to build the test framework distribution
- Run tests on local VMs and in a chroot
- Partial implementation of running tests on OpenTofu provisioned cloud VMs, but this is not production ready and AWS only for now, so this part is unused in GitHub actions for now and only included as a starting point for a proper implementation


tests-ng/util/build.makefile
```diff
@@ -0,0 +1,34 @@
...
+	echo '🛠️  bundling test framework as disk image'
+	./$^ $@
+
+.build/runtime.tar.gz: util/build_runtime.sh util/requirements.txt | .build
+	echo '🛠️  building python runtime'
+	./$^ $@
+
+.build/edk2-%: | .build
+	echo '⬇️  fetching EDK2 ($*)'
+	curl -sSLf "https://github.com/gardenlinux/edk2-build/releases/download/edk2-stable202505/edk2-$*" > $@
```
fwilhe, line 34: https://github.com/gardenlinux/edk2-build/blob/a56ccdf9a6208fbc079d76980157597ccfa6eaf8/.github/workflows/build.yml#L23C15-L23C42

minor: I think this violates one of our security considerations (should use a commit hash, or even better not use a third party action if we can avoid it, creating a release is also possible using the `gh` cli tool which is available on runners)

tests-ng/_test_dummy.py
```diff
@@ -0,0 +1,42 @@
...
+	pass
+
+@pytest.mark.root
+def test_root_user(shell):
+	assert shell("id -u", capture_output=True).stdout.strip() == "0"
+
+def test_regular_user(shell):
+	assert shell("id -u", capture_output=True).stdout.strip() != "0"
+
+def test_shell_print(shell):
```
fwilhe, line 26: general comment/question: is there a good reason against using type hintes here? (and in all other places in the framework)? I think this would improve usability for developers a lot so it is easy to discover what the type of this thing is and what methods/properties it has.

the old framework seems to use type hints

tests-ng/plugins/booted.py
```diff
@@ -0,0 +1,24 @@
...
+	return system_booted
+
+def pytest_addoption(parser):
+	parser.addoption(
+		"--system-booted",
+		action="store_true",
+		help="Set if the system under test was booted instead of running in a chroot or container. This will enable kernel, systemd, and other system level tests."
+	)
+
+def pytest_configure(config):
```
fwilhe, line 15: here type hints would be of large value imo because this defines an api and knowing the type of `config`, `parser` etc here is essential

fwilhe:
General question/suggestion on test style: What about using small functions for readability here like in this example?

Screenshot 2025-08-04 at 13 17 12

I think that this strikes a good balance between the DSL suggestions we had and the 'we don't have confusing abstractions' argument. I think a function with a good name here is better than a comment, but still this should make the test easier to read by an external auditor. Of course those functions need to be written with care and can also make things worse if applied in a bad way, but just overall I think expressing intent in the test with the actual implementation detail being one abstraction layer down would be good design.

@nkraetzschmar thoughts on this?

tests-ng/test_users.py
```diff
@@ -0,0 +1,25 @@
+import os
+import pwd
+import pytest
+import stat
+
+def test_service_accounts_have_nologin_shell():
+	for entry in pwd.getpwall():
+		if entry.pw_uid >= 1000:
+			continue
+		if entry.pw_name in {"root", "sync"}:
```
fwilhe: what is the deal with this 'sync' account? why exclude that here?

fwilhe: Okay I can see this is taken from tests/helper/tests/users.py but that still does not explain the logic

fwilhe:
Thought: would it make sense to introduce a common logging system already? I think it would be good if we do this early to avoid multiple people re-inventing the wheel. It would be great if the `test-ng` script has a `debug` flag or similar to enable verbose logging in tests, where the test itself should then take care of logging things that might be useful to understand test failures easily

tests-ng/util/run_qemu.sh
```diff
@@ -0,0 +1,123 @@
...
+	-chardev file,id=test_output,path="$tmpdir/serial.log"
+	-device virtio-serial
+	-device virtserialport,chardev=test_output,name=test_output
+	-chardev socket,id=chrtpm,path="$tmpdir/swtpm.sock"
+	-tpmdev emulator,id=tpm0,chardev=chrtpm
+	-device "$qemu_tpm_dev",tpmdev=tpm0
+)
+
+swtpm socket --tpmstate backend-uri="file://$tmpdir/swtpm.permall" --ctrl type=unixio,path="$tmpdir/swtpm.sock" --tpm2 --daemon --terminate
+"qemu-system-$arch" "${qemu_opts[@]}" | stdbuf -i0 -o0 sed 's/\x1b\][0-9]*\x07//g;s/\x1b[\[0-9;!?=]*[a-zA-Z]//g;s/\t/    /g;s/[^[:print:]]//g'
```
fwilhe, line 124: what is happening here? I'm sure there is a good reason for this, but it is not at all obvious what's going on here. please if it can't be done in an easier to understand way at least drop a comment explaining why we need this line.

tests-ng/test_users.py
nkraetzschmar: The sync user is a historical unix relict that just runs /bin/sync and exits. Not really used anymore (I think), but still around for compatibility. We skip it here since its shell isn’t a real login shell, but still harmless.

nkraetzschmar:
> tests fail because of unexpected users, but also it does not use the run_cloud script yet.. so I guess that is expected? @nkraetzschmar can you elaborate on what the status is there?

The capability to mark users as expected is there, but for the QEMU test this is not used yet. So yes, we'll need to fix that before merging this into main

tests-ng/util/run_qemu.sh
nkraetzschmar: It's some regex voodoo to strip (mostly) all terminal control sequences, such that the output on stdout is plain text and the serial console log doesn't clear or modify your terminal. The final `s/` command then strips all remaining non printable characters, such that even if we forgot certain control sequences, those would become random characters in the output rather then true control sequences.

tests-ng/test_users.py
```diff
@@ -16,7 +16,7 @@ def test_root_home_permissions():
 	perm = stat.S_IMODE(mode)
 	assert perm == 0o700, f"/root has incorrect permissions: {oct(perm)}"
 
-@pytest.mark.feature("not _dev")
+@pytest.mark.feature("not _dev and not lima")
```
nkraetzschmar: What's the reason to exclude lima specifically here? I assume lima adds some expected users? then maybe we should add those accordingly

fwilhe: ok what would be the appropriate way to resolve this? My assumption was that since we're skipping _dev, this is the way to go with all features that add a user

it seems like the lima build is adding a `debian` user, where would I configure this correctly so the test passes?

nkraetzschmar: hmmmmm 🤔 I'm also not too happy with excluding `_dev` here, but for now (since it's just dev) I wouldn't worry too much, but for non _dev features I'd like to avoid this.... Ideally we need some way to properly define what users a given feature expects..... for cloud-init created users I've just build https://github.com/gardenlinux/gardenlinux/pull/3249/commits/e6c4f2a8a3283860ae25d7339f8fcc29dde8aada but obviously that does not cover lima. The alternative is have some hardcoded list of feature => user somewhere in the run tests script, but I also don't really like that idea.... Input for better way to do this highly appreciated.

fwilhe:
Noticed your files use tabs, I think we consistently use spaces, please format your files accordingly with the established standard in the project

fwilhe:
@nkraetzschmar I played around with type hints and pushed a commit to feat/test-ng-type-hints that seems to work fine
please incorporate those if that works on your side too

PR #3256, opened 2025-08-05
use type hints for python api

fwilhe:


tests-ng/plugins/booted.py
```diff
@@ -1,24 +1,29 @@
...
 system_booted = False
 
-def is_system_booted():
+
+def is_system_booted() -> bool:
     return system_booted
 
-def pytest_addoption(parser):
+
+def pytest_addoption(parser: pytest.Parser):
```
NotTheEvilOne, line 11: If nothing is returned, usually `-> None` [1] is defined in these cases.

[1] https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html

tests-ng/plugins/booted.py
```diff
@@ -1,24 +1,29 @@
...
+
+def pytest_configure(config: pytest.Config):
     global system_booted
     system_booted = config.getoption("--system-booted")
 
     config.addinivalue_line("markers", "booted: mark test to run only on a booted target, i.e. not in a container or chroot")
 
-def pytest_collection_modifyitems(config, items):
+
+def pytest_collection_modifyitems(config: pytest.Config, items):
```
NotTheEvilOne: `items: List[Any]` or `Any` if we don't know for sure that the item's value is a `pytest.Item` object.

nkraetzschmar: It should be properly defined _somewhere_ what type items is here, right? so `List[Any]` wouldn't make too much sense.

tests-ng/plugins/booted.py
fwilhe: compare to @ccwienk 's comment, it seems like there is no consensus on that. I don't have a strong opinion on that.

tests-ng/plugins/booted.py
fwilhe: oh that was an oversight, there is a type for that. thanks for spotting

PR #3260, opened 2025-08-06
Allow publishing lima images for released Garden Linux versions

fwilhe:
Add a 'version' parameter to the workflow for publishing lima images, which allows us to publish images for stable versions.

This is not yet part of the proper release workflow.

Part of https://github.com/gardenlinux/gardenlinux/issues/2963


saubhikdattagithub (approved):
lgtm

Issue #3181, opened 2025-07-11
Better Test Framework: Migrate all Tests


nkraetzschmar:
> [!NOTE]  
> See Sub-issues for open test migrations.

PR #3265, opened 2025-08-08
Port SSH tests to new framework

fwilhe:
Convert test to assert desired properties are part of the sshd config.

Part of https://github.com/gardenlinux/gardenlinux/issues/3181

### DIFF
--- a/tests-ng/plugins/sshd.py
+++ b/tests-ng/plugins/sshd.py
@@ -22,13 +22,7 @@ class Sshd:
                     else:
                         self._sshd_config[key] = {self._sshd_config[key], value}
                 else:
-                    # fixme: this is not ideal. what we want here is to have comma separated string values
-                    # as a set to make comparing them easier because the order of entries does not matter
-                    # the 3 is an arbitrary number to allow for real string values with commas
-                    if value.count(",") >= 3:
-                        self._sshd_config[key] = set(value.split(","))
-                    else:
-                        self._sshd_config[key] = value
+                    self._sshd_config[key] = value
             elif len(parts) == 1:
                 key = parts[0]
                 self._sshd_config[key] = None
--- a/tests-ng/plugins/sshd.py
+++ b/tests-ng/plugins/sshd.py
@@ -1,0 +1,9 @@
+import pytest
+from .shell import ShellRunner
+
+class Sshd:
+    def __init__(self, shell: ShellRunner):
+        self._shell = shell
+
+    def get_config():
+        
--- a/tests-ng/plugins/sshd.py
+++ b/tests-ng/plugins/sshd.py
@@ -42,17 +42,6 @@ class Sshd:
         self._sshd_config = config
         return config
 
-    def get_config_section(self, key: str) -> str:
+    def get_config_section(self, key: str) -> str|set:
         config = self.read_config()
-        return config.get(str.lower(key))
-
-    def get_normalized_sets(
-        self,
-        sshd_config_item: str,
-        actual_value: set,
-        expected_value: set,
-    ) -> tuple[set, set]:
-        assert isinstance(actual_value, set), f"{sshd_config_item} should be a set"
-        actual_set = {str(v).lower() for v in actual_value}
-        expected_set = {str(v).lower() for v in expected_value}
-        return actual_set, expected_set
+        return config.get(str.casefold(key))
--- a/tests-ng/plugins/sshd.py
+++ b/tests-ng/plugins/sshd.py
@@ -4,44 +4,39 @@ from .shell import ShellRunner
 
 class Sshd:
     def __init__(self, shell: ShellRunner):
-        self._shell = shell
-        result = shell("/usr/sbin/sshd -T", capture_output=True, ignore_exit_code=True)
+        sshd_config_test_command = shell("/usr/sbin/sshd -T", capture_output=True, ignore_exit_code=True)
         assert (
-            result.returncode == 0
-        ), f"Expected return code 0, got {result.returncode}"
-        self._sshd_config_text = result.stdout
-        self._sshd_config = None
-
-    def read_config(self) -> dict:
-        if self._sshd_config is not None:
-            return self._sshd_config
-        config = {}
-        for line in self._sshd_config_text.splitlines():
+            sshd_config_test_command.returncode == 0
+        ), f"Expected return code 0, got {sshd_config_test_command.returncode}"
+        self._sshd_config = {}
+        for line in sshd_config_test_command.stdout.splitlines():
             if not line.strip():
                 continue
             parts = line.split(maxsplit=1)
             if len(parts) == 2:
                 key, value = parts
                 # If key already exists, convert to set and append
-                if key in config:
-                    if isinstance(config[key], set):
-                        config[key].add(value)
+                if key in self._sshd_config:
+                    if isinstance(self._sshd_config[key], set):
+                        self._sshd_config[key].add(value)
                     else:
-                        config[key] = {config[key], value}
+                        self._sshd_config[key] = {self._sshd_config[key], value}
                 else:
                     # fixme: this is not ideal. what we want here is to have comma separated string values
                     # as a set to make comparing them easier because the order of entries does not matter
                     # the 3 is an arbitrary number to allow for real string values with commas
                     if value.count(",") >= 3:
-                        config[key] = set(value.split(","))
+                        self._sshd_config[key] = set(value.split(","))
                     else:
-                        config[key] = value
+                        self._sshd_config[key] = value
             elif len(parts) == 1:
                 key = parts[0]
-                config[key] = None
-        self._sshd_config = config
-        return config
+                self._sshd_config[key] = None
+
+
+    def get_config(self) -> dict:
+        return self._sshd_config
+
 
     def get_config_section(self, key: str) -> str|set:
-        config = self.read_config()
-        return config.get(str.casefold(key))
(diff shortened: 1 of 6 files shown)


tests-ng/test_basics.py
```diff
@@ -1,15 +1,18 @@
...
+from plugins.shell import ShellRunner
+from plugins.systemd import Systemd
+
 
 def test_gl_is_support_distro():
     with open("/etc/os-release", "r") as f:
         assert "ID=gardenlinux" in [ line.strip() for line in f], "/etc/os-release does not contain gardenlinux vendor field"
 
-def test_no_man(shell):
+def test_no_man(shell: ShellRunner):
```
nkraetzschmar, line 11: not sure if type hints within the tests are good 🤔 
in the more complex plugin code: sure, but in the tests themself I'd vote for keeping it as short and simple as possible.

tests-ng/test_ssh.py
```diff
@@ -0,0 +1,81 @@
...
+    # "UsePrivilegeSeparation sandbox",
+    # Disable Tunneling and Forwarding
+    # AllowTcpForwarding no # required for Gardener
+    "X11Forwarding no",
+    # use PAM
+    "UsePAM yes",
+]
+
+
+def _create_list_of_tuples(input: str) -> list[tuple[str, set[str]]]:
```
nkraetzschmar: We should aim for no util functions directly in tests. so probably a good idea to introduce some util directory or something like that to contain such functions.

tests-ng/test_ssh.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.mark.root
+@pytest.mark.feature("ssh")
+@pytest.mark.parametrize("sshd_config_item", required_sshd_config)
+def test_sshd_has_required_config(sshd_config_item: str, shell: ShellRunner):
+    result = shell("/usr/sbin/sshd -T", capture_output=True, ignore_exit_code=True)
+    assert result.returncode == 0, f"Expected return code 0, got {result.returncode}"
+
+    sshd_config = _create_list_of_tuples(result.stdout)
```
nkraetzschmar: Although seeing how the util function is used here it might actually be better to move the whole `sshd` exec plus parsing into a fixture plugin.

I'd suggest something like a sshd fixture that has a `get_config` or similar function that returns the sshd config as a simple dict rather than some list of tuples. Then we could make the `required_sshd_config ` a simple dictionary to make that part more legible too.

tests-ng/test_ssh.py
```diff
@@ -0,0 +1,81 @@
+import pytest
+from plugins.shell import ShellRunner
+from plugins.systemd import Systemd
+
+required_sshd_config = [
```
nkraetzschmar: this would probably be more legible as a dict

tests-ng/test_basics.py
fwilhe: the reason I've put it in is because this allows vs code to give better completions, i.e. it knows `shell` can be called as a function. I think this adds enough benefit to accept downsides.

tests-ng/test_ssh.py
fwilhe: agreed, was not sure where to put it, but that is a legit todo item

tests-ng/test_ssh.py
fwilhe: oh that makes sense, I like this, not yet used too much to thinking in those terms

tests-ng/plugins/sshd.py
```diff
@@ -0,0 +1,47 @@
...
+
+class Sshd:
+    def __init__(self, shell: ShellRunner):
+        self._shell = shell
+        result = shell("/usr/sbin/sshd -T", capture_output=True, ignore_exit_code=True)
+        assert (
+            result.returncode == 0
+        ), f"Expected return code 0, got {result.returncode}"
+        self._sshd_config_text = result.stdout
+        self._sshd_config = None
```
nkraetzschmar: What's the point of splitting the exec and parse stage? Like why already run the command here, but then only parse it into a dict on demand? I'd say either always read and parse the config at `__init__` time and make read_config a simple getter method, or move loading and parsing fully into the `read_config` function without any caching

tests-ng/plugins/utils.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+# Various utility functions to make tests more readable
+# This should not contain test-assertions, but only abstract details that make tests harder to read
+
+def equals_ignore_case(a: str, b: str) -> str:
+    return a.casefold() == b.casefold()
+
+
+def get_normalized_sets(
```
nkraetzschmar: not sure a function that always handles 2 sets is the best abstraction for a utility function, since that is pretty specific to this usecas.

tests-ng/test_ssh.py
```diff
@@ -0,0 +1,64 @@
...
+    "Subsystem": "sftp /usr/lib/openssh/sftp-server -f AUTHPRIV -l INFO",
+    "PermitRootLogin": "No",
+    "X11Forwarding": "no",
+    "UsePAM": "yes",
+}
+
+@pytest.mark.booted
+@pytest.mark.root
+@pytest.mark.feature("ssh")
+@pytest.mark.parametrize("sshd_config_item", required_sshd_config)
```
nkraetzschmar, line 53: Why do we need parameterize here? Wouldn't it be easier to read to just have a simple test case that compares the dicts to each other? that would probably be more readable then having the parameterize do this per element.

fwilhe: The advantage I see with parametrize is that it is easier to see why a test failed

Example with all tests working (nice to see which properties were tested)

Screenshot 2025-08-11 at 15 59 46

Example with failing test

Screenshot 2025-08-11 at 16 01 06

I am open to changing this, I do see the upside of comparing the whole dict in one go.

tests-ng/plugins/utils.py
fwilhe: agree, did not think of it in this generic way, just wanted to get rid of the details in the test. should be possible to do it with an arbitrary number of sets I think

tests-ng/plugins/sshd.py
fwilhe: moved the logic to init

not sure if this is 'pythonic', at least in java it is typically not appreciated to do 'real work' in the constructor

I don't have a strong opinion on this

nkraetzschmar: I don't have much opinion on in constructor or in getter either, just having half the work in one place and half in the other isn't great 😅

tests-ng/test_ssh.py
nkraetzschmar: ok, that's a good point and the nicer test output probably outweighs the slightly more complex syntax 👍

tests-ng/test_ssh.py
```diff
@@ -0,0 +1,64 @@
...
+def test_sshd_has_required_config(sshd_config_item: str, shell: ShellRunner):
+    sshd = Sshd(shell)
+    actual_value = sshd.get_config_section(sshd_config_item)
+    expected_value = required_sshd_config[sshd_config_item]
+    if is_set(expected_value):
+        assert is_set(actual_value), f"{actual_value} should be a set"
+        actual_set, expected_set = get_normalized_sets(actual_value, expected_value)
+        missing_sshd_configuration = expected_set - actual_set
+        assert not missing_sshd_configuration, f"{sshd_config_item}: missing values {missing_sshd_configuration}"
```
nkraetzschmar: Is it intended, that we only fail on entries missing in the set, and ignore additional set entries in the actual_set?

nkraetzschmar: If not, then we could just move this into the `equals_ignore_case` making this function able to compare both strings and sets, then we don't even need a `get_normalized_sets` function at all. That would obviously require to also consider additional entries in the `actual_set` as errors.

fwilhe: yes

I think this is what the current tests do (unless I am misremembering badly), and also I think it makes sense to allow a diff that way. I view the expected as 'this has to be in, but it is okay if there is more'.

tests-ng/plugins/sshd.py
```diff
@@ -0,0 +1,42 @@
...
+                if key in self._sshd_config:
+                    if isinstance(self._sshd_config[key], set):
+                        self._sshd_config[key].add(value)
+                    else:
+                        self._sshd_config[key] = {self._sshd_config[key], value}
+                else:
+                    # fixme: this is not ideal. what we want here is to have comma separated string values
+                    # as a set to make comparing them easier because the order of entries does not matter
+                    # the 3 is an arbitrary number to allow for real string values with commas
+                    if value.count(",") >= 3:
```
nkraetzschmar: Instead of splitting here where we don't know if it should be a set or a string with commas: we could initially parse everything as a string and then only convert to a set in the comparison based on wether the type of the expected value is a set or a string

tests-ng/test_ssh.py
nkraetzschmar: instead of subtracting the sets and checking their difference is empty we could just use `expected_set.issubset(actual_set)`. That would make this more readable.

nkraetzschmar (changes requested):

tests-ng/plugins/sshd.py
fwilhe: yes, that's much better than what I did. thanks for the suggestion. updated.

tests-ng/test_ssh.py
fwilhe: yes that's nicer to read. updated.

PR #3278, opened 2025-08-12
Enable test runner setup using lima-vm

fwilhe:
This should allow iterating on tests quickly


fwilhe:
Using the VS Code remote ssh plugin and the python plugin, this setup allows to run tests in the vs code debugger 💪 

Screenshot 2025-08-13 at 14 16 52

PR #3282, opened 2025-08-12
Test denylisted packages are installed

fwilhe:
Part of https://github.com/gardenlinux/gardenlinux/issues/3181

### DIFF
--- a/tests-ng/test_packages.py
+++ b/tests-ng/test_packages.py
@@ -1,0 +1,22 @@
+import pytest
+from plugins.dpkg import Dpkg
+from plugins.shell import ShellRunner
+
+denylist = [
+    "rlogin",
+    "rsh",
+    "rcp",
+    "telnet",
+    "libdb5.3:amd64",
+    "libdb5.3:arm64",
+    "libssl1.1:amd64",
+    "libssl1.1:arm64",
+]
+
+
+@pytest.mark.parametrize("denied_package", denylist)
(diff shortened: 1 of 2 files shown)


tests-ng/test_packages.py
```diff
@@ -0,0 +1,22 @@
...
+    "libdb5.3:amd64",
+    "libdb5.3:arm64",
+    "libssl1.1:amd64",
+    "libssl1.1:arm64",
+]
+
+
+@pytest.mark.parametrize("denied_package", denylist)
+def test_no_denylisted_packages(denied_package: str, shell: ShellRunner):
+    dpkg = Dpkg(shell)
```
nkraetzschmar, line 19: Rather than importing and then creating a class here, dpkg could just be a fixture

tests-ng/test_packages.py
```diff
@@ -0,0 +1,22 @@
...
+from plugins.dpkg import Dpkg
+from plugins.shell import ShellRunner
+
+denylist = [
+    "rlogin",
+    "rsh",
+    "rcp",
+    "telnet",
+    "libdb5.3:amd64",
+    "libdb5.3:arm64",
```
nkraetzschmar, line 11: We should consider if we can improve this, rather then having a hardcoded list, just do the cross product of the package names and all architects returned by `dpkg --print-architecture` and `dpkg --print-foreign-architectures`

Issue #3283, opened 2025-08-13
in-place update failed with Error: 1877.2.0: not found


pnpavlov:
Sonu is trying to update OS inplace from 1877.1.0 to 1877.2.0 . It fails with error - Error: [ghcr.io/gardenlinux/gardenlinux:1877.2.0](http://ghcr.io/gardenlinux/gardenlinux:1877.2.0): not found

```
gardenlinux-update 1877.2
Verified OK
downloading ghcr.io/gardenlinux/gardenlinux@sha256:24b6dc1699d81a29018e8f29b89d834316ae1a1b4f2344fd6415d3d80b583ff1 -> /efi/EFI/Linux/aws-gardener_prod_usi-amd64-1877.2+3.efi
```
works, so I think we have to do the same what we did last time. copy this image and create a tag 1877.2.0  as well

Vincinator:
Done:

```
docker buildx imagetools create  --tag ghcr.io/gardenlinux/gardenlinux:1877.2.0 ghcr.io/gardenlinux/gardenlinux@sha256:0279faf60dcc79690ccc6758bc6ad77e22f20ab010916a4480931e397ca89be8
```

```
sha256:0279faf60dcc79690ccc6758bc6ad77e22f20ab010916a4480931e397ca89be8
```
is hash of 1877.2

PR #3287, opened 2025-08-13
fix: add .0 to version of published GL OCI Image

Vincinator:
Adds additional tag to oci manifest containing all published GL flavours. 

* uses Oras to tag full manifest (not only the subset docker/podman would pull)

In-Place update tool currently requires `MAJOR.MINOR.PATCHLEVEL` but GL only tags `MAJOR.MINOR`

Additionally, colleagues in Gardener are adapting gardenlinux-update tool so `MAJOR.MINOR.PATCHLEVEL` will translated to `MAJOR.MINOR`.

Decision if we need additional OCI tag with this PR is still open. 

**Release note**:

```feature user

```

### DIFF
--- a/.github/workflows/publish_oci_containers.yml
+++ b/.github/workflows/publish_oci_containers.yml
@@ -153,6 +153,7 @@ jobs:
           podman manifest add ${{ needs.determine_environment.outputs.repository }}:${version} ${{ needs.determine_environment.outputs.repository }}:amd64-${version}
           podman manifest add ${{ needs.determine_environment.outputs.repository }}:${version} ${{ needs.determine_environment.outputs.repository }}:arm64-${version}
           podman manifest push ${{ needs.determine_environment.outputs.repository }}:${version}
+
   bare_flavors:
     needs: determine_environment
     name: Publish bare flavors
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_oci_containers.yml
```diff
@@ -345,3 +346,43 @@ jobs:
             --container ${{ needs.determine_environment.outputs.repository }} \
             --version ${{ inputs.version }} \
             --manifest_folder manifests
+  add_semver_tag_to_manifest:
```
NotTheEvilOne, line 349: `gl-oci` supports the argument `--additional_tag` to handle cases like additional tags easily. Call should look similar to:
```
          gl-oci push-manifest \
            --dir "${CNAME}" \
            --container ${{ needs.determine_environment.outputs.repository }} \
            --arch ${{ matrix.arch }} \
            --version ${{ inputs.version }} \
            --cname "${CNAME}" \
            --cosign_file digest.txt \
            --manifest_file "oci_manifest_entry_${CNAME}.json"`
            --additional_tag "${{ inputs.version }}.0"
```

NotTheEvilOne (changes requested):

.github/workflows/publish_oci_containers.yml
Vincinator: Thanks, could you please take over this PR? 
Then we do not need to manually create those tags in the future and avoid #3283

PR #3288, opened 2025-08-13
test: add tests for sapmachine feature

fwilhe:
Related to https://github.com/gardenlinux/gardenlinux/issues/3181, but this feature did not have tests before


tests-ng/test_sapmachine.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+from plugins.dpkg import Dpkg
+from plugins.apt import Apt
```
nkraetzschmar, line 3: The apt plugin really should just implement a fixture that instantiates and returns an Apt class object, that way there would be no need for this import or the later instantiation and the function could just use the apt fixture as an input directly

nkraetzschmar (dismissed):

tests-ng/test_sapmachine.py
fwilhe: yeah, that's much better, thanks for the hint

PR #3290, opened 2025-08-14
feat: test-ng: add tests-ng/util/.python.lock and caching of downloaded python archives

yeoldegrove:

Add `tests-ng/util/.python.lock` and caching of downloaded python archives of downloaded python archives.
This makes version bumps and visibility a lot easier and speeds up local development (python archives do not need to be downloaded every time).

Supersedes #3192


tests-ng/util/build_runtime.sh
```diff
@@ -2,47 +2,98 @@
...
-	exec podman run --rm -v "$(realpath -- "${BASH_SOURCE[0]}"):/init:ro" -v "$PWD:/mnt" -w /mnt "$image_id" /init "$@"
+	exec podman run --rm -v "$(realpath -- "${BASH_SOURCE[0]}"):/init:ro" -v "$PWD:$mntdir" -w "$mntdir" "$image_id" /init "$@"
 fi
 
 requirements="$1"
 output="$2"
 
+python_lock_file="$mntdir/util/.python.lock"
+if [ -f "$python_lock_file" ]; then
+	. "$python_lock_file"
```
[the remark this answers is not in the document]
yeoldegrove: excluded this

tests-ng/util/.python.lock
NotTheEvilOne, line 1: I wonder if `python.env.sh` or similar would not be a better name based on the sourceable content of it?

yeoldegrove: Yes, that makes it more clear. Renamed it.

tests-ng/util/build_runtime.sh
```diff
@@ -2,47 +2,101 @@
 
 set -eufo pipefail
 
+mntdir="/mnt"
```
nkraetzschmar: What's the point of making this a variable? What benefit does writing `$mntdir` give us over `/mnt`?

tests-ng/util/python.env.sh
```diff
@@ -0,0 +1,10 @@
+# python.env.sh
```
nkraetzschmar, line 1: Since this file essentially specifies how the runtime is build, this would need to be added to the dependencies of `.build/runtime.tar.gz` in `build.makefile` to ensure any changes here will also trigger a re-build.

tests-ng/util/build_runtime.sh
```diff
@@ -2,47 +2,101 @@
...
 	cleanup
-	exec podman run --rm -v "$(realpath -- "${BASH_SOURCE[0]}"):/init:ro" -v "$PWD:/mnt" -w /mnt "$image_id" /init "$@"
+	exec podman run --rm -v "$(realpath -- "${BASH_SOURCE[0]}"):/init:ro" -v "$PWD:$mntdir" -w "$mntdir" "$image_id" /init "$@"
 fi
 
 requirements="$1"
 output="$2"
 
+python_lock_file="$mntdir/util/python.env.sh"
+if [ -f "$python_lock_file" ]; then
```
nkraetzschmar: Since this file is part of the repo and not something user provided, we can safely assume it is here and just rely on the `set -e` of the script to fail on non existent file rather then explicit checking.

yeoldegrove:
> In general I'd question the usefulness of caching the python runtime download. The runtime part already is its own make target in `build.makefile` precisely to allow fast local iterations of re-building the test framework, without having to re-download and re-build the runtime component each time. Adding an additional step to cache the base runtime would only really cover the case where we update the `requirements.txt` and presumably that will occur infrequently enough, that adding additional complexity just to speed up those cases appears unnecessary.
> 

Maybe my intention comes from somebody with a slow downstream internet connection that spent several "minutes" already waiting for the some tar file to be downloaded ;)
Caching is of course also a factor if you change `build_dist.sh` itself.


> I do quite like the env file specifying the version and checksums though 👍

@nkraetzschmar As you like this and the checksumming part is at least as complex as the caching, are you OK to keep it in?

The rest of your comments will be addressed of course.

tests-ng/util/build_runtime.sh
yeoldegrove: Added this out of a habit but do not care and can switch this back.

tests-ng/util/build_runtime.sh
yeoldegrove: :+1:

tests-ng/util/python.env.sh
yeoldegrove: Nice catch. Will add this.

yeoldegrove: @nkraetzschmar `util/python.env.sh` is now a Makefile dependency and an additional input parameter of `util/build_runtime.sh`.

Issue #3296, opened 2025-08-14
Test Framework: Add marker for mutating test


fwilhe:
Tests should declare that they mutate system state if they do so one can skip them if running on a system where mutation of state is not acceptable

PR #3297, opened 2025-08-14
Add marker for mutating system state

fwilhe:
Add a marker so a test can declare that it mutates system state.
Adds a test that starts and tests the ssh service as an example.

Fixes https://github.com/gardenlinux/gardenlinux/issues/3296


fwilhe:
looks good:

Screenshot 2025-08-14 at 18 07 12

nkraetzschmar:
I'd suggest to do this the other way round: have non mutating be the default and only allow mutating test with an explicit `--allow-system-modifications` flag. That way we can better prevent accidentally running these tests on the wrong targets.

fwilhe:
> I'd suggest to do this the other way round: have non mutating be the default and only allow mutating test with an explicit `--allow-system-modifications` flag. That way we can better prevent accidentally running these tests on the wrong targets.

Adapted.

I now enabled this flag for qemu and chroot tests, I think that is safe because anyway those are short-lived 'machines'.

Did not enable it in run_cloud, not sure if it should be enabled there

tests-ng/plugins/systemd.py
```diff
@@ -38,6 +38,9 @@ def is_running(self, unit_name: str) -> bool:
         result = self._shell(f"systemctl is-active {unit_name}", capture_output=True, ignore_exit_code=True)
         return result.stdout.strip() == "active"
 
+    def start_unit(self, unit_name: str):
+        self._shell(f"systemctl start {unit_name}")
```
nkraetzschmar, line 45: maybe add some safety check here, to ensure this is never run when modifying tests are off. similar to how sysctl fixture checks system booted state: https://github.com/gardenlinux/gardenlinux/blob/a0d77a9bdce38490a4b2ce62a03b25015a553014/tests-ng/plugins/sysctl.py#L14

fwilhe:
@yeoldegrove @nkraetzschmar addressed your feedback. If you have time for another review, that would be appreciated

PR #3298, opened 2025-08-14
Unify `python-gardenlinux-lib` based flavors parsing

NotTheEvilOne:
Currently `bin/flavors_parse.py` provides duplicated code and features with `python-gardenlinux-lib` to parse and return information about GardenLinux flavors. To unify and ensure code quality this PR replaces it with the CLI utilities of `python-gardenlinux-lib`.


nkraetzschmar (approved):
why the rename from `flavors_parse` to `gl-flavors-parse`?

Otherwise LGTM, and don't really have a strong opinion on the name, so approved, but would love a reason for the name change in the PR for documentation purposes

PR #3300, opened 2025-08-15
test: Allow expected users with home dir

fwilhe:
'expected users' are typically users created by configuration management, and they typically have a home dir which is expected and not a defect

Related to https://github.com/gardenlinux/gardenlinux/issues/3181

### DIFF
--- a/tests-ng/test_users.py
+++ b/tests-ng/test_users.py
@@ -1,13 +1,14 @@
 import os
 import pwd
+from typing import Set
 import pytest
 import stat
 
-def test_service_accounts_have_nologin_shell(expected_users):
+def test_service_accounts_have_nologin_shell(expected_users: Set[str]):
     for entry in pwd.getpwall():
         if entry.pw_uid >= 1000:
             continue
-        if entry.pw_name in expected_users + ["root", "sync"]:
+        if entry.pw_name in expected_users + {"root", "sync"}:
             continue
         assert entry.pw_shell in [ "/usr/sbin/nologin", "/bin/false" ], f"User {entry.pw_name} has unexpected shell: {entry.pw_shell}"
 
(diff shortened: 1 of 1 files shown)


tests-ng/test_users.py
```diff
@@ -1,13 +1,14 @@
...
 import pytest
 import stat
 
-def test_service_accounts_have_nologin_shell():
+def test_service_accounts_have_nologin_shell(expected_users: Set[str]):
     for entry in pwd.getpwall():
         if entry.pw_uid >= 1000:
             continue
-        if entry.pw_name in {"root", "sync"}:
+        if entry.pw_name in expected_users | {"root", "sync"}:
```
nkraetzschmar, line 11: Even for expected users, if they are service accounts (UID < 1000) they should NOT have a login shell. For regular users with UID starting at 1000 this is skipped anyways.

fwilhe: Sadly, lima creates the user with id 501

`fwi:x:501:1000:Wilhelm:/home/fwi.linux:/bin/bash`

so either you allow me to skip this test when the lima feature is enabled, or that. I'd like to have green tests in my lima images :)

nkraetzschmar: If lima uses UID 501 for regular users, then we should probably set `UID_MIN` in `/etc/login.defs` accordingly (in the lima feature) and modify this test to use that value dynamically as the range of system users rather than hardcoded 1000

fwilhe: ok I was actually not aware that lima
a ) by default uses the uid from the host (which seems to be 501 in my case on macos)
b ) allows you to configure custom user details https://github.com/lima-vm/lima/blob/bf3d444055983d41b8835209dfeeeb56cdbc4176/templates/default.yaml#L321-L338

thus I can live with the current situation and just configure my user accordingly.

fwilhe:
ok I was actually not aware that lima
a ) by default uses the uid from the host (which seems to be 501 in my case on macos)
b ) allows you to configure custom user details https://github.com/lima-vm/lima/blob/bf3d444055983d41b8835209dfeeeb56cdbc4176/templates/default.yaml#L321-L338

thus I can live with the current situation and just configure my user accordingly.

Issue #3302, opened 2025-08-15
Address faults in gardenlinux lima setup


fwilhe:
There are a few known issues when running Garden Linux guests in lima. They should be addressed.

- [ ] There are failed units in the booted Garden Linux instance
- [ ] The Garden Linux instance can only boot with QEMU, not with vz (Apple Virtualisation)

fwilhe:
With regards to failed units: Those seem to be related to filesystem mounts

https://lima-vm.io/docs/config/mount/ describes that different versions of lima use different tools for mounting host directories.

In lima >= 1.0, 9P is the default and for some reason this does not work in Garden Linux

Configuring `mountType: "reverse-sshfs"` seems to work at least on macos hosts, mounts are working and there are no failed units.

fwilhe:
regarding vz, if I build a `kvm-lima` image, that does not boot, if I build a `metal-lima` image it does (but fails again with mount issues). Assuming this is of something missing in our cloud kernel config, but have not investigated further.

PR #3303, opened 2025-08-15
fix: configure 9p as mountTypesUnsupported for lima image

fwilhe:
This addresses an issue with failed systemd units in a Garden Linux lima guest.

This configuration is also done for Debian guests, so it should be fine for us to do the same.

See https://github.com/gardenlinux/gardenlinux/issues/3302 for more info.


saubhikdattagithub (approved):
lgtm

PR #3305, opened 2025-08-18
Add vmware test in new testframework

SujanaSubr:
Port vmware test to new Test Framework. Currently test_content_in_file.py and test_check_files.py is ported
Fixes #

**Definition of Done:**

**Release note**:

```feature user

```


tests-ng/test_vmware.py
```diff
@@ -0,0 +1,32 @@
...
+        "/usr/bin/dscheck_VMwareGuestInfo",
+        "/usr/lib/python3/dist-packages/cloudinit/sources/DataSourceVMwareGuestInfo.py",
+        "/etc/cloud/cloud.cfg.d/01_debian-cloud.cfg",
+        "/etc/cloud/cloud.cfg.d/99_disable-network-config.cfg",
+        "/etc/cloud/cloud.cfg.d/99_enabled-datasources.cfg"
+    ]
+
+@pytest.mark.feature("vmware")
+@pytest.mark.parametrize("filename", file)
+@pytest.mark.parametrize("content", args)
```
fwilhe: I think this is not what you wanted to do, this should be one parameter with a suitable data structure

tests-ng/test_vmware.py
```diff
@@ -0,0 +1,32 @@
...
+
+def test_file_content(filename: str, content: str):
+    if os.path.isfile(filename):
```
fwilhe: prefer

```
assert os.path.isfile(filename), "your message"
```

over the if / else assert False

tests-ng/test_vmware.py
```diff
@@ -0,0 +1,32 @@
...
+        if (file_content(filename, content)):
+            assert False, f"Found {content} in {filename} which is not expected."
+    else:
+        assert False, "File does not exist"
+
+@pytest.mark.parametrize("files_to_check", vmware_files)
+
+def test_check_file(files_to_check: str):
+    assert os.path.exists(files_to_check), (f"File {files_to_check} could not be found.")
```
fwilhe: please don't use the plural here `files_to_check`, this is called once for each file, so what we have here is a single file

`file_to_check` is much clearer

tests-ng/test_vmware.py
```diff
@@ -0,0 +1,32 @@
...
+    ]
+
+@pytest.mark.feature("vmware")
+@pytest.mark.parametrize("filename", file)
+@pytest.mark.parametrize("content", args)
+
+def test_file_content(filename: str, content: str):
+    if os.path.isfile(filename):
+        if (file_content(filename, content)):
+            assert False, f"Found {content} in {filename} which is not expected."
```
fwilhe: also here put the test condition in the assert

tests-ng/plugins/file_content.py
```diff
@@ -0,0 +1,13 @@
+import re
+import os
+
+
+def file_content(fname, args):
```
fwilhe: this method name does not communicate what the function does

this function returns true if the file contains the line in `args`, so it would be core clear if it was called `line_in_file` or something like that

tests-ng/plugins/file_content.py
```diff
@@ -0,0 +1,13 @@
+import re
+import os
+
+""" This function checks each line and return true if the args is found in the file """
+def check_val_in_file(fname, args):
```
fwilhe: still not happy with this name

`check` sounds like there would be an assertion

```suggestion
def is_line_in_file(file_name, line) -> bool:
```

would be much better imo

tests-ng/plugins/file_content.py
```diff
@@ -0,0 +1,13 @@
+import re
+import os
```
fwilhe: none of those imports are used here

tests-ng/test_vmware.py
```diff
@@ -0,0 +1,29 @@
...
+@pytest.mark.parametrize("filename,content", val_to_check)
+
+def test_file_content(filename: str, content: str):
+    assert os.path.isfile(filename), "File does not exist"
+    assert not check_val_in_file(filename, content), (
+            f"Found {content} in {filename} which is not expected." )
+
+@pytest.mark.parametrize("file_to_check", vmware_files)
+
+def test_check_file(file_to_check: str):
```
fwilhe: test should have a name that clearly communicates what it does, for example

```suggestion
def test_required_files_exist(file_to_check: str):
```

tests-ng/test_vmware.py
```diff
@@ -0,0 +1,29 @@
...
+        "/usr/lib/python3/dist-packages/cloudinit/sources/DataSourceVMwareGuestInfo.py",
+        "/etc/cloud/cloud.cfg.d/01_debian-cloud.cfg",
+        "/etc/cloud/cloud.cfg.d/99_disable-network-config.cfg",
+        "/etc/cloud/cloud.cfg.d/99_enabled-datasources.cfg"
+    ]
+
+@pytest.mark.feature("vmware")
+@pytest.mark.parametrize("filename,content", val_to_check)
+
+def test_file_content(filename: str, content: str):
```
fwilhe: test should have a name that clearly communicates what it does, for example
```suggestion
def test_config_does_not_contain_forbidden_value(filename: str, forbidden_value: str):
```

tests-ng/test_vmware.py
```diff
@@ -0,0 +1,29 @@
...
+    assert os.path.isfile(filename), "File does not exist"
+    assert not check_val_in_file(filename, content), (
+            f"Found {content} in {filename} which is not expected." )
+
+@pytest.mark.parametrize("file_to_check", vmware_files)
+
```
fwilhe, line 23: [minor/style] please remove the blank line between the decorator and the function it decorates

tests-ng/test_vmware.py
```diff
@@ -0,0 +1,29 @@
...
+        "/usr/bin/dscheck_VMwareGuestInfo",
+        "/usr/lib/python3/dist-packages/cloudinit/sources/DataSourceVMwareGuestInfo.py",
+        "/etc/cloud/cloud.cfg.d/01_debian-cloud.cfg",
+        "/etc/cloud/cloud.cfg.d/99_disable-network-config.cfg",
+        "/etc/cloud/cloud.cfg.d/99_enabled-datasources.cfg"
+    ]
+
+@pytest.mark.feature("vmware")
+@pytest.mark.parametrize("filename,content", val_to_check)
+
```
fwilhe: [minor/style] please remove the blank line between the decorator and the function it decorates

tests-ng/test_vmware.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+import os
+from plugins.file_content import check_val_in_file
+
+val_to_check = [ ("/etc/cloud/cloud.cfg", "ntp"),
+             ("/etc/cloud/cloud.cfg", "resizefs"),
+             ("/etc/cloud/cloud.cfg", "growpart") ] 
+
+vmware_files = [
```
fwilhe: make intent clear:
```suggestion
vmware_required_files = [
```

tests-ng/test_vmware.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+import os
+from plugins.file_content import check_val_in_file
+
+val_to_check = [ ("/etc/cloud/cloud.cfg", "ntp"),
```
fwilhe: make intent clear:

```suggestion
forbidden_config_values = [ ("/etc/cloud/cloud.cfg", "ntp"),
```

tests-ng/test_vmware.py
SujanaSubr: Done

tests-ng/test_vmware.py
SujanaSubr: Done

tests-ng/test_vmware.py
```diff
@@ -0,0 +1,28 @@
...
+        "/etc/cloud/cloud.cfg.d/01_debian-cloud.cfg",
+        "/etc/cloud/cloud.cfg.d/99_disable-network-config.cfg",
+        "/etc/cloud/cloud.cfg.d/99_enabled-datasources.cfg"
+    ]
+
+@pytest.mark.feature("vmware")
+@pytest.mark.parametrize("filename,content", forbidden_config_values)
+def test_config_for_forbidden_value(filename: str, content: str):
+    assert os.path.isfile(filename), "File does not exist"
+    assert not is_line_in_file(filename, content), (
```
nkraetzschmar: What's the benefit of the helper function here, and why have the helper function iterate over lines? Couldn't we just directly do something like the following in-place here:

```python
with open(filename) as f:
        assert content not in f.read(), f"Found {content} in {filename} which is not expected."
```

SujanaSubr: i thought this function will be used by multiple tests thats why i thought of having it separately here

SujanaSubr: Also earlier check was doing line by line. I am not sure if this will be required in other test cases for some reason so thought of doing the same way. I can also use read function if that is prefered

nkraetzschmar: iirc the old tests used a helper function because the file read had to be abstracted via a remote ssh based file access. with direct local access to the file such a helper function just adds indirection without any benefit.

SujanaSubr: got it. I will update

SujanaSubr: Done

Issue #3319, opened 2025-08-20
Better Test Framework: Add 'reason' field to custom markers like `root` or `booted`?


fwilhe:
I am wondering if it would be good to add a string argument `reason` to those marker, similar to the built-in `skip` marker in pytest.

in this example 

```
@pytest.mark.root # for journalctl
@pytest.mark.booted
def test_no_failed_units(systemd: Systemd, shell: ShellRunner):
    units = systemd.list_units()
    assert len(units) > 1, f"Failed to load systemd units: {units}"
    failed_units = [u for u in units if u.load == 'loaded' and u.active != 'active']
    for u in failed_units:
        print(f'FAILED UNIT: {u}')
        shell(f"journalctl --unit {u.unit}")
    assert not failed_units, f"{len(failed_units)} systemd units failed to load"
```

the only thing that needs root permission is journalctl, and I feel the need to document this. Yes we can use comments, but I think it would be much nicer if this was part of the code itself.

Not a pytest expert, so I'm wondering if it would be good to add logging also in the case for when a test is really executed as root.

Something like 

`PASSED test_basics.py::test_no_failed_units: running as root to enable journalctl output in case of failed units`

Same should go for `booted`, `modifying` and other markers too imo

PR #3320, opened 2025-08-20
Enhance test markers to include optional 'reasons' argument

fwilhe:
This PR allows us to better keep track of the reason for certain markers. This field is optional, so for obvious cases it can be left empty, but I think in most cases it is actually useful to document the reasoning for setting a certain marker.

Fixes https://github.com/gardenlinux/gardenlinux/issues/3319

### DIFF
--- a/tests-ng/plugins/booted.py
+++ b/tests-ng/plugins/booted.py
@@ -20,10 +20,16 @@ def pytest_configure(config: pytest.Config):
     global system_booted
     system_booted = config.getoption("--system-booted")
 
-    config.addinivalue_line("markers", "booted: mark test to run only on a booted target, i.e. not in a container or chroot")
+    config.addinivalue_line("markers", "booted(reason=None): mark test to run only on a booted target, i.e. not in a container or chroot. Optionally provide a reason why mutation is required.")
 
 
 def pytest_collection_modifyitems(config: pytest.Config, items: List[pytest.Item]):
     for item in items:
-        if item.get_closest_marker("booted") and not system_booted:
-            item.add_marker(pytest.mark.skip(reason="not running on a booted system"))
+        booted_marker = item.get_closest_marker("booted")
+        if booted_marker and not system_booted:
+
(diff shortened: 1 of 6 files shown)


tests-ng/plugins/booted.py
```diff
@@ -20,10 +20,16 @@ def pytest_configure(config: pytest.Config):
     global system_booted
     system_booted = config.getoption("--system-booted")
 
-    config.addinivalue_line("markers", "booted: mark test to run only on a booted target, i.e. not in a container or chroot")
+    config.addinivalue_line("markers", "booted(reason=None): mark test to run only on a booted target, i.e. not in a container or chroot. Optionally provide a reason why mutation is required.")
```
nkraetzschmar: ~why mutation is required~ why system must be booted

tests-ng/test_basics.py
```diff
@@ -39,7 +39,7 @@ def test_fhs(shell: ShellRunner):
     for dir in sorted(expected_dirs):
         assert os.path.isdir(f"/{dir}"), f"expected FHS directory /{dir} does not exist"
 
-@pytest.mark.booted
+@pytest.mark.booted(reason="Only useful in a booted system because chroot does not 'boot'")
```
nkraetzschmar: This message is a bit confusing. Maybe just make it "We can only measure startup time if we actually boot the system"

nkraetzschmar (changes requested):
Besides those minor issues: LGTM

tests-ng/plugins/booted.py
fwilhe: copy and waste

thanks for spotting

yeoldegrove (approved):
A small documentation snipped would also be nice here. You could already allign this with https://github.com/gardenlinux/gardenlinux/pull/3321/commits/d31dcae6cd3acd8ccdd1718f0b1eaead7c7ca666 if you like.

Otherwise, great feature.

Issue #3228, opened 2025-07-30
use .requirements file in tofu logic


nkraetzschmar:
**What would you like to be added**:

Instead of making image and VM settings in the tofu setup based on hardcoded mappings of features to required settings along the lines of enabling secureboot if `contains(var.features, "_trustedboot")`, use the `.requirements` file introduced by https://github.com/gardenlinux/builder/pull/121 instead.

**Why is this needed**:

This will make it way clearer when certain image / VM configs are required and will allow for future extendability where potentially at some point different features might require capabilities such as secboot or tpm support as well.

PR #3321, opened 2025-08-20
tests-ng: possibility to login to test target and to skip tests

### LINKED ISSUE #3309: tests-ng: possibility to login to test target

When running `./test-ng --cloud` or also for the qemu use case a developer should be able to login easily to the system under test to do debugging.
`./tests-ng --cloud --login $image` could be build that calls dedicated scripts, e.g. `util/login_cloud.sh` (gathers tofu IP) to login via ssh. These dedicated scripts

**Why is this needed:**

A developer needs to be able debug the (short lived) system under test.

yeoldegrove:

When running `./test-ng --cloud` or also for the qemu use case a developer should be able to login easily to the system under test to do debugging.

- `./test-ng` gets 4 new flags (also look at the README)
  - `--ssh`: Enable SSH access to QEMU VM (`gardenlinux@127.0.01:2222`).
    - cloud: SSHD is always enabled via `cloud-init`.
  - `--skip-tests`: Skip running the actual test suite
  - `--skip-cleanup`: Skip cleanup of cloud resources after testing.
    - QEMU VM: After running/skipping the tests, you can stop/cleanup the VM with `ctrl + c`.
    - cloud: To cleanup up cloud resources after passing the flag, just re-run without the flag.
  - `--test-args`: Pass any commandline argument to `pytest`. Put multiple arguments inside `""`.

- new login scripts:
  - `tests-ng/utils/login_qemu.sh .build/$artifact.raw`
    - uses default user/ip

  - `tests-ng/utils/login_cloud.sh .build/$artifact.raw`
    - uses default user/ip
    - gets user/ip from tofu

- restructure tofu files and variables as preparation for additional modules
  - parts of this, e.g. the login user is related to the rest of the changes

- arch, uefi, secureboot, tpm2 are set in `.build/$image.requirements` ([#3228](https://github.com/gardenlinux/gardenlinux/issues/3228))
  - arch was a unneeded parameter in the `run*.sh` scripts and is eliminated this way

- documentation on how to run tests and login to systems under test

**Why is this needed:**

A developer needs to be able debug the (short lived) system under test.

Fixes #3309

**Notes to reviewer**:

Needs commits from #3315 merged first.

### DIFF
--- a/tests-ng/util/run_cloud.sh
+++ b/tests-ng/util/run_cloud.sh
@@ -57,6 +57,7 @@ cleanup() {
 		echo "⚙️  cleaning up cloud resources"
 		(
 			cd "${tf_dir}"
+			tofu init -var-file "$image_name.tfvars"
 			tofu workspace select "$image_name"
 			tofu destroy -var-file "$image_name.tfvars" --auto-approve
 			tofu workspace select default
@@ -108,7 +109,7 @@ EOF
 echo "⚙️  setting up cloud resources via OpenTofu"
 (
 	cd "${tf_dir}"
-	tofu init
+	tofu init -var-file "$image_name.tfvars"
 	tofu workspace select -or-create "$image_name"
 	tofu apply -var-file "$image_name.tfvars" --auto-approve
 )


tests-ng/README.md
```diff
@@ -1,3 +1,174 @@
...
+- `--skip-cleanup`: Skip cleanup of cloud resources after testing
+
+### Examples
+
+```bash
+# Run chroot tests on a tar image
+./test-ng .build/aws-gardener_prod-amd64-today-13371337.tar
+
+# Run QEMU tests on a raw image with SSH access and skip cleanup
+./test-ng --ssh --skip-cleanup .build/aws-gardener_prod-amd64-today-13371337.raw
```
NotTheEvilOne, line 87: Could you elaborate how `ssh` is activated on cloud images? As far as I know GardenLinux at least researches the opportunity to not enable `sshd` at all on images. Do you think a static runtime in the test framework or a Python based implementation could be used to provide ssh functionality without `openssh-server` being installed?

NotTheEvilOne (approved):
LGTM

Please see question below for deciding if additional work is required to enable `ssh` in all expected environments.

tests-ng/util/run_qemu.sh
```diff
@@ -53,23 +69,42 @@ if [ "$arch" = aarch64 ]; then
...
 exec 1>/dev/virtio-ports/test_output
 exec 2>&1
+EOF
 
-trap 'poweroff -f > /dev/null 2>&1' EXIT
-
-mkdir /run/gardenlinux_tests
-mount -o ro /dev/disk/by-label/GL_TESTS /run/gardenlinux_tests
+if ((ssh)); then
+	ssh_public_key="$(cat ~/.ssh/id_ed25519.pub)"
```
fwilhe: this is an undocumented pre-requisite, users might also have different key formats such as id_rsa, does anything speak against testing for a few common ssh public key filenames and picking the one that is there?

tests-ng/util/run_qemu.sh
```diff
@@ -53,23 +69,42 @@ if [ "$arch" = aarch64 ]; then
...
+    ssh_user="gardenlinux"
+	cat >>"$tmpdir/fw_cfg-script.sh" <<EOF
+systemctl stop sshguard
+systemctl enable --now ssh
+useradd -U -m -G wheel -s /bin/bash $ssh_user
+mkdir -p /home/$ssh_user/.ssh
+chmod 700 /home/$ssh_user/.ssh
+echo "$ssh_public_key" >> /home/$ssh_user/.ssh/authorized_keys
+chown -R $ssh_user:$ssh_user/home/$ssh_user/.ssh
```
fwilhe:
```
chown: missing operand after ‘gardenlinux:gardenlinux/home/gardenlinux/.ssh’
```

not sure what the intention here was, is there a space missing? https://tldr.inbrowser.app/pages/common/chown

fwilhe: might this be a linux/gnu chown vs macos/bsd chown thing?

tests-ng/README.md
yeoldegrove: https://github.com/gardenlinux/gardenlinux/blob/812cebd2aaa1d63a73bbf69f6bc5d038ef358124/tests-ng/util/run_cloud.sh#L68 enables ssh via cloud-init's user_data. This is the same way we currently do it in tests(-non-ng).

I guess getting rid of sshd would be another feature/task (I briefly discussed possibilities with  @nkraetzschmar but we said that this is currently not in focus).

tests-ng/util/run_qemu.sh
yeoldegrove: I'll add something like this to also try rsa keys
```
	for keyfile in ~/.ssh/id_ed25519.pub ~/.ssh/id_rsa.pub; do
		if [ -f "$keyfile" ]; then
			ssh_public_key="$(cat "$keyfile")"
			break
		fi
	done
	if [ -z "$ssh_public_key" ]; then
		echo "No SSH public key found in ~/.ssh (id_ed25519.pub, id_rsa.pub)" >&2
		exit 1
	fi
```

yeoldegrove: After thinking a bit about it I decided that just hard coding the path and generating the key (if missing) sounds better.

tests-ng/util/run_qemu.sh
yeoldegrove: This is clearly a typo (missing space) and will be fixed.

yeoldegrove:
> > new login scripts:
> > tests-ng/utils/login_qemu.sh .build/$artifact.raw
> > gets user/ip from tofu
> 
> this does not seem right
> 
> the path is wrong, should be `tests-ng/util/login_qemu.sh`, and the script does not take the disk image as an argument
> 
> edit: this is from the pr description, in the readme file it seems to be correct. please update the pr description accordingly

Fixed this in the PR description.

yeoldegrove:
> Not going to block a merge, but I still have a few questions/ open points.
> 
> It seems like not everything works on macos as the development host platform, it would be appreciated if we get this tested.
> 
> I'm getting this test failure
> 
> ```
> __________________________________________ test_users_have_no_authorized_keys __________________________________________
> test_ssh.py:80: in test_users_have_no_authorized_keys
>     assert not os.path.exists(key_path), (
> E   AssertionError: user 'gardenlinux' should not have an authorized_keys file: /home/gardenlinux/.ssh/authorized_keys
> ```
> 
> Not sure if setting the expected user parameter will fix this, but definitely users created for running the tests should not make the tests fail, not exactly sure what the right solution is for that.

Yes, this needs to be adressed.

> 
> Also, unclear to me: If I start the vm for interactive usage with skip-tests, then how can I manually run the tests? Did I overlook this in the readme? I would expect some shell script or command. The gardenlinux user's home directory is empty.

Will add something about manual test execution to the README.

tests-ng/README.md
```diff
@@ -1,3 +1,206 @@
...
+
+Before running the test framework, make sure the following dependencies are installed:
+
+- `podman`
+- `make`
+- `curl`
+- `jq`
+- `unzip`
+- `qemu`
+- `qemu-utils`
```
Akendo, line 34: I think you need to have `tofu` installed as well.

Akendo:
``` ./test-ng --cloud aws .build/aws-gardener_prod-arm64-today-local.raw
🛠️   building python runtime
🛠️   building test framework distribution
🛠️   bundling test framework as disk image
⚙️   setting up cloud resources via OpenTofu
./util/run_cloud.sh: line 104: tofu: command not found
⚙️   cleaning up cloud resources
./util/run_cloud.sh: line 53: tofu: command not found

File: tests-ng/README.md [hunk 5]
```diff
@@ -1,3 +1,206 @@
...
+
+```bash
+# For chroot testing (tar files)
+./test-ng .build/$image.tar
+
+# For QEMU VM testing (raw files)
+./test-ng .build/$image.raw
+
+# For cloud provider testing (raw files only)
+./test-ng --cloud aws .build/$image.raw
```
Akendo, line 62: This seems not to work; it asks me for some value. I would expect this to work without any interaction.
```
./test-ng --cloud aws .build/aws-gardener_prod-arm64-today-local.raw
⚙️   setting up cloud resources via OpenTofu
var.cloud_provider
  Which cloud provider to target

  Enter a value:

Comment (Akendo) [reply tests-ng/README.md:62]: Okay, it continues on once `aws` was entered.

File: tests-ng/README.md (hunk 4 above)
Comment (yeoldegrove) [reply tests-ng/README.md:34]: @Akendo This will be added in https://github.com/gardenlinux/gardenlinux/pull/3327/commits/2bf1850e9fd8766fd4533f1405f74140d2ce8d5d in https://github.com/gardenlinux/gardenlinux/pull/3327

File: tests-ng/README.md [hunk 6]
```diff
@@ -1,3 +1,206 @@
...
+./test-ng --cloud gcp .build/$image.raw
+./test-ng --cloud azure .build/$image.raw
+./test-ng --cloud ali .build/$image.raw
+```
+
+### Command Line Flags
+
+- `--cloud <provider>`: Specify cloud provider (aws, gcp, azure, ali). Leave empty to run locally in QEMU VM.
+- `--ssh`: Enable SSH access to QEMU VM (`gardenlinux@127.0.01:2222`). For cloud providers, SSHD is always enabled via `cloud-init`.
+- `--skip-tests`: Skip running the actual test suite
```
Akendo, line 74: I love that we can skip it, but can we also skip a selected set of tests, or can we select to run just *one* ?

tests-ng/README.md
yeoldegrove: There should be a file called `tests-ng/util/tf/aws-gardener_prod-arm64-today-local.tfvars` which sets the `cloud_provider = "aws"`. Could you debug why this is not working for you by e.g. setting `set -x` in `tests-ng/util/run_cloud.sh`?

tests-ng/README.md
```diff
@@ -1,3 +1,206 @@
...
+- `--skip-cleanup`: Skip cleanup of cloud resources after testing. To cleanup up cloud resources after passing the flag, just re-run without the flag.
+
+### Examples
+
+```bash
+# Run chroot tests on a tar image
+./test-ng .build/aws-gardener_prod-amd64-today-13371337.tar
+
+# Run QEMU tests on a raw image with SSH access and skip cleanup
+./test-ng --ssh --skip-cleanup .build/aws-gardener_prod-amd64-today-13371337.raw
```
Akendo, line 87: I've tried this and my VM wasn't starting:

```
./test-ng --ssh --skip-cleanup .build/aws-gardener_prod-arm64-today-local.raw
⚙️  preparing test VM
🚀  starting test VM
UEFI firmware (version  built at 15:27:29 on Jul 21 2025)
```

I think the resouce may become stucked. I've pressed Ctrl + C and saw some test and so on. And I think the VM is running in the background.

tests-ng/README.md
```diff
@@ -1,3 +1,206 @@
...
+
+## Login Scripts
+
+### QEMU Environment
+
+To connect to a running QEMU VM:
+
+```bash
+# Get a SSH Session
+./util/login_qemu.sh
```
Akendo, line 110: Cool!

tests-ng/README.md
```diff
@@ -1,3 +1,206 @@
...
+
+# Get a SSH Session with custom user
+./util/login_qemu.sh --user admin
```
Akendo, line 113: This wasn't working: 
```
./util/login_qemu.sh --user admin
Warning: Permanently added '[127.0.0.1]:2222' (ED25519) to the list of known hosts.
admin@127.0.0.1: Permission denied (publickey).

Comment (Akendo) [reply tests-ng/README.md:113]: And I started it [here](https://github.com/gardenlinux/gardenlinux/pull/3321/files#r2298149429) with the `--ssh` flag.

File: tests-ng/README.md [hunk 10]
```diff
@@ -1,3 +1,206 @@
...
+
+# Execute a command directly
+./util/login_qemu.sh "uname -a"
+
+# Run tests manually after login
+cd /run/gardenlinux-tests && ./run_tests --system-booted --allow-system-modifications --expected-users gardenlinux
```
Akendo, line 119: Is there any way to 'update' the code on the system? Do we mount things through or do I  always have to copy the newer version of the test to the `/run/gardenlinux-tests/tests/` folder?

tests-ng/README.md
Akendo: This was failing, I assume because my user is missing some privileges, but that does not seem right:

```
module.cloud.aws_ebs_snapshot_import.root_disk: Creating...
╷
│ Error: creating EBS Snapshot Import: operation error EC2: ImportSnapshot, https response error StatusCode: 400, RequestID: eff8a1c0-c46f-4951-8253-99393acbb418, api error InvalidParameter: User: arn:aws:sts:::assumed-role/vmimport/vm_import_image is not authorized to per
form: s3:GetObject on resource: "arn:aws:s3:::gl-tests-ng-6c4594d4-images/aws-gardener_prod-arm64-today-local.raw" because no identity-based policy allows the s3:GetObject action
│
│   with module.cloud.aws_ebs_snapshot_import.root_disk,
│   on modules/aws/disks.tf line 107, in resource "aws_ebs_snapshot_import" "root_disk":
│  107: resource "aws_ebs_snapshot_import" "root_disk" {
│
╵
╷
│ Error: creating EBS Snapshot Import: operation error EC2: ImportSnapshot, https response error StatusCode: 400, RequestID: 6bc3546c-b0ee-4162-a32c-6bb2521962f6, api error InvalidParameter: User: arn:aws:sts:::assumed-role/vmimport/vm_import_image is not authorized to per
form: s3:GetObject on resource: "arn:aws:s3:::gl-tests-ng-6c4594d4-images/dist.ext2" because no identity-based policy allows the s3:GetObject action
│
│   with module.cloud.aws_ebs_snapshot_import.test_disk,
│   on modules/aws/disks.tf line 121, in resource "aws_ebs_snapshot_import" "test_disk":
│  121: resource "aws_ebs_snapshot_import" "test_disk" {
│
╵
⚙️   cleaning up cloud resources

Comment (Akendo) [reply tests-ng/README.md:62]: Another hint would be great: This takes a long time; maybe you state that it takes some time to get done.

File: tests-ng/README.md (hunk 9 above)
Comment (yeoldegrove) [reply tests-ng/README.md:113]: `./util/login_qemu.sh` will work without passing `--user admin` and chooses the default `gardenlinux` user in this case.

File: tests-ng/README.md (hunk 7 above)
Comment (yeoldegrove) [reply tests-ng/README.md:87]: Currently this is built in away that if you pass `--skip-cleanup`, qemu is forked into the background and the logfile of the test is tailed. This enables you to login from another shell with the `login_qemu.sh` and kill the QEMU VM by pressing `ctrl+c`.

I can add the `ctrl+c` thingy to the README.

File: tests-ng/README.md (hunk 5 above)
Comment (yeoldegrove) [reply tests-ng/README.md:62]: @Akendo Can you check the `*.tfvars` file and the `set -x`?

On the cloud permissions, I cannot help out... this uses mostly the same tf-code code as our current implementation. Please also double check which AWS profile/account you are using.

File: tests-ng/README.md (hunk 10 above)
Comment (yeoldegrove) [reply tests-ng/README.md:119]: This is taken care of automatically due to `tests-ng/util/tf/build.makefile` dependencies. In case any important test file is touched, the test distribution is rebuild. Chroot and QEMU tests will immediately pick up the new changes (once restarted) and cloud tests will upload a new test-disk (re-running the tests might need some work still).

File: tests-ng/util/run_qemu.sh [hunk 11]
```diff
@@ -107,22 +158,39 @@ qemu_opts=(
...
+fi
+
+echo "🚀  starting test VM"
+
 swtpm socket --tpmstate backend-uri="file://$tmpdir/swtpm.permall" --ctrl type=unixio,path="$tmpdir/swtpm.sock" --tpm2 --daemon --terminate
-"qemu-system-$arch" "${qemu_opts[@]}" | stdbuf -i0 -o0 sed 's/\x1b\][0-9]*\x07//g;s/\x1b[\[0-9;!?=]*[a-zA-Z]//g;s/\t/    /g;s/[^[:print:]]//g'
-cat "$tmpdir/serial.log"
-! ( tail -n1 "$tmpdir/serial.log" | grep failed > /dev/null )
+if ((skip_cleanup)); then
+	"qemu-system-$arch" "${qemu_opts[@]}" | stdbuf -i0 -o0 sed 's/\x1b\][0-9]*\x07//g;s/\x1b[\[0-9;!?=]*[a-zA-Z]//g;s/\t/    /g;s/[^[:print:]]//g' &
```
Akendo, line 204: I wish that this `sed` statment is explained. It's a bit difficult to grasp what it does.

tests-ng/README.md
yeoldegrove: For these use cases I would suggest running the tests manually by e.g. using:
```
./test-ng --ssh --skip-cleanup --skip-tests $image
# other shell
./tests-ng/util/login_qemu.sh /run/gardenlinux-tests/run_tests --system-booted --allow-system-modifications --expected-users gardenlinux $whatever_pytest_param_you_want
```

This would allow immediate feedback without re-provisioning the QEMU VM.

We could also add another flag for just running a single test. Let me look into that.

yeoldegrove: A `--test-args` flag is now added in 1b6baeaf3498bae2696d81309012f25126a46f7e

tests-ng/README.md
yeoldegrove: README is extended.

tests-ng/util/run_qemu.sh
yeoldegrove: An inline explanation is added.

tests-ng/test_ssh.py
```diff
@@ -65,13 +65,13 @@ def test_sshd_has_required_config(sshd_config_item: str, sshd: Sshd):
         assert equals_ignore_case(actual_value, expected_value), f"{sshd_config_item}: expected {expected_value}, got {actual_value}"
 
 @pytest.mark.feature("ssh")
-def test_users_have_no_authorized_keys():
+def test_users_have_no_authorized_keys(expected_users):
```
Akendo: I think we make then the assumption that this parameter is *always* provided. However, I would argue that this should have the default value of `expected_users=None`. This way we do not break anything and it should be seen rather exceptional.

tests-ng/test_ssh.py
```diff
@@ -65,13 +65,13 @@ def test_sshd_has_required_config(sshd_config_item: str, sshd: Sshd):
...
     skip_users = {"nologin", "sync"}
     skip_shells = {"/bin/false"}
     files_to_check = ["authorized_keys", "authorized_keys2"]
 
     for entry in pwd.getpwall():
-        if entry.pw_name in skip_users or entry.pw_shell in skip_shells:
+        if entry.pw_name in skip_users or entry.pw_shell in skip_shells or entry.pw_name in expected_users:
```
Akendo: Please use the `any` function. Otherwise it becomes harder to maintain/read.

```
        if any([entry.pw_name in skip_users,
               entry.pw_shell in skip_shells,
               entry.pw_name in expected_users]):
```

tests-ng/util/run.sh
```diff
@@ -26,33 +45,31 @@ cd "$(realpath -- "$(dirname -- "$(realpath -- "${BASH_SOURCE[0]}")")/../")"
 
 basename="$(basename "$artifact")"
 
-extension="$(grep -E -o '(\.[a-z][a-zA-Z0-9_\-]*)*$' <<< "$basename")"
+extension="$(grep -E -o '(\.[a-z][a-zA-Z0-9_\-]*)*$' <<<"$basename")"
```
Akendo, line 48: A comment of what this regex is trying to match is welcome!

tests-ng/util/tf/modules/aws/disks.tf
```diff
@@ -12,6 +12,11 @@ resource "terraform_data" "test_disk_hash" {
 
 resource "aws_s3_bucket" "upload" {
   bucket = local.bucket_name
+
+  tags = merge(
+    local.labels,
+    { Name = local.bucket_name }
```
Akendo, line 18: One thing I wish to have is a tag to see that this was a user-specified run rather than a CI one. Presently, I couldn't tell very well if that was our testing framework. So maybe we can add to our tags more meta information like the commit ID of the branch and that this was executed from the new framework?

Akendo: My concern here is that we can clean this up later in case we might have leftovers or a person is forgetting to clean up.

tests-ng/util/tf/modules/aws/locals.tf
```diff
@@ -1,17 +1,26 @@
 locals {
-  bucket_name = "${var.prefix}-disk-upload"
-  ami_name    = "${var.prefix}-image"
-  vpc_name    = "${var.prefix}-vpc"
-  subnet_name = "${var.prefix}-subnet"
-  igw_name    = "${var.prefix}-igw"
-  rt_name     = "${var.prefix}-rt"
-  sg_name     = "${var.prefix}-sg"
-  key_name    = "${var.prefix}-ssh-key"
+  test_name   = replace(var.prefix, "_", "-")
```
Akendo, line 2: Just a suggestion since my OCD kicks in: Can we sort the variables alphabetically? Feels like I can read that more better.

tests-ng/util/tf/modules/aws/variables.tf
```diff
@@ -25,6 +25,18 @@ variable "user_data_script_path" {
   type        = string
 }
 
+variable "instance_type_amd64" {
+  description = "Default instance type for amd64"
+  type        = string
+  default     = "m5.large"
```
Akendo, line 31: Is that not too big for our testing? Do we need to use such a big instance?

tests-ng/test_ssh.py
yeoldegrove: :+1:

tests-ng/test_ssh.py
yeoldegrove: :+1:

tests-ng/util/run.sh
yeoldegrove: I personally find that the variable name explains this and the regex is not that complex. Will add a comment anyways.

tests-ng/util/tf/modules/aws/disks.tf
yeoldegrove: TLDR: no code changes needed

In the current framework we actually do that for this exact reason (identifying it later). https://github.com/gardenlinux/gardenlinux/blob/324d4e7e439727c1fae331d491dee6c53a0b6277/.github/workflows/test_platform_flavor.yml#L201

Our cleanup will never clean up things that are build up by a developer, because this ends up in a different OpenTofu state (S3 bucket) than the local state on the developer's machine.

Nevertheless to make it human identifiable, as soon as we run this in github actions (not yet the case), we can pass:

```
export TF_VAR_prefix=github-ci
tofu apply ...
```

And we can easily overwrite the default `prefix` this way.

https://github.com/gardenlinux/gardenlinux/blob/8137338563bdde890bcaa23575f1b8a9985a47b6/tests-ng/util/tf/variables.tf#L4

tests-ng/util/tf/modules/aws/locals.tf
yeoldegrove: They are grouped by names, labels, values actually. Will add some header and reorder the groups.

tests-ng/util/tf/modules/aws/variables.tf
yeoldegrove: We had problems with secureboot on smaller instance types in the past (due to initrd extraction if I remember correctly). This is what we're currently using in production and what works reliably.

tests-ng/test_ssh.py
yeoldegrove: Actually this needs to be `expected_users=[]` as we iterate it later.

tests-ng/README.md
Akendo: Solved! Great job!

tests-ng/util/login_cloud.sh
```diff
@@ -0,0 +1,25 @@
...
+image_name=${image_basename/.*/}
+workspace="$image_name"
+
+tf_dir="$(realpath -- "$(dirname -- "${BASH_SOURCE[0]}")/tf")"
+
+vm_ip="$(cd "$tf_dir" && tofu workspace select "$workspace" >/dev/null && tofu output --raw vm_ip)"
+ssh_user="$(cd "$tf_dir" && tofu workspace select "$workspace" >/dev/null && tofu output --raw ssh_user)"
+ssh_opts=(-o ConnectTimeout=5 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i ~/.ssh/id_ed25519_gl)
+
+exec ssh "${ssh_opts[@]}" "$ssh_user@$vm_ip" "$@"
```
Akendo, line 25: A general pattern I observe in bash scripts is we write them very loosely. This makes it difficult for a person from outside to understand at first what's going on. This script is simple enough, and you grasp the logic quickly. However, I would suggest that we use more functions. Starting with a main function:

```
main() {
    # Documentation what the script does goes here. 
    do_things
 }

if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
        main "$@"
fi
```

Then try putting the different logic steps into a function and let them be invoked from the `main` function. This make the bash script more maintainable in the long run.

PR #3326, opened 2025-08-21
Tests: Implement 'no failed units' testcase

fwilhe:
Excludes the tests on features where the machine does not properly boot up in qemu for various reasons, such as backend systems that are required but not available 

This is okay because those features are not made for running in a plain qemu vm

Adds an optional 'reason' field to the 'features' fixture

### DIFF
--- a/tests-ng/test_basics.py
+++ b/tests-ng/test_basics.py
@@ -58,7 +58,7 @@ def test_startup_time(systemd: Systemd):
         f"(tolerated {tolerated_userspace}s)"
     )
 
-@pytest.mark.feature("not gcp and not metal and not openstack", reason='Does not work on gcp, metal and openstack features for various reasons, for example because it cannot communicate with external backends.')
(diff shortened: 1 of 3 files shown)


tests-ng/test_basics.py
```diff
@@ -57,3 +57,15 @@ def test_startup_time(systemd: Systemd):
         f"Userspace startup too slow: {userspace:.1f}s "
         f"(tolerated {tolerated_userspace}s)"
     )
+
+@pytest.mark.feature("not gcp and not metal and not openstack", reason='Not working on various features, for example because the vm cannot communicate with external backends.')
+@pytest.mark.root(reason="Needed for journalctl which is only needed when the test fails, but still very useful for understanding test failures")
```
NotTheEvilOne: While I like the general approach of reasons I'm not sure if we add too much information / text here. Could we sum these texts up to be as short as possible? E.g.: `Not compatible, usually because of missing external backend communication channel` or `Required for journalctl in case if errors`

fwilhe: yes, I agree, this has been a bit verbose

thanks

PR #3338, opened 2025-08-26
test: migrate test_proc to new testing framework.

commel:
Migrate the `feature/base/test_proc.py` to the new testing framework.

Relates to #3181

**Definition of Done:**
- [x] Test works in chroot and is skipped everywhere else

This test requires the `/proc` to be empty. Thus we need to discuss if rebinding and mounting `/proc` from the host to the container is necessary for the chroot environment. See this line: https://github.com/gardenlinux/gardenlinux/blob/main/tests-ng/util/run_chroot.sh#L52


tests-ng/util/run_chroot.sh
```diff
@@ -49,7 +49,7 @@ mount -t tmpfs -o mode=0755 none "$tmpdir/chroot"
 
 tar --extract --xattrs --xattrs-include 'security.*' --directory "$tmpdir/chroot" < "$rootfs_tar"
 
-mount --rbind --make-rprivate /proc "$tmpdir/chroot/proc"
+#mount --rbind --make-rprivate /proc "$tmpdir/chroot/proc"
```
Akendo: I think this might be a problem here. I am not sure why this was here, but we have to check with @nkraetzschmar if that's not breaking things. There are good reasons to have procfs mounted to the chroot. It just makes no sense for the current test. This might be a challenge to get right. It might imply that we need to have different stages for testing?

Akendo: Have you tried to unmount the procfs first and remount afterwards?

tests-ng/test_proc.py
```diff
@@ -0,0 +1,19 @@
...
+
+@pytest.mark.feature("not booted")
+@pytest.mark.skipif(is_system_booted(), reason="Check on proc only in chroot")
+def test_proc_is_empty(shell: ShellRunner):
+    """ Test for an empty /proc within the given rootfs tarball """    
+
+    # Execute local command (this does not need to be
+    # performed within a platform itself) since "/proc"
+    # will never be empty on a running/used system.
+    cmd = "ls -1 /proc"
```
Akendo: Please use the `os` feature to list the dir.

tests-ng/util/run_chroot.sh
Akendo: My suggestion here would be the following: Create a setup fixture and a teardown one that ensure that the procfs is mounted. I'm uncertain if we can do this from within the container; technically, this should be possible.

tests-ng/test_proc.py
commel: Sure thing, I was trying to keep the theme when porting the test. Switched to `os.listdir` now.

tests-ng/test_proc.py
```diff
@@ -0,0 +1,18 @@
...
+
+@pytest.mark.feature("not booted")
+@pytest.mark.skipif(is_system_booted(), reason="Check on proc only in chroot")
+def test_proc_is_empty():
+    """ Test for an empty /proc within the given rootfs tarball """    
+
+    # Execute local command (this does not need to be
+    # performed within a platform itself) since "/proc"
+    # will never be empty on a running/used system.
+    proc_files: list[str] = os.listdir("/proc")
```
Akendo: I love how much more tidy this is! I would suggest another test first. You can check if the procfs was mounted beforehand with the `os.path.ismount(path)`. This appears to be  necessary to me as well to ensure that we only test when we know that the procfs wasn't mounted.

tests-ng/util/run_chroot.sh
commel: The `/proc` is mounted from the container to the chroot directory hierarchy. Umounting only works with "lazy mode", but works. But as the proc is not available in the fstab properly, I cannot mount again from within the chroot.

I've created a fixture now to move the mount point temporarily. This way we can keep the extern mount and also enable this test to look behind the mountpoint.

tests-ng/test_proc.py
commel: Added the second test.

tests-ng/plugins/shell.py
```diff
@@ -14,15 +14,16 @@ def __init__(self, user: Optional[Tuple[int, int]]):
     def __call__(
         self,
         cmd: str,
+        shell: bool = False,
```
Akendo: Since we used the python `os` modules, do we still have to set the `shell`?

Akendo: I guess we need it for the fixture, right?

tests-ng/plugins/proc.py
```diff
@@ -0,0 +1,25 @@
...
+from pathlib import Path
+
+from .shell import ShellRunner
+
+PROC_DIRECTORY="/proc"
+PROC_CHROOT_DIRECTORY="/proc_chroot"
+
+@pytest.fixture
+def transfer_proc_mount(shell: ShellRunner):
+    """Temporarily moves the proc mountpoint out of the way."""
```
Akendo: Please document a bit why you made the change, so future developers can skip out on a WTF, lowering the curse count.

tests-ng/plugins/shell.py
commel: This is not needed for this PR anymore, but I would keep it in. It allows the default value of the paramter to be changed for future changes.

tests-ng/plugins/proc.py
commel: Done

tests-ng/plugins/shell.py
Akendo: When it's not necessary, just remove it. I would rather not like to see the `shell` utilized too much and only to be used unless absolutely necessary. I would argue that this change is outside the scope of the test then. Feel free to put it into a seperated PR.

commel: In the sense of cleanliness you are absolutly right. Removed now.

tests-ng/test_proc.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+import os
+
+from plugins.booted import is_system_booted
+
+from plugins.proc import PROC_DIRECTORY, PROC_CHROOT_DIRECTORY
+
+@pytest.mark.root
+@pytest.mark.feature("not booted")
```
nkraetzschmar: the feature mark works on gardenlinux features. `booted` is a mark of itself and not a feature. this mark is therefore essentially a no-op.

nkraetzschmar (changes requested):
Instead of move mounting `/proc` somewhere else, why not just create a non-recursive bind mount of `/` somewhere? As in have a fixture that creates a temporary directory, bind mounts `/` there and returns the path of the temporary directory (with cleanup afterwards to unmount and remove temporary mount point). Then the is empty check can just operate on `{loop_mount}/proc` and check that is empty.

tests-ng/test_proc.py
commel: Good to know, I am just keeping the skip_if now.

commel:
> Instead of move mounting `/proc` somewhere else, why not just create a non-recursive bind mount of `/` somewhere? As in have a fixture that creates a temporary directory, bind mounts `/` there and returns the path of the temporary directory (with cleanup afterwards to unmount and remove temporary mount point). Then the is empty check can just operate on `{loop_mount}/proc` and check that is empty.

Good point, this is even more clean that my approach.

tests-ng/test_proc.py
```diff
@@ -0,0 +1,37 @@
+import pytest
+import os
+
+from plugins.booted import is_system_booted
+
+@pytest.mark.root
+@pytest.mark.skipif(is_system_booted(), reason="Check on proc only in chroot")
```
nkraetzschmar: why only check in chroot? with the bind mount this should work just fine in booted systems

tests-ng/test_proc.py
```diff
@@ -0,0 +1,37 @@
...
+    # Execute local file listing (this does not need to be
+    # performed within a platform itself) since "/proc"
+    # will never be empty on a running/used system.
+    proc_files: list[str] = os.listdir(temp_proc)
+
+    # Within temporary "/proc" no files or directories should be found.
+    assert len(proc_files) == 0, f"{temp_proc} is not empty."
+
+@pytest.mark.root
+@pytest.mark.skipif(is_system_booted(), reason="Check on proc only in chroot")
```
nkraetzschmar: this check should probably be done on booted only though. cause in the chroot case we're not really in control of what is or isn't in `/proc` (coming from host os), so any checks on it are not 100% consistent.

nkraetzschmar (changes requested):

tests-ng/test_proc.py
commel: The original test extracted the tarball and checked the proc-directory. But you are right, in a booted environment we should be able to check the underlying image as well.

Akendo: Moving `/proc` would have a disruptive nature on the booted system, right?

tests-ng/test_proc.py
```diff
@@ -0,0 +1,37 @@
...
+def test_container_proc_is_not_empty():
+    """
+    negative test as mounted proc should contain expected files
+    """
+    root_proc = os.path.join("/", "proc")
```
Akendo: This feels a bit unnecessary; why not just have a variable with the path to procfs?

tests-ng/test_proc.py
commel: It very much would, if we would be able to at all (device is busy e.g.). But the remount opens access to the underlying image directory structure.

tests-ng/test_proc.py
commel: Probably done too much crossplattform development, as only linux is supported, we can go for a string.

tests-ng/test_proc.py
commel: Moved to booted only.

tests-ng/test_proc.py
commel: Changed for a string.

tests-ng/test_proc.py
```diff
@@ -0,0 +1,36 @@
+import pytest
+import os
+
+from plugins.booted import is_system_booted
+
+@pytest.mark.root
+@pytest.mark.booted
```
nkraetzschmar: why is booted needed for the empty proc test?

nkraetzschmar (changes requested):

tests-ng/test_proc.py
commel: Chroot und booted differ in their environment so I made both tests (positive and negative) run in the same environment.

Should I run the test in both environments?

PR #3345, opened 2025-08-27
test: migrate test_orphan to new testing framework.

commel:
Migrate [test_orphaned.py](https://github.com/gardenlinux/gardenlinux/blob/main/features/base/test/test_orphaned.py) to new testing framework.

Refs #3181 

**Definition of Done:**

The chroot environment had not yet a proper `resolv.conf`, which made the apt-update fail. This PR adds to `run_chroot.sh` a mount that binds the `/etc/resolv.conf` from the container to the chroot environment.

A negative test could not be added yet, as from [Debian's list of orphaned packages](https://www.debian.org/devel/wnpp/orphaned_byage) I could not pick a package find a package. Maybe we need to provide a package as offline file?

**Release note**:
NONE

### DIFF
--- a/tests-ng/plugins/apt.py
+++ b/tests-ng/plugins/apt.py
@@ -19,6 +19,26 @@ class Apt:
                     repos.add(line.split()[1])
         return list(repos)
 
+    def update(self) -> bool:
+        """
+        Updates the repository cache
+        """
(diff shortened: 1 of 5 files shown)


commel:
Closed until we have a deborphan successor.

Issue #3147, opened 2025-07-02
Add container test with multiple container repositories


pnpavlov:
In order to guarantee container compatibility against multiple pull destinations, please add a test covering pulling and starting a basic container from
- [ ] GitHub Container Registry
- [ ] Google Container Registry
- [ ] Docker Hub
- [ ] jFrog

other that you identify as relevant. Test case can be
pull latest debian stable container from all those sources and execute simple command in the container, confirming the result is as expected

vivus-ignis:
PR: https://github.com/gardenlinux/gardenlinux/pull/3348

PR #3348, opened 2025-08-29
basic tests for containerd

vivus-ignis:

Fixes https://github.com/gardenlinux/gardenlinux/issues/3147

Running tests that involve containerd requires passing this parameter to the test runner: `--test-args --default-user=root`

### DIFF
--- a/tests-ng/plugins/containerd.py
+++ b/tests-ng/plugins/containerd.py
@@ -14,7 +14,7 @@ class CtrRunner:
         self.shell = shell
         systemd.start_unit("containerd")
 
-    def pull(self, uri, capture_output=False, ignore_exit_code=False):
+    def pull_image(self, uri, capture_output=False, ignore_exit_code=False):
         validators.url(uri)
         command = f"ctr image pull {uri}"
         return self.shell(command, capture_output=capture_output, ignore_exit_code=ignore_exit_code)
--- a/tests-ng/test_containers.py
+++ b/tests-ng/test_containers.py
@@ -10,7 +10,7 @@ TEST_IMAGES = [
 
 @pytest.fixture
 def container_image_setup(uri: str, ctr: CtrRunner):
-    ctr.pull(uri)
+    ctr.pull_image(uri)
     yield
     ctr.remove_image(uri)
 


tests-ng/test_containers.py
```diff
@@ -0,0 +1,27 @@
...
+def containerd(shell: ShellRunner):
+    shell("systemctl start containerd", capture_output=False)
+    yield
+    shell("systemctl stop containerd", capture_output=False)
+
+
+def test_basic_container_functionality(shell: ShellRunner, containerd):
+    for uri in TEST_IMAGES:
+        pull_command = f"ctr image pull {uri}"
+        shell(pull_command, capture_output=False)
```
nkraetzschmar: instead of using a shell here it might be cleaner to have the `containerd` fixture return some object abstracting away these operations as functions

tests-ng/test_containers.py
```diff
@@ -0,0 +1,27 @@
...
+TEST_IMAGES = [
+    "ghcr.io/gardenlinux/gardenlinux:1592.12",    # Github Container Registry, https://github.com/orgs/gardenlinux/packages?ecosystem=container
+    "docker.io/library/debian:12.11-slim",        # Docker Hub, https://hub.docker.com/_/debian
+    "public.ecr.aws/debian/debian:11",            # AWS ECR, https://gallery.ecr.aws/debian/debian
+]
+
+
+@pytest.fixture
+def containerd(shell: ShellRunner):
+    shell("systemctl start containerd", capture_output=False)
```
commel: I would suggest using the systemd-fixture: https://github.com/gardenlinux/gardenlinux/blob/main/tests-ng/plugins/systemd.py#L51

tests-ng/test_containers.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+from plugins.shell import ShellRunner
+
+TEST_IMAGES = [
+    "ghcr.io/gardenlinux/gardenlinux:1592.12",    # Github Container Registry, https://github.com/orgs/gardenlinux/packages?ecosystem=container
```
commel: Whenever we want to change the image versions, we have to change source code for that. We should move these to some configuration file or a fixture providing these values.

vivus-ignis: Replaced static tags with ":latest".

tests-ng/test_containers.py
vivus-ignis: Fixed

tests-ng/test_containers.py
vivus-ignis: Fixed, added `plugins/containerd.py`.

tests-ng/plugins/containerd.py
```diff
@@ -0,0 +1,22 @@
...
+from .systemd import Systemd
+
+
+class CtrRunner:
+    def __init__(self, shell: ShellRunner, systemd: Systemd):
+        self.shell = shell
+        systemd.start_unit("containerd")
+
+    def pull(self, uri):
+        command = f"ctr image pull {uri}"
```
nkraetzschmar, line 19: The uri part should be properly quoted / escaped to prevent shell splitting. I'd suggest checking no `'` characters are in the uri (throwing an exception otherwise), and then `'` quoting the uri => we also prevent any shell expansion or things like that.

tests-ng/plugins/containerd.py
```diff
@@ -0,0 +1,22 @@
...
+        self.shell(command, capture_output=False)
+
+    def run(self, uri, container_name, cmd, capture_output=False, ignore_exit_code=False):
+        command = f"ctr run --rm {uri} {container_name} {cmd}"
```
nkraetzschmar, line 31: See comment on quoting / escaping strings above

tests-ng/test_containers.py
```diff
@@ -0,0 +1,18 @@
...
+
+TEST_IMAGES = [
+    "ghcr.io/gardenlinux/gardenlinux:latest", # Github Container Registry, https://github.com/orgs/gardenlinux/packages?ecosystem=container
+    "docker.io/library/debian:latest",        # Docker Hub, https://hub.docker.com/_/debian
+    "public.ecr.aws/debian/debian:latest",    # AWS ECR, https://gallery.ecr.aws/debian/debian
+]
+
+
+def test_basic_container_functionality(ctr: CtrRunner):
+    for uri in TEST_IMAGES:
```
nkraetzschmar: Rather then looping, this might benefit from `@pytest.mark.parametrize` similar to how it's done in https://github.com/gardenlinux/gardenlinux/blob/24ab1bcb0d4424ff572eb42b3a5b406d7ea7ef5f/tests-ng/test_packages.py#L5-L16

vivus-ignis: Fixed.

tests-ng/plugins/containerd.py
vivus-ignis: Rather than writing (any) custom validation -- which has to be tested as well -- I've added the "validators" package and respective calls to validate image URIs. Now if anything but a correct URI is passed to CtrRunner, it will raise an exception.

tests-ng/plugins/containerd.py
vivus-ignis: Added uri validation.

tests-ng/plugins/containerd.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+import validators
+from .shell import ShellRunner
+from .systemd import Systemd
+
+
+class CtrRunner:
+    def __init__(self, shell: ShellRunner, systemd: Systemd):
```
Akendo, line 13: Please provide a docstring for documentation.

vivus-ignis: Fixed.

tests-ng/plugins/containerd.py
nkraetzschmar: either the container name also must be validated, or the container name should be derived from the uri here rather then being an input

nkraetzschmar (changes requested):

PR #3358, opened 2025-09-01
Add support to run GitHub workflow tests staged again

NotTheEvilOne:
This PR provides support to run GitHub workflow tests staged again. This means that subsequent tests like platform tests are only executed if chrooted ones are completed.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,6 +27,10 @@ on:
         description: 'Execute bare flavors container tests'
         type: boolean
         default: true
+      qemu_test:
+        description: 'Execute QEMU tests'
+        type: boolean
+        default: true
     secrets:
       gcp_identity_provider:
         required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -134,10 +143,17 @@ jobs:
       arch: ${{ matrix.arch }}
       flavor: ${{ matrix.flavor }}
   test_flavors_qemu:
-    needs: intersect_matrices
+    needs: [ intersect_matrices, test_flavors_chrooted ]
```
nkraetzschmar, line 146: I'm not sure I see the advantage in waiting for the chroot tests before running the QEMU test.
Both are running on free github runners, so unlike with the platform tests we do not safe any money from skipping the QEMU tests. However, enforcing an ordering here will mean that the total runtime is going to suffer due to decreased parallelism.

.github/workflows/tests.yml
```diff
@@ -146,10 +162,17 @@ jobs:
       arch: ${{ matrix.arch }}
       flavor: ${{ matrix.flavor }}
   test_platform_flavors:
-    needs: intersect_matrices
+    needs: [ intersect_matrices, test_flavors_chrooted ]
```
nkraetzschmar, line 165: I'd suggest also making the platform tests dependent on the QEMU tests. That way the QEMU tests can already catch any issues preventing machines from booting at all => we don't waste resources on starting cloud VMs that will never even boot.

nkraetzschmar (changes requested):

PR #3369, opened 2025-09-04
Tests: Migrate `features/base/test/test_network` to new testing framework

ByteOtter:

Migrates the tests from `features/base/test/test_network.py` to a new file for our new testing framework. The original test relied on external connections to the general internet being available. These were replaced by tests that work with our contained environment.

Fixes: [#3438](https://github.com/gardenlinux/gardenlinux/issues/3438)
Tracks https://github.com/gardenlinux/gardenlinux/issues/3181

**Definition of Done:**

### DIFF
--- a/tests-ng/test_network.py
+++ b/tests-ng/test_network.py
@@ -34,7 +34,7 @@ def test_local_tcp_stack(ip_version, loopback, tcp_echo_server):
     """
     # Arrange
     msg = b"Hello, TCP!"
-    result = tcp_echo_server(ip_version, loopback, msg)
+    result, done = tcp_echo_server(ip_version, loopback, msg)
 
     # Wait until the server bound to a port
     while "port" not in result:
@@ -45,6 +45,7 @@ def test_local_tcp_stack(ip_version, loopback, tcp_echo_server):
         conn.sendall(msg)
 
     # Assert
+    done.wait(timeout=2)
     assert result.get("data") == msg
 
 
--- a/tests-ng/test_network.py
+++ b/tests-ng/test_network.py
@@ -1,0 +1,129 @@
+import datetime
+import socket
+import threading
+import pytest
+
+
+def has_ipv6():
+    """Helper function to detect IPv6 availability."""
+    try:
+        s = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+        s.close()
+        return True
+    except OSError:
+        return False
+
+
+# Test parameters. IPv6 skipped if not supported.
+LOCAL_TEST_PARAMS = [
+    pytest.param(socket.AF_INET, "127.0.0.1"),
+    pytest.param(
+        socket.AF_INET6,
+        "::1",
+        marks=pytest.mark.skipif(not has_ipv6(), reason="IPv6 not available"),
+    ),
+]
+
+
+@pytest.mark.booted(reason="ping requires full network stack")
+def test_loopback_interface(shell):
+    """Assert loopback device present."""
+    result = shell("ping -c 1 127.0.0.1", capture_output=True)
+    assert result.returncode == 0
+    assert "1 received" in result.stdout
+
+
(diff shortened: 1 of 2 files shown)


tests-ng/test_network.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # Clean
+    client.close()
+    server.close()
+
+
+@pytest.mark.booted
+def test_resolv_conf_exists(shell):
+    """Test if local DNS config is available."""
+    result = shell("test -s /etc/resolv.conf", ignore_exit_code=True)
```
fwilhe: I think we wanted to use native python apis over shell in cases where there is no really good reason to use shell, here I think `os.path.isfile(fname)` (or some newer api in python?) would be easier to read. I had to look up what the `-s` argument for `test` means, and we also want to make this readable to non-linux gurus as far as possible.

tests-ng/test_network.py
```diff
@@ -0,0 +1,129 @@
+import datetime
+import socket
+import threading
+import pytest
+
+
+def has_ipv6():
```
fwilhe: helper functions should go into a plugin imo

tests-ng/test_network.py
```diff
@@ -0,0 +1,129 @@
...
+
+@pytest.mark.booted(reason="ping requires full network stack")
+def test_loopback_interface(shell):
+    """Assert loopback device present."""
+    result = shell("ping -c 1 127.0.0.1", capture_output=True)
+    assert result.returncode == 0
+    assert "1 received" in result.stdout
+
+
+@pytest.mark.parametrize("family,loopback", LOCAL_TEST_PARAMS)
```
fwilhe: 'family' is not self explanatory, I understand this means 'ip family', maybe call it ip_version' or 'protocol_family'?

tests-ng/test_network.py
```diff
@@ -0,0 +1,129 @@
...
+            s.bind((loopback, 0))
+            port = s.getsockname()[1]  # Get automatically chosen port
+            result["port"] = port
+            s.listen(1)
+            conn, _ = s.accept()
+            data = conn.recv(1024)
+            result["data"] = data
+
+    thread = threading.Thread(target=server, daemon=True)
+    thread.start()
```
fwilhe: can those details be abstracted in a plugin? when reading the test, it should not matter how exactly the internal server is constructed, but only that we have an internal server

tests-ng/test_network.py
```diff
@@ -0,0 +1,129 @@
...
+    client.sendto(b"ping", (addr, port))
+    data, src = server.recvfrom(1024)
+    assert data == b"ping"
+
+    server.sendto(b"pong", src)
+    data, _ = client.recvfrom(1024)
+    assert data == b"pong"
+
+    # Clean
+    client.close()
```
fwilhe: this could be a fixture like in this sample, right? https://github.com/gardenlinux/gardenlinux/blob/817b42b1938c823a824b1bb688bab70717e1a647/tests-ng/test_containers.py#L12

ByteOtter: Oh yeah completely valid point. I can for sure change this into a yielding fixture :)

tests-ng/test_network.py
ByteOtter: I agree. Done wit the next push. 👍🏼

tests-ng/test_network.py
ByteOtter: Got it. Will update with next push.

tests-ng/test_network.py
ByteOtter: Moved this and the UDP version into a yielding fixture that takes care of creating a destroying the connection.

tests-ng/test_network.py
ByteOtter: Applied your suggestion with `os.path.isfile` and `os.path.getsize` to mimic the behaviour of `test -s` in a more explicit way. Thanks for the suggestion.

Issue #3414, opened 2025-09-08
Fix flakey behaviour of 'no failed units' test in the new test framework


fwilhe:
We are observing 'random' failures of the 'no failed units' test case as in this examples

- https://github.com/gardenlinux/gardenlinux/actions/runs/17433873246/job/49499961589
- https://github.com/gardenlinux/gardenlinux/actions/runs/17424917320/job/49471025704
- https://github.com/gardenlinux/gardenlinux/actions/runs/17408106560/job/49419518315

This did not occur while I was working on the tests themselves, so it seems like the problem is not easily reproducible on purpose.

After a call with @5kt, I think we can make this more resilient with two changes:

1. Use [`systemctl is-system-running --wait`](https://manpages.debian.org/trixie/systemd/systemctl.1.en.html#System_Commands) to wait until all activating units are actually started

2. Use `systemctl is-failed` instead of our manual filtering for failed units; this should avoid false positives for non-failed statuses like 'activating'

PR #3416, opened 2025-09-08
fix: Make 'no failed units' test more resilient by waiting for the system to be booted

fwilhe:

This should fix 'random' test failures by waiting for systemd to report the system is 'running', and by using systemd's check for failed units.

Fixes https://github.com/gardenlinux/gardenlinux/issues/3414

**Definition of Done:**

**Release note**:

```feature user

```

### DIFF
--- a/tests-ng/plugins/systemd.py
+++ b/tests-ng/plugins/systemd.py
@@ -53,7 +53,6 @@ class Systemd:
             pytest.skip("starting units is only supported when system state modifications are allowed")
         self._shell(f"systemctl start {unit_name}")
 
-
     def list_units(self) -> list[SystemdUnit]:
         result = self._shell(f"{self._systemctl}", capture_output=True, ignore_exit_code=True)
         units = []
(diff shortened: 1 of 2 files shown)


tests-ng/plugins/systemd.py
```diff
@@ -53,14 +60,20 @@ def start_unit(self, unit_name: str):
...
-            parts = line.split()
-            units.append(SystemdUnit(parts[0], parts[1], parts[2], parts[3]))
-        return units
+        return _parse_units(result.stdout)
+
+    def list_failed_units(self) -> list[SystemdUnit]:
+        result = self._shell(f"{self._systemctl} --failed", capture_output=True, ignore_exit_code=True)
+        return _parse_units(result.stdout)
+
+    def wait_is_system_running(self) -> Tuple[str, int, float]:
```
commel: Maybe a dataclass as return value would be more descriptive on what these values mean.

tests-ng/plugins/systemd.py
```diff
@@ -53,14 +60,20 @@ def start_unit(self, unit_name: str):
...
         result = self._shell(f"{self._systemctl}", capture_output=True, ignore_exit_code=True)
-        units = []
-        for line in result.stdout.splitlines():
-            parts = line.split()
-            units.append(SystemdUnit(parts[0], parts[1], parts[2], parts[3]))
-        return units
+        return _parse_units(result.stdout)
+
+    def list_failed_units(self) -> list[SystemdUnit]:
+        result = self._shell(f"{self._systemctl} --failed", capture_output=True, ignore_exit_code=True)
```
commel, line 82: I wonder if using `-o json` as output would be a bit more resilient against displayed changes. Also it might be easier to parse as well.

fwilhe: I was considering this before, but I think I did not use it because not all commands support it. 

Screenshot 2025-09-10 at 11 13 37

You might argue that it is still better to use it, and for parsing the unit states I agree this is more clean.

PR #3453, opened 2025-09-15
lima with vmtype:vz ready

### LINKED ISSUE #3430: Garden Linux lima image works with apple vz

saubhikdattagithub:
Modify `features/lima` for possibility to deploy lima in GardenLinux with both `vmtype: vz` and `vmtype: qemu`

Fixes #3430 

**Definition of Done:**
- [X] The code is appropriately tested

image

### DIFF
--- a/features/lima/exec.config
+++ b/features/lima/exec.config
@@ -18,9 +18,5 @@ mounts:
 mountTypesUnsupported: [9p]
 EOF
 
-# growpart is done in initramfs, growroot by systemd
-mv /etc/cloud/cloud.cfg /etc/cloud/cloud.cfg.bak
-cat /etc/cloud/cloud.cfg.bak | grep -v "^ - growpart$" | grep -v "^ - resizefs$" | grep -v "^ - ntp$" >/etc/cloud/cloud.cfg
-rm /etc/cloud/cloud.cfg.bak
 
 


features/lima/info.yaml
```diff
@@ -2,7 +2,8 @@ description: "image suitable for using with [lima](https://lima-vm.io)"
 type: flag
```
fwilhe:
```suggestion
type: platform
```

we should have always one 'platform' feature, previously this was kvm, but now as we replace it the 'lima' feature needs to be of that type.

features/lima/pkg.include
```diff
@@ -1,3 +1,11 @@
 cloud-init
 sshfs
 openssh-server
+dosfstools
+qemu-guest-agent
+efibootmgr
+efitools
+linux-image-$arch
+tpm2-tools
+ipmitool
```
fwilhe: I don't think we need that, this seems to come from the metal feature and have no application in vms.

features/lima/pkg.include
```diff
@@ -1,3 +1,11 @@
 cloud-init
 sshfs
 openssh-server
+dosfstools
+qemu-guest-agent
+efibootmgr
+efitools
+linux-image-$arch
```
fwilhe: any reason not to use the cloud variant of the kernel?

features/lima/exec.config
```diff
@@ -1,14 +1,26 @@
...
+  loadDotSSHPubKeys: true
+  forwardAgent: true
+mounts:
+  - location: "~"
+    writable: false
 mountTypesUnsupported: [9p]
 EOF
+
+# growpart is done in initramfs, growroot by systemd
+mv /etc/cloud/cloud.cfg /etc/cloud/cloud.cfg.bak
```
fwilhe: is this still needed? what's the reasoning?

saubhikdattagithub: yes, not needed anymore after cloud.cfg removal

PR #3469, opened 2025-09-17
Tests: Migrate test_password_hashes to new framework

ByteOtter:

Migrate the `base/test/test_password_hashes.py` test to a new testing module in accordance to the new testing framework.

Fixes: https://github.com/gardenlinux/gardenlinux/issues/3468
Tracks: #3181 

**Definition of Done:**


tests-ng/test_password_hashes.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Ensure that exactly one password entry with [success=... default=ignore]
+    exists in PAM config.
+    """
+    assert (
+        len(pam_candidates) == 1
+    ), f"Expected exactly one password entry, found {len(pam_candidates)}: {pam_candidates}"
+
+
+@pytest.mark.security_id(325)
```
Akendo, line 18: Does this will be seen in the output already?

tests-ng/plugins/password_hashes.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@pytest.fixture
+def pam_common_password_text():
+    """
+    Read the raw text from /etc/pam.d/common-password file.
+    """
+    path = Path("/etc/pam.d/common-password")
+    assert path.exists(), f"'{path}' is missing"
+    return path.read_text(encoding="utf-8", errors="ignore")
```
Akendo: I do not like how we're doing this. I feel like the `pam.d` files are generic enough that we could put that into a class of its own. You have to keep in mind that the `common-password` supports more arguments, and the goal is to ensure that we do not have parameters we deem to be invalid. 

This code will return just a set of raw data that you still have to filter.

tests-ng/test_password_hashes.py
ByteOtter: I'm sorry I can't seem to follow. What will be seen in the output already?

Akendo: The new testing framework does not support the `mark.security_id` yet. Not sure if this work well?

tests-ng/test_password_hashes.py
```diff
@@ -0,0 +1,32 @@
...
+def test_password_entry_uses_strong_hash(pam_candidates):
+    """
+    Ensure that the password entry uses a strong hash algorithm (yescrypt or sha512).
+    """
+    assert (
+        len(pam_candidates) == 1
+    ), "No valid password entry to check for hash algorithm"
+    test_line = pam_candidates[0]
+
+    has_yescrypt = re.search(r"\byescrypt\b", test_line, re.IGNORECASE)
```
Akendo: The code we had in the original code was much clearer.

tests-ng/test_password_hashes.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    assert (
+        len(pam_candidates) == 1
+    ), f"Expected exactly one password entry, found {len(pam_candidates)}: {pam_candidates}"
+
+
+@pytest.mark.security_id(325)
+def test_password_entry_uses_strong_hash(pam_candidates):
+    """
+    Ensure that the password entry uses a strong hash algorithm (yescrypt or sha512).
```
Akendo, line 21: We had some comments in the old test that are important to keep.

tests-ng/test_password_hashes.py
ByteOtter: Ah yes now I know what you mean. I have to admit that I just carried that over from the old tests. It doesn't influence the test's behaviour and as long as we don't have a `plugin` to collect it, it's essentially a no-op.
I kept it because maybe might carry the original handling over to the new framework?

tests-ng/plugins/password_hashes.py
ByteOtter: So you would prefer a wrapper object that abstracts the parsing into itself and returns only valid entries?

tests-ng/test_password_hashes.py
ByteOtter: Okay valid, I can change it back to the original

```python
assert "yescrypt" in test_line or "sha512" in test_line, ...
```

tests-ng/test_password_hashes.py
ByteOtter: Right, that's what I forgot to copy, thanks for the catch!

Akendo:
This code will be more dicey to begin with, since we have to validate that our PAM configurations are working as intended. Additionally, we need to validate the input others might have done on PAM as well. That implies we need a clean abstraction of the existing PAM configuration that allows us to inspect the given state correctly. That also implies that the task is much more effort to 'move' this test here. 

Since we might deal with systems that have deviating configurations from our own. This implies that this will be security-critical code. It's one of the few security tests we migrate; however, those tested back then were directed towards our systems only and not with the idea that others might have modified them.

ByteOtter:
> This code will be more dicey to begin with, since we have to validate that our PAM configurations are working as intended. Additionally, we need to validate the input others might have done on PAM as well. That implies we need a clean abstraction of the existing PAM configuration that allows us to inspect the given state correctly. That also implies that the task is much more effort to 'move' this test here.
> 
> Since we might deal with systems that have deviating configurations from our own. This implies that this will be security-critical code. It's one of the few security tests we migrate; however, those tested back then were directed towards our systems only and not with the idea that others might have modified them.

Okay so I would propose this: A `dataclass` that just represents each individual entry as a set of fields. And a `PamConfig` class that, if provided with a config file, can parse the file, collect all entries and return any entry we wish for by exposing a search function. This way we can really dissect the individual entries rather than going the regex route. What do you think?

tests-ng/plugins/password_hashes.py
ByteOtter: I have now rewritten the plugin for this code completely. Now we have only one fixture that returns a `PamConfig` object that allows us to search for entries based on on search parameters.

tests-ng/test_password_hashes.py
ByteOtter: I did change it back to something more resembling the original logic for now. Though I did also lay the groundwork to use a `hash_algo` property with the new `PamEntry` dataclass in case we want to go that route. I can remove/change whatever we choose then.

ByteOtter:
I have rewritten the plugin extensively to give us the ability to filter for entries and expand our options in dealing with PAM entries later on. The tests seem to work correctly now as well.
I would be in favour of changing the name of the plugin from `password_hashes.py` to `pam.py` or something similar to indicate that the plugin can be used by more than one set of tests and has more generalized responsibilities.

Looking forward to more feedback.

ByteOtter:
> I think this is great progress! Keep in mind we need to test for more configuration aspects within PAM. This breaks the scope of the overall test right now. I suggest using this test as an MVP for the respected test. However, we need this specific test since it's to ensure [compliance](https://www.tenable.com/audits/items/CIS_Distribution_Independent_Linux_Workstation_L1_v2.0.0.audit:2ecda5602069de22f53cf45659f91103).

Thanks for the input. I am unsure which other configurations we need to include in this specific test. I think as a simple port of the old tests, this is sufficient?
I could also expand on it if you wish to cover more ground in this PR.

tests-ng/plugins/password_hashes.py
```diff
@@ -0,0 +1,212 @@
...
+    :param control: raw control token (either a single token like 'required'
+                    or bracketed expression like '[success=1 default=ignore]')
+    :type control: str
+    :param module: module name, e.g. 'pam_unix.so'
+    :type module: str
+    :param options: list of module arguments (tokens after the module name)
+    :type options: List[str]
+    """
+
+    type_: str
```
Akendo, line 37: Why do we use `type_`? Is it because of Python?

tests-ng/plugins/password_hashes.py
```diff
@@ -0,0 +1,212 @@
...
+        for opt in self.options:
+            if opt in SUPPORTED_HASH_ALGORITHMS:
+                return opt
+        return None
+
+    def __repr__(self) -> str:
+        return f"PamEntry(type_={self.type_!r}, control={self.control!r}, module={self.module!r}, options={self.options!r})"
+
+
+class PamConfig:
```
Akendo, line 95: Some documentation would be nice.

tests-ng/plugins/password_hashes.py
```diff
@@ -0,0 +1,212 @@
...
+    @property
+    def control_dict(self) -> Dict[str, str]:
+        """
+        Parse bracketed control expressions into a dict.
+        Example:
+            '[success=1 default=ignore]' -> {'success': '1', 'default': 'ignore'}
+        If control is a simple token (e.g. 'required'), return {}
+        """
+        control_line = self.control.strip()
+        if control_line.startswith("[") and control_line.endswith("]"):
```
Akendo: This code might run into problems since PAM allows you to use the `[` and `]` more than once in an expression. You can escape the `]` with a backspace. 

```
[..[..\]..]    -->   ..[..]..
```

tests-ng/plugins/password_hashes.py
```diff
@@ -0,0 +1,212 @@
...
+
+SUPPORTED_HASH_ALGORITHMS = ["yescrypt", "sha512"]
+
+
+@dataclass
+class PamEntry:
+    """
+    Representation of one non-comment PAM configuration line.
+
+    Example:
```
Akendo, line 16: We should also include the `@` since it's possible that a service configuration will utilize a different configuration file. For instance, the service for the `passwd` will point with the `@` symbol to the `@include common-password`.

tests-ng/plugins/password_hashes.py
```diff
@@ -0,0 +1,212 @@
...
+    :type options: List[str]
+    """
+
+    type_: str
+    control: str
+    module: str
+    options: List[str]
+
+    @property
+    def control_dict(self) -> Dict[str, str]:
```
Akendo, line 44: I wonder if the usage of the Python [`parse`](https://pypi.org/project/parse/) module is a more desirable strategy to deal with the parsing problem here?

tests-ng/plugins/password_hashes.py
ByteOtter: Yeah the underscore is necessary to avoid confusion with the `type` keyword.

tests-ng/plugins/password_hashes.py
Akendo: Added it to the wrong line.

tests-ng/plugins/password_hashes.py
```diff
@@ -0,0 +1,212 @@
...
+        ]
+        self.entries: List[PamEntry] = self._parse_entries()
+
+    def _parse_entries(self) -> List[PamEntry]:
+        """
+        Parse non-empty, non-comment lines into PamEntry objects.
+
+        Handles bracketed control expressions that may contain spaces,
+        e.g. '[success=1 default=ignore]'.
+        """
```
Akendo, line 144: I wonder if the usage of the Python [`parse`](https://pypi.org/project/parse/) module is a more desirable strategy to deal with the parsing problem here?

tests-ng/plugins/password_hashes.py
ByteOtter: Okay I now updated the `PamEntry` dataclass to have a field `include_target: Optional[str]` that is set on lines starting with `@include`. I also updated the filter functions accordingly.

tests-ng/plugins/password_hashes.py
ByteOtter: I will use a regex to find all bracketed expressions then and fold them into the same dictionary.

tests-ng/plugins/password_hashes.py
ByteOtter: I would say it's not worth it here. As  far as I understand the `parse` module, it relies heavily on predictable structures so we would have to manually handle anything aside from that manually anyway. Also for this usecase alone, I do not think it is worth it to pull in another outside dependency. Of course if we had more use cases for parse, that would be a different story.

ByteOtter: I will try to rewrite the function using parse and see if maybe the readability improvements would justify it and share the resulting code here if you like.

tests-ng/plugins/password_hashes.py
ByteOtter: Just added proper documentation. 🫡

tests-ng/plugins/password_hashes.py
ByteOtter: So, a reimplementation of that function with `parser` would look somewhat like this:

```python
def _parse_entries(self) -> List[PamEntry]:

    entries: List[PamEntry] = []
    
    for line in self.lines:
       stripped = line.lstrip()
       if not stripped or stripped.starswith("#"):
           continue
       
       include_target = parse("@include {include_target}", stripped)
       if include_target:
           entries.append(
               PamEntry(
                   type_ = "",
                   control = "",
                   module = "",
                   options = [],
                   include_target = include_target["include_target"],
               )
           )
           continue
       
       
       control_tokens = parse("{type_} {control} {module} {options}", stripped)
       if control_tokens:
           # Options may be multiple; split on spaces
           options = control_tokens["options"].split() if control_tokens["options"] else []
           entries.append(...)
           continue
       
       # Fallback: parse without control tokens
       result = parse("{type_} {module} {options}", stripped)
       if result:
           # Split options on spaces like before
           options = control_tokens["options"].split() if control_tokens["options"] else []
           entries.append(
               PamEntry(
                   type_ = result["type_"],
                   control = "",
                   module = result["module"],
                   options = options,
              )
          )
          continue
     
         # if no match, continue
         continue
     return entries
```
I guess it looks a little bit more tidy, but I personally don't think it's worth it. If you feel differently, I am open to discuss this change further.

fwilhe: > but I personally don't think it's worth it

Agree, I would vote for being very conservative with adding external dependencies to the test framework. I don't know this module and I don't know how well it is maintained, but overall I would like to keep the list of non-stdlib dependencies small. Of course there will be cases where the benefits outweigh the cost by a wide margin (like with the boolean algebra library), but this does not look like that to me.

ByteOtter: Yeah I agree

Akendo: I disagree; one reason for this refactoring was to **not** have to code all the functions ourselves but resort to solutions that make it easier to implement. We have to write basically a parser for more than just PAM, but for any configuration, we will need to validate properly`. In general, I want to see the use of modules helping make our lives easier. But I guess, for the tests we have, this is good enough. 

However, I think this isn't the spot to discuss this overall point of what Praser is suited for best.

ByteOtter: If we were to have more use cases for config parsijg logic I agree we should find a well maintained external dependency.
I agree that we should tlak about the general topic, maybe next week? Maybe we can create a proper dependency policy in our documentation. Similar to this one:
https://the-nazara-project.github.io/Nazara/contributors/dependencies.html

PR #3475, opened 2025-09-18
fix(usi): use discard only if supported

5kt:

Fixes #


pnpavlov (approved):
OK and tested

gehoern (approved):

PR #3501, opened 2025-09-25
Move "Test Distribution Build" to the correct GH workflow stage

NotTheEvilOne:
This PR moves the `Build Test Distribution` workflow to the GitHub build workflow stage. Building and testing should be separated and isolated workflows for easier maintenance and separation of tasks.


yeoldegrove (approved):
Makes sense to me.

Issue #3482, opened 2025-09-22
test-ng: fetching EDK2 randomly fails


yeoldegrove:
**What happened:**

Fetching EDK2 randomly fails in tests-ng. This should be build with caching or some retry logic.

**Debug Logs:**

https://github.com/gardenlinux/gardenlinux/actions/runs/17906087455/job/50909135984

```
Run ./test-ng ".build/$CNAME.tar"
🛠️  building python runtime
  WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProtocolError('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))': /packages/a8/a4/20da314d277121d6534b3a980b29035dcd51e6744bd79075a6ce8fa4eb8d/pytest-8.4.2-py3-none-any.whl.metadata
🛠️  building test framework distribution
🛠️  bundling test framework as disk image
Formatting '.build/dist.ext2.raw', fmt=raw size=1073741824
⬇️  fetching EDK2 (qemu-x86_64-code)
⬇️  fetching EDK2 (qemu-x86_64-vars)
⬇️  fetching EDK2 (qemu-aarch64-code)
⬇️  fetching EDK2 (qemu-aarch64-vars)
curl: (35) Recv failure: Connection reset by peer
make: *** [util/build.makefile:34: .build/edk2-qemu-aarch64-vars] Error 35
make: *** Deleting file '.build/edk2-qemu-aarch64-vars'
```

PR #3502, opened 2025-09-25
3482 test ng fetching edk2 randomly fails

vivus-ignis:
This enables retries on all curl calls used by the tests-ng framework

Fixes https://github.com/gardenlinux/gardenlinux/issues/3482

### DIFF
--- a/tests-ng/plugins/utils.py
+++ b/tests-ng/plugins/utils.py
@@ -3,7 +3,7 @@ import pytest
 # Various utility functions to make tests more readable
 # This should not contain test-assertions, but only abstract details that make tests harder to read
 
-def equals_ignore_case(a: str, b: str) -> str:
+def equals_ignore_case(a: str, b: str) -> bool:
     return a.casefold() == b.casefold()
 
 def get_normalized_sets(*sets: set) -> tuple[set, ...]:


.github/workflows/test_flavor_qemu.yml
```diff
@@ -65,6 +65,13 @@ jobs:
           rm "$CNAME.tar.gz"
           tree .build
         # TODO: handle skipping better
+      - name: Restore EDK2 cache
+        id: tests-ng-edk2-cache-restore
+        uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # pin@v4.3.0
+        with:
+          path: 
+            .build/edk2-*
+          key: tests-ng-edk2-${{ hashFiles('tests-ng/util/build.makefile') }}
```
yeoldegrove: I do not think we need this.

https://github.com/gardenlinux/gardenlinux/pull/3502/files#diff-522b408fac8deafb9e5429fc5a54cafb9697645fd31bfb18852f5cab16752cfcR39-R43 already puts the test distribution in the correct path `tests-ng/.build`.

Restoring this should actually already cache the `edk2-*` files.

.github/workflows/test_flavor_qemu.yml
```diff
@@ -99,6 +106,13 @@ jobs:
             echo "no raw or pxe file, QEMU test failed"
             exit 1
           fi
+      - name: Save EDK2 cache
+        id: tests-ng-edk2-cache-save
+        uses: actions/cache/save@0057852bfaa89a56745cba8c7296529d2fc39830 # pin@v4.3.0
+        with:
+          path: 
+            .build/edk2-*
+          key: tests-ng-edk2-${{ hashFiles('tests-ng/util/build.makefile')}}
```
yeoldegrove: I do not think we need this.

https://github.com/gardenlinux/gardenlinux/pull/3502/files#diff-522b408fac8deafb9e5429fc5a54cafb9697645fd31bfb18852f5cab16752cfcR39-R43 already puts the test distribution in the correct path `tests-ng/.build`.

Restoring this should actually already cache the `edk2-*` files.

yeoldegrove (changes requested):
If I understand it correctly, caching should already be enabled on the test site (`test_falvor_qemu.yml`) because the whole `tests-ng/.build` directory is restored as an artifact.

`.github/workflows/build_test_dist.yml` on the other hand is not yet utilize the cache in `tests-ng/.cache` (currently only used for `cpython-*.tar.gz`) and the `.cache` could even be extended for `edk2-*` files.
This would be a neat feature I guess and would make way less curl calls to retrieve cpython and edk2.

Please also rebase the branch.

.github/workflows/test_flavor_qemu.yml
vivus-ignis: Removed.

.github/workflows/test_flavor_qemu.yml
vivus-ignis: Agreed, removed.

PR #3506, opened 2025-09-25
test-ng: compare system state before/after tests ran

### LINKED ISSUE #3488: test-ng: compare system state before/after tests ran

I would like to have a mechanism to snapshot/diff the "state" of the system under test, before and after our test suite ran. By "state" I mean important system settings like:
  - installed packages
  - running services
  - loaded kernel modules
  - ...

I would design it like this:
- add `plugins/sysdiff.py` that collects snapshots of a system's state
- add `test_sysdiff.py`
  - add new marker `@pytest.mark.first` to make sure 1st snapshot is run at the very beginning
  - add new marker `@pytest.mark.last` to make sure 2nd snapshot and diff are run at the very end

**Why is this needed**:

When reviewing tests I noticed that it can get hard keeping track of what modifications to the system state are actually done.
We already have `@pytest.mark.modify` but no way of enforcing/checking that the state after the tests is the same.

yeoldegrove:

This PR adds the sysdiff plugin and tests that run before/after all other tests. To achieve this it uses the pytest-order plugin.

The sysdiff plugin creates snapshots (stored in `~/.local/state/sysdiff/` ) of ...
  - installed packages
  - systemd unit states
  - file checksums
  - sysctl parameters
  - kernel modules
... and diffs them at the beginning and end of the pytest run.

It can easily be extended.

This allows us to keeping track of what modifications to the system state are actually done in a test run.

A nice output is shown if something is found in the diff.
Example output if failures are found (installed vim, loaded some kernel module, changed some kernel settings,  changed some password):

```
🧪  running tests with args: ./run_tests --junit-xml=/dev/virtio-ports/test_junit --system-booted --allow-system-modifications --expected-users gardenlinux
...
================================================= test session starts ==================================================
platform linux -- Python 3.13.5, pytest-8.4.2, pluggy-1.6.0
rootdir: /run/gardenlinux-tests/tests
plugins: order-1.3.0
collected 54 items

test_sysdiff.py .                                                                                                [  1%]
test_basics.py ......                                                                                            [ 12%]
test_containers.py .                                                                                             [ 14%]
test_network.py .......s                                                                                         [ 29%]
test_nodejs.py ss                                                                                                [ 33%]
test_packages.py ......                                                                                          [ 44%]
test_proc.py ..                                                                                                  [ 48%]
test_sapmachine.py sss                                                                                           [ 53%]
test_ssh.py ..........s..                                                                                        [ 77%]
test_users.py ...                                                                                                [ 83%]
test_vmware.py ssssssss                                                                                          [ 98%]
test_sysdiff.py F                                                                                                [100%]

======================================================= FAILURES =======================================================
_______________________________________________ test_sysdiff_after_tests _______________________________________________
test_sysdiff.py:38: in test_sysdiff_after_tests
    pytest.fail(f"System changes were detected during the test run:\n{diff_output}")
E   Failed: System changes were detected during the test run:
E   === Package changes (before-tests -> after-tests) ===
E   --- packages@snapshot_a
E   +++ packages@snapshot_b
E   @@ -116,6 +116,7 @@
E    libgmp10:amd64	2:6.3.0+dfsg-3
E    libgnutls30t64:amd64	3.8.9-3
E    libgpg-error0:amd64	1.51-4
E   +libgpm2:amd64	1.20.7-11+b2
E    libgssapi-krb5-2:amd64	1.21.3-5gl0
E    libhogweed6t64:amd64	3.10.1-1
E    libibverbs1:amd64	56.1-1
E   @@ -201,6 +202,7 @@
E    libsepol2:amd64	3.8.1-1gl0
E    libsgutils2-1.48:amd64	1.48-2
E    libsmartcols1:amd64	2.41-5
E   +libsodium23:amd64	1.0.18-1+b2
E    libsqlite3-0:amd64	3.46.1-7
E    libss2:amd64	1.47.2-3+b3
E    libssh2-1t64:amd64	1.11.1-1
E   @@ -352,6 +354,7 @@
E    udev	257.7-1gl0
E    util-linux	2.41-5
E    uuid-runtime	2.41-5
E   +vim	2:9.1.1230-2
E    vim-common	2:9.1.1230-2
E    vim-tiny	2:9.1.1230-2
E    wget	1.25.0-2
E   
E   === File content changes (sha256, path) (before-tests -> after-tests) ===
E   --- files@snapshot_a
E   +++ files@snapshot_b
E   @@ -573,7 +573,7 @@
E    628447dcfa3bfbe4c3df9f0885dcbe93a757b82d5dc91369da5139864a969663  /etc/sensors3.conf
E    20c48954659cf753baa383ecde0e6f026fadc06c2c9fbe29d88d928188c3ec17  /etc/services
E    3c46704b553c4b55ce928ffe89badfcfd08a02f0e6558211dfd57d9ae1e72aa4  /etc/sgml/xml-core.cat
E   -aa4fc4af04e70bb1bc7d403b5200c96ba3c1ba11abca064f296e436b934716d1  /etc/shadow
E   +af6ebfda101320be90b225dd61a8ad70b4d66799412dc8eecc693311c2effecd  /etc/shadow
E    400621734cb3f750d24c9a41770c7a445c292948be89723aae2317d2183d9359  /etc/shadow-
E    7c3c5b8e005446bad0b8f5ba8a121a9a0501c8dc417510871a0097d2ca329755  /etc/shells
E    26882b79471c25f945c970f8233d8ce29d54e9d5eedcd2884f88affa84a18f56  /etc/skel/.bash_logout
E   
E   === Sysctl parameter changes (before-tests -> after-tests) ===
E   --- sysctl@snapshot_a
E   +++ sysctl@snapshot_b
E   @@ -465,7 +465,7 @@
E    net.ipv4.tcp_app_win=31
E    net.ipv4.tcp_autocorking=1
E    net.ipv4.tcp_available_congestion_control=reno cubic
E   -net.ipv4.tcp_available_ulp=mptcp
E   +net.ipv4.tcp_available_ulp=mptcp tls
E    net.ipv4.tcp_backlog_ack_defer=1
E    net.ipv4.tcp_base_mss=1024
E    net.ipv4.tcp_challenge_ack_limit=2147483647
E   
E   === Kernel module changes (before-tests -> after-tests) ===
E   --- kernel_modules@snapshot_a
E   +++ kernel_modules@snapshot_b
E   @@ -33,6 +33,7 @@
E    nls_ascii
E    nls_cp437
E    nvme_fabrics
E   +nvme_tcp
E    qemu_fw_cfg
E    quota_tree
E    quota_v2
E   @@ -43,6 +44,7 @@
E    scsi_transport_iscsi
E    serio_raw
E    sunrpc
E   +tls
E    vfat
E    virtio_blk
E    virtio_console
E   
E   Tip: lines starting with '+' mean new/changed; '-' mean removed/previous state.
...
=============================================== short test summary info ================================================
PASSED test_sysdiff.py::test_sysdiff_before_tests
...
FAILED test_sysdiff.py::test_sysdiff_after_tests - Failed: System changes were detected during the test run:
================================ 1 failed, 38 passed, 15 skipped, 8 warnings in 18.22s =================================
```

Fixes #3488

**Notes for reviewers**:

- The containerd tests fail and will be fixed in the PR, too.
- The nvme tests fail and will be fixed in this PR, too.

### DIFF
--- a/tests-ng/conftest.py
+++ b/tests-ng/conftest.py
@@ -9,6 +9,13 @@ pytest_plugins = [
     if f.endswith(".py") and not f.startswith("_")
 ]
 
+handler_dir = os.path.join(os.path.dirname(__file__), "handlers")
+pytest_plugins += [
+    f"handlers.{f[:-3]}"
+    for f in os.listdir(handler_dir)
+    if f.endswith(".py") and not f.startswith("_")
+]
+
 for plugin in pytest_plugins:
     pytest.register_assert_rewrite(plugin)
 
--- a/tests-ng/test_containers.py
+++ b/tests-ng/test_containers.py
@@ -1,5 +1,4 @@
 import pytest
-from handlers.services import service_containerd
 from plugins.containerd import CtrRunner
 
 TEST_IMAGES = [
--- a/tests-ng/test_nvme.py
+++ b/tests-ng/test_nvme.py
@@ -4,7 +4,6 @@ import subprocess
 from pathlib import Path
 
 import pytest
-from handlers.configure_nvme import nvme_device
 from plugins.shell import ShellRunner
 
 module = ["nvme-tcp"]
--- a/tests-ng/test_ssh.py
+++ b/tests-ng/test_ssh.py
@@ -2,7 +2,6 @@ import os
 import pwd
 
 import pytest
-from handlers.services import service_ssh
 from plugins.sshd import Sshd
 from plugins.systemd import Systemd
 from plugins.utils import equals_ignore_case, get_normalized_sets, is_set
--- a/tests-ng/test_systemd_unit.py
+++ b/tests-ng/test_systemd_unit.py
@@ -1,5 +1,4 @@
 import pytest
-from handlers.services import service_parametrize
 from plugins.kernel_module import KernelModule
 from plugins.systemd import Systemd
 


tests-ng/plugins/sysdiff.py
```diff
@@ -0,0 +1,482 @@
...
+    paths: List[str]
+    ignore_file: bool
+
+
+@dataclass
+class Snapshot:
+    """Complete system snapshot"""
+    name: str
+    metadata: SnapshotMetadata
+    packages: Dict[str, str]  # package -> version
```
fwilhe: any reason for not using data classes for packages, files, systcl params instead of dict[str, str]? this would be easier to read and safer. tbh, I'm not sure if this then could also be used with a dict or if it would have to be a list, but I assume this would not be too bad with the amount of data we have here and readability/safety is in my view the more important aspect than efficiency

tests-ng/plugins/sysdiff.py
```diff
@@ -0,0 +1,482 @@
...
+
+    def normalize_paths(self, paths: List[str]) -> List[str]:
+        """Deduplicate and keep only existing directories/files"""
+        existing_paths = []
+        seen = set()
+
+        for path in paths:
+            if path and path not in seen:
+                try:
+                    result = self.shell(f'[ -e "{path}" ]', capture_output=True, ignore_exit_code=True)
```
fwilhe: why use shell here? sounds like this would be simpler to do in native python

tests-ng/plugins/sysdiff.py
```diff
@@ -0,0 +1,482 @@
...
+from .kernel_module import KernelModule
+from .sysctl import Sysctl
+
+
+DEFAULT_PATHS = [
+    "/etc", "/boot", "/usr/local/bin", "/usr/local/sbin",
+    "/usr/local/lib", "/opt", "/proc/mounts"
+]
+
+STATE_DIR = Path(os.environ.get("XDG_STATE_HOME", Path.home() / ".local/state")) / "sysdiff"
```
fwilhe: the sysdiff test can only be run as root, thus this home directory is always root's home directory.. not sure if that is the right place to put this

also, [the spec](https://specifications.freedesktop.org/basedir-spec/latest/#variables) mentions that this dir is supposed for state that is available for multiple runs of the application, but we are deleting it after the test has run

not exactly sure what the ideal solution would be and I think the way it is is good enough, but maybe there is a more fitting place for this

tests-ng/plugins/sysdiff.py
```diff
@@ -0,0 +1,482 @@
...
+        return snapshot
+
+    def _save_snapshot(self, snapshot: Snapshot, snapdir: Path):
+        """Save snapshot to disk"""
+
+        with open(snapdir / "packages.txt", 'w') as f:
+            for package, version in snapshot.packages.items():
+                f.write(f"{package}\t{version}\n")
+
+        with open(snapdir / "systemd_units.txt", 'w') as f:
```
fwilhe: would it be better to use json here? I see there is an unused json import, so maybe you wanted to do this and found an issue with that? this tab-separated format does not seem very stable to me.

tests-ng/plugins/sysdiff.py
yeoldegrove: Having dataclasses for everything is way more consistent, right. Every class that we use to gather data now has a propper dataclass.

tests-ng/plugins/sysdiff.py
yeoldegrove: Thanks for pointing this out. Removed shell where python suits better, e.g. also in the "finding files" section.

tests-ng/plugins/sysdiff.py
yeoldegrove: sysdiff can be run as any user but having root access is needed to get the full picture (e.g. file access).
Initially I planned to have a `sysdiff` cli command and that was the reason to put the output to the user's $HOME.
I could also put the outputs to `/tmp/` to be gone the next reboot :shrug: and do not have a strong opinion here.

I'll change it in a way that `~/.local/state/sysdiff/${DATE}-${NAME}.json` is now the format to save snapshots. This looks more like "multiple runs" are supported.
Also, I'll add back the CLI part so that there is a way to easily test changes to `tests-ng/plugins/sysdiff.py` by calling `tests-ng/util/sysdiff.py`.

tests-ng/plugins/sysdiff.py
yeoldegrove: My initial version of sysdiff was a cli in bash and having those text files is a relict from that.
I am fine moving this to json that uses the (new) dataclasses you mentioned above. Just committed that.

tests-ng/plugins/dpkg.py
```diff
@@ -26,6 +38,22 @@ def own_and_foreign_architectures(self) -> list[str]:
         architectures.append(self.architecture())
         return architectures
 
+    def collect_packages(self) -> dict[str, str]:
+        """Collect all installed packages and their versions"""
+        result = self._shell(
+            "dpkg-query -W -f='${binary:Package}\t${Version}\n'",
```
fwilhe: opinion: in scripts, it is better to use long versions of flags for the sake of readability. I had to look up what `-W` does, and I guess I'm not the only one who does not know by heart

https://manpages.debian.org/trixie/dpkg/dpkg-query.1.en.html#W~2

tests-ng/util/sysdiff.py
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env python3
```
fwilhe, line 1: not sure if I understand how I'm supposed to use this tool. I like it, I verified I can use it in a gl vm, but that setup was a bit cumbersome. Am I missing something? How did you intend this to be used @yeoldegrove ?

yeoldegrove: @fwilhe Using it in a GardenLinux VM/container/chroot is a bit tricky because you have to make sure the python env and all packages are available. Also it is not shipped in the test distribution as it lays in `tests-ng/utils` (on purpose). Shipping it together with the test distribution could be evaluated... Do you think this would be a nice feature?
I just used it on my workstation locally to quickly verify new functionality.

tests-ng/plugins/dpkg.py
yeoldegrove: fixed

Akendo: Maybe you can take the extra step and convert this into a dataclass to read the `/var/lib/dpkg/status` out. This would allow us to query for packages in a more effective way.

tests-ng/plugins/sysctl.py
```diff
@@ -9,8 +20,99 @@ def __getitem__(self, key: str):
             value = f.read().strip()
         return int(value) if value.isdigit() else value
 
+class Sysctl:
+    """Collects kernel parameters from /proc/sys/"""
+
+    # Parameters that change frequently and should be allow-listed
+    ALLOWLISTED_PARAMS = {
```
Akendo: Could this be put into a separate file that can be configured?

tests-ng/plugins/sysctl.py
```diff
@@ -9,8 +20,99 @@ def __getitem__(self, key: str):
...
+
+    def collect_sysctl_parameters(self) -> dict[str, str]:
+        """Collect all readable sysctl parameters, excluding allow-listed ones"""
+        if not self.is_sysctl_available():
+            return {}
+
+        sysctl_params = {}
+
+        try:
+            result = self.shell("sysctl -a 2>/dev/null", capture_output=True, ignore_exit_code=True)
```
Akendo: @nkraetzschmar did build a sysctl module; could we not use this instead?

tests-ng/plugins/dpkg.py
Akendo: I hacked a small class to do this:

```
class Package()
    def __init__(self, **kwargs):
        self.__dict__.update(kwargs)

    def __repr__(self):
        return self.Package

with open("/var/lib/dpkg/status") as file:
    dpkg_status=file.read()

packages_list = []
for n in dpkg_status.splitlines():
    if len(n) != 0:
        print(n.split(": "))
        if n.split(": ")[0] == "Package":
            packages_list.append(Package(Package=n.split(": ")[1]))
        else:
            if len(n.split(": ")) == 1:
                packages_list[-1].__dict__[field] += n.split(": ")[0].lstrip()
            else:
                field = n.split(": ")[0]
                packages_list[-1].__dict__[n.split(": ")[0]] = n.split(": ")[1]

```

This provides you with a list of packages and almost all its fields we have in the status file. There are still some cases missed, like `Conffiles` not being added to it. But we can put this into a `Dpkg` class and make it this way to inspect the database in a more pythonic way.

yeoldegrove: @Akendo 2c7650be3514c1c6a56410c62c8cea7aebec92a2 implements proper dataclasses and enriches them by parsing `/var/lib/dpkg/status`.

tests-ng/plugins/sysctl.py
yeoldegrove: @Akendo In ef9915aa9e2deb7bfafb81074b81d5ef4534ac18 I now made `IGNORED_SYSCTL_PARAMS` a constant in `sysdiff.py`. Now it works the same way as for systemd and kernel modules. We can decide later if e.g. a config file is needed for this. For now, this should be pretty static.

tests-ng/plugins/sysctl.py
yeoldegrove: @Akendo In 664d008eaeab9ead33cd794ccffcde22e975f978 I now parse `/proc/sys` instead of using `sysctl`. I removed the unused `SysctlWrapper` class, too.

yeoldegrove:
> > The containerd tests fail and will be fixed in #3507.
> > The nvme tests fail and will be fixed in #3509.
> 
> To be honest I am hesitant to approve this PR with red ci.. we should not do this. What is the reason against doing it in this PR? Yes it will be larger, but also the tests will be green. We can do this if you think it is a good idea, but then all those PRs should be merged in a short time frame to avoid a red build on `main` for a long time.

@Akendo What is your take on this?

yeoldegrove:
@Akendo @fwilhe Added an ADR that explains the reasoning and also mentions initial failures in CI for tests that are not explicit about intended modifications.

docs/architecture/decisions/0013-test-ng-system-state-diffing.md
```diff
@@ -0,0 +1,73 @@
...
+
+- Installed packages (dpkg).
+- Systemd unit states.
+- Filesystem checksums for selected paths.
+- Sysctl parameters.
+- Loaded kernel modules.
+
+Implementation choices:
+
+- Snapshots are stored in the invoking user's data directory at `~/.local/state/sysdiff/`. This allows (in theory) that it can be run by any use.
```
fwilhe:
```suggestion
- Snapshots are stored in the invoking user's data directory at `~/.local/state/sysdiff/`. This allows (in theory) that it can be run by any user.
```

tests-ng/plugins/sysdiff.py
```diff
@@ -0,0 +1,550 @@
...
+- sysctl parameters
+- loaded kernel modules
+"""
+
+import difflib
+import gzip
+import json
+import os
+import re
+import tempfile
```
fwilhe: minor: unused import

yeoldegrove: :+1:

docs/architecture/decisions/0014-test-ng-system-state-diffing.md
```diff
@@ -0,0 +1,73 @@
+# ADR 0013: System State Diffing in tests-ng
```
fwilhe, line 1: ID 13 is now claimed by https://github.com/gardenlinux/gardenlinux/pull/3632

please rename the file and change the heading

```suggestion
# ADR 0014: System State Diffing in tests-ng
```

docs/architecture/decisions/0014-test-ng-system-state-diffing.md
```diff
@@ -0,0 +1,73 @@
...
+
+- Installed packages (dpkg).
+- Systemd unit states.
+- Filesystem checksums for selected paths.
+- Sysctl parameters.
+- Loaded kernel modules.
+
+Implementation choices:
+
+- Snapshots are stored in the invoking user's data directory at `~/.local/state/sysdiff/`. This allows (in theory) that it can be run by any user.
```
nkraetzschmar: why use a user home relative path here? wouldn't a random tmp directory that is created per test-ng run be cleaner?

tests-ng/plugins/dpkg.py
```diff
@@ -1,31 +1,131 @@
 import pytest
-from .shell import ShellRunner
+import json
+from dataclasses import dataclass, field
+from typing import Dict, List
+
+
+@dataclass
```
nkraetzschmar: instead of re-engineering the debian package / index format ourselves we might just wanna use existing solutions like `debian.deb822`. Yes this would increase the number of dependencies of the test framework, but since deb package indices are quite fundamental to GL I'd argue this is fine to pull in

tests-ng/plugins/sysctl.py
```diff
@@ -1,16 +1,52 @@
+import os
 import pytest
+from dataclasses import dataclass
+from .shell import ShellRunner
 
-from plugins.booted import is_system_booted
 
-class SysctlWrapper:
-    def __getitem__(self, key: str):
```
nkraetzschmar, line 6: the `__getitem__` method should not be abandoned imo. it allows for clean test syntax like https://github.com/gardenlinux/gardenlinux/blob/b7f3bfc723a3c292cf585e1d573539b3db4c45f5/tests-ng/_test_dummy.py#L41

which reads way cleaner in test asserts then having a sysctl fixture an which one calls `collect_sysctl_parameters` and then accesses that dict.

tests-ng/plugins/sysdiff.py
```diff
@@ -0,0 +1,548 @@
...
+
+from .shell import ShellRunner
+from .dpkg import Dpkg, Package
+from .systemd import Systemd, SystemdUnit
+from .kernel_module import KernelModule, LoadedKernelModule
+from .sysctl import Sysctl, SysctlParam
+
+STATE_DIR = Path(os.environ.get("XDG_STATE_HOME", Path.home() / ".local/state")) / "sysdiff"
+
+DEFAULT_PATHS = [
```
nkraetzschmar, line 34: Not sure how compatible this default set will be with use cases like running the tests on live gardener nodes. While yes, the *tests* should not modify those paths, it is quite likely that the main workload deployed on the node will.

tests-ng/plugins/sysdiff.py
```diff
@@ -0,0 +1,548 @@
...
+            print(f"Error collecting file hashes: {e}")
+
+        return file_hashes
+
+
+class SnapshotManager:
+    """Manages snapshot creation, storage, and retrieval"""
+
+    def __init__(self, state_dir: Path = None):
+        self.state_dir = state_dir or STATE_DIR
```
nkraetzschmar: I'd vote for some tmp dir that has a random per test-ng run id

docs/architecture/decisions/0014-test-ng-system-state-diffing.md
yeoldegrove: The dir cannot be random as the cli functionality (useful for development/debuggin) would not be able to pick up previous runs. Also it should be tied to the user somehow as root will be able to read more files than any other user.

Having the state in `/tmp/sysdiff/${date}-${user}-${name}.json.gz` could work I guess. @nkraetzschmar Are you fine with that?

tests-ng/plugins/dpkg.py
yeoldegrove: I also thought about that initially but thought that is against our "no dependency" paradigm.
I can rework this to use https://pypi.org/project/python-debian/ ... but "quite fundamental" would also be true for other plugins we use here like sysctl or systemd (would pull in binary wheels).

tests-ng/plugins/sysctl.py
yeoldegrove: This got lost during refactoring as no real test is currently using it.
Will add back the useful functionality.

tests-ng/plugins/sysdiff.py
yeoldegrove: These are the defaults that are needed now to detect/skip things. These are constants but of course can be adapted/extended. If we see the need to change those based on the environment, we could pass flags from pytest to this plugin and extend/overwrite them.

tests-ng/plugins/sysdiff.py
yeoldegrove: See comment above.

docs/architecture/decisions/0014-test-ng-system-state-diffing.md
nkraetzschmar: then I'd suggest the cli tool passes in some directory to use in those cases, but for the regular test runs I'd really prefer a random tmp directory, so that theoretically even 2 simultaneous test runs wouldn't conflict

tests-ng/plugins/sysctl.py
nkraetzschmar: there should be some tests that need that, those were just not ported over to the new framework yet. can't fully remember which tests though

docs/architecture/decisions/0014-test-ng-system-state-diffing.md
yeoldegrove: Reworked it as mentioned and if one now executes this twice in the same second one gets a useful error message.

```
root@3ee40d205527:/# /mnt/tests-ng/util/sysdiff.py -n foo
podman run -it --rm \
    -v $PWD:/mnt \
    ghcr.io/gardenlinux/nightly:nightly \
    bash -c "
        apt update;
        apt install -y python3-pip;
        pip install --break-system-packages -r /mnt/tests-ng/util/requirements.txt;
        bash"

root@3ee40d205527:/# /mnt/tests-ng/util/sysdiff.py -n foo
✓ Snapshot created successfully: 2025-10-13-12-30-15-root-foo
  Location: /tmp/sysdiff/2025-10-13-12-30-15-root-foo.json.gz
  Packages: 103
  Files: 1531
  Sysctl params: 963
  Kernel modules: 256
  Systemd units: 0
root@3ee40d205527:/# /mnt/tests-ng/util/sysdiff.py -n foo
✗ Error: Snapshot '2025-10-13-12-30-15-root-foo' already exists at /tmp/sysdiff/2025-10-13-12-30-15-root-foo.json.gz
```

tests-ng/plugins/sysctl.py
yeoldegrove: Yep this will be useful in the future. It is added back.

tests-ng/plugins/dpkg.py
yeoldegrove: `debian.deb822` is now used throughout `dpkg.py` and `sysdiff.py`.

docs/architecture/decisions/0014-test-ng-system-state-diffing.md
yeoldegrove: Renamed the ADR.

docs/architecture/decisions/0014-test-ng-system-state-diffing.md
yeoldegrove: Changed it to iso date format and microsecond resolution.

```
root@f5edb6798e9c:/# /mnt/tests-ng/util/sysdiff.py -n foo
✓ Snapshot created successfully: 2025-10-13T12:43:06.189118-root-foo
  Location: /tmp/sysdiff/2025-10-13T12:43:06.189118-root-foo.json.gz
  Packages: 103
  Files: 1531
  Sysctl params: 963
  Kernel modules: 256
  Systemd units: 0
```

tests-ng/plugins/sysctl.py
```diff
@@ -1,16 +1,67 @@
...
+    """Represents a sysctl parameter"""
+    name: str
+    value: str
+
+    def __str__(self) -> str:
+        return f"{self.name}={self.value}"
+
+
+class Sysctl:
+    """Collects kernel parameters from /proc/sys/"""
```
Akendo, line 21: What's the reason that we do not use the [`py-sysctl`](https://pypi.org/project/sysctl/) module here?

yeoldegrove: @Akendo Could you elaborate which module you mean exactly and why using it would be beneficial?

Akendo: Sure. The benefit is that it's less code to maintain, and we use the C API to perform the syscalls. Which means we interact less with the FS.

yeoldegrove: [`sysctl`](https://pypi.org/project/sysctl/) needs some native C components and we currently do not consume python binary wheels (on purpose)
https://github.com/gardenlinux/gardenlinux/blob/cf6c4fdc55b73d27c3d3153b64c9f65fa6758561/tests-ng/util/build_runtime.sh#L93

So this library is currently not supported by test-ng.

tests-ng/handlers/services.py
```diff
@@ -0,0 +1,93 @@
+import time
```
nkraetzschmar, line 1: Could you please explain why this is done as a handler and not a plugin? As this is defining fixtures I'd assume this should go into a plugin.

tests-ng/test_sysdiff.py
```diff
@@ -0,0 +1,67 @@
+import pytest
+from plugins.sysdiff import Sysdiff
+
+
+@pytest.mark.order("first")
```
nkraetzschmar, line 8: How stable is this in case any other test may in the future use `@pytest.mark.order("first")` as well?

tests-ng/test_sysdiff.py
```diff
@@ -0,0 +1,67 @@
...
+def test_sysdiff_before_tests(sysdiff: Sysdiff):
+    """
+    Verifies no system changes were detected during the test run.
+    This test runs before all other tests and creates the before-tests snapshot.
+    """
+    name = "before-tests"
+    try:
+        snapshot = sysdiff.create_snapshot(name)
+        # Store the snapshot name in a global variable
+        globals()["before_tests_snapshot_name"] = snapshot.name
```
nkraetzschmar: What's the reason to do this over `global before_tests_snapshot_name`?
Also why store the name as a global rather then the snapshot object or something like that?

tests-ng/handlers/services.py
yeoldegrove: "Handlers are pytest fixtures that manage test setup and teardown" [1].
They are basically plugins that `yield` something.

[1] https://github.com/gardenlinux/gardenlinux/blob/63ebea7f62471de53992ab8deebc5c615c37c390/tests-ng/DEVELOPER.md#handlers-tests-nghandlers

tests-ng/test_sysdiff.py
yeoldegrove: `pytest-order` is quite flexible and e.g. ordering by index or even in relation to other tests is possible.
https://pytest-order.readthedocs.io/en/latest/usage.html

tests-ng/test_sysdiff.py
yeoldegrove: Good finding. This was written before the discussions about the global variables we had back last week.
[0c4612b](https://github.com/gardenlinux/gardenlinux/pull/3506/commits/0c4612be837da19724f6c089882d0913dfb90301) now gets rid of those variables and finds the created before-snapshot based on it's suffix with `list_snapshots()`.

tests-ng/handlers/services.py
nkraetzschmar: If we are going to have fixtures and other stuff that pytest should be aware of in handlers, then I'd say we also need to consider the handlers here: https://github.com/gardenlinux/gardenlinux/blob/0ec4654f82c3e38d6c039c7b51328e7a67b8c252/tests-ng/conftest.py#L5

I guess it does work without due to explicitly importing the handler, but we should stick with some convention that either everything that is providing pytest relevant overwrites / fixtures should be loaded dynamically via conftests, or everything is loaded only via imports, but I dislike inconsistent mix-and-match. And I guess doing everything via imports does not work as this would only work for fixtures and not for things like `pytest_collection_modifyitems`.

Or, alternatively, we could have things like `handle_service` in the handler and then the fixtures using this in a similarly named plugin which imports the handler.

yeoldegrove: To be honest I missed this auto registration so far. I guess having this is nice. Maybe we need to rethink the naming of the plugin/handler functions at some point in time, e.g. have some `handler_*` namespace. @ByteOtter also already suggested that.

0e6db9e48b7b0190c2cc78c8aea4728553d91a4f  does the registering and removes the imports.

PR #3518, opened 2025-09-29
Tests: Migrate test_password_shadow to new testing framework

### LINKED ISSUE #3470: Migrate `base/test/test_password_shadow.py` to new testing framework.

ByteOtter:

This PR migrates the tests in `features/base/test/test_password_shadow.py` to a new structure in accordance with the new testing framework.

To achieve this goal, a new generalized helper function was created to safely parse entries from any `/etc/*` file. This function could be used to update the `test_password_hashes` test or fixtures later.

Fixes #3470 

**Definition of Done:**

### DIFF
--- a/tests-ng/plugins/utils.py
+++ b/tests-ng/plugins/utils.py
@@ -1,7 +1,5 @@
 from pathlib import Path
-from typing import Dict, List, Optional
-
-import pytest
+from typing import Dict, List
 
 # Various utility functions to make tests more readable
 # This should not contain test-assertions, but only abstract details that make tests harder to read
@@ -21,7 +19,7 @@ def is_set(obj) -> bool:
 
 
 def parse_etc_file(
-    path: Path, field_names: List[str], min_fields: Optional[int] = None
+    path: Path, field_names: List[str], min_fields: int | None = None
 ) -> List[Dict[str, str | List[str]]]:
     """
     Generic parser for /etc/* files into structured dictionaries.


tests-ng/plugins/utils.py
```diff
@@ -1,14 +1,81 @@
...
 def get_normalized_sets(*sets: set) -> tuple[set, ...]:
     normalized_sets = tuple({str(v).casefold() for v in s} for s in sets)
     return normalized_sets
 
+
 def is_set(obj) -> bool:
     return isinstance(obj, set)
+
+
+def parse_etc_file(
```
fwilhe, line 21: the intention of this file was to put in code from the tests that would clutter them. the plugins are already an abstraction, and I don't see value of this indirection here. any reason against inlining this in the shadow plugin? @ByteOtter

tests-ng/plugins/utils.py
```diff
@@ -1,14 +1,81 @@
...
+from typing import Dict, List, Optional
+
 import pytest
 
 # Various utility functions to make tests more readable
 # This should not contain test-assertions, but only abstract details that make tests harder to read
 
-def equals_ignore_case(a: str, b: str) -> str:
+
+def equals_ignore_case(a: str, b: str) -> bool:
```
fwilhe, line 8: good catch, how did this not show up before? would have assumed the editor shows a warning in this case

fwilhe: extracted this into a separate pr https://github.com/gardenlinux/gardenlinux/pull/3527

tests-ng/test_password_shadow.py
```diff
@@ -0,0 +1,33 @@
...
+        pw = entry["passwd"]
+        assert pw in ["*", "x"], f"Malformed passwd entry for {entry['user']}: {pw}"
+
+
+@pytest.mark.root
+@pytest.mark.parametrize("command", ["pwck -r", "grpck -r"])
+def test_system_integrity_tools(shell, command):
+    """System tools must confirm consistency of passwd and group files."""
+    # Skip test if unable to access the files (e.g. running without root permissions)
+    if not os.access("/etc/shadow", os.R_OK) or not os.access("/etc/gshadow", os.R_OK):
```
fwilhe: is this not handled by the `root` decorator? what would happen if we remove this if statement?

tests-ng/plugins/utils.py
ByteOtter: Yes, what I wanted to do here is create a function for parsing `etc` files that is as generic as possible, so we can reuse that parsing logic over and over again independently from any test's specific scope. That's why I thought putting it there was the best way to signal "this is not a function specific to the `shadow` tests, but can be used anywhere"

tests-ng/plugins/utils.py
ByteOtter: Not sure what editor everyone is using. For me, both in LunarVim and Codium, I had to install the pyright type checker for that sort of error. It seems the regular Python extension in Codium (and therefore in Code) does not cover that really well.
Thanks for splitting this out. :)

tests-ng/test_password_shadow.py
ByteOtter: Ah yes, while writing the test initially, I completely forgot that the `root` decorator existed and added the decorator back after the fact. I'm pretty sure that the decorator would ensure that these files are accessible unless there is some kind of more severe issue with the system. In this case that would be caught much earlier. I think I can safely remove it.

tests-ng/test_password_shadow.py
```diff
@@ -0,0 +1,33 @@
+import os
+
+import pytest
+
+
+def test_shadow_passwords_are_locked(shadow_entries):
+    """No user in shadow should have a valid password hash."""
+    for entry in shadow_entries:
+        pw = entry["passwd"]
+        assert pw and pw[0] in [
```
fwilhe: this might be my missing python knowledge, but I find this hard to read

an example entry from shadow looks like this

```
sshd:!*:20361::::::
```

so we are asserting that the string is non-empty (?) and that the first element is either a `*` or a `!`? why? What is expected, what would be unexpected? can we make that more explicit so that it is easier to understand why the test fails, if it fails?

tests-ng/test_password_shadow.py
```diff
@@ -0,0 +1,33 @@
...
+            "*",
+            "!",
+        ], f"Unexpected password hash in shadow for {entry['user']}: {pw}"
+
+
+def test_passwd_password_field_is_valid(passwd_entries):
+    """All passwd entries must use 'x' or '*' in the password field."""
+    for entry in passwd_entries:
+        pw = entry["passwd"]
+        assert pw in ["*", "x"], f"Malformed passwd entry for {entry['user']}: {pw}"
```
fwilhe, line 23: here it is a bit more clear to me compared to above

tests-ng/test_password_shadow.py
ByteOtter: The test should validate that:
1. The entry contains a value for the password field.
2. The entry is locked, indicated by `*` or `!` at the first character of the password's entry.

But I agree the way that I wrote it is a bit confusing.
I would decompress this assert into two like this:
```suggestion
        # Assert that the password field for the given entry is not empty (not invalid)
        assert pw, f"Empty password field in shadow for {entry['user']}"
        # Assert whether the first character is a '!' or '*' marking the user as locked.
        assert pw[0] in ["*", "!"], f"Unexpected password hash in shadow for {entry['user']}: {pw}"
```
The test asserts that no local account can be accessed via password login, which aligns with my understanding of the fixture in `tests/helper/password_shadow.py?` called `password_shadow` where it states:

> """check that no passwords are set in /etc/shadow or /etc/passwd"""

I hope this explanation helps.

tests-ng/test_password_shadow.py
ByteOtter: Yes I agree. Mashing two asserts into one was not my brightest idea.

PR #3519, opened 2025-09-29
Document running tests as root

fwilhe:
It might not be obvious that this is an option, so better document it explicitly.


ByteOtter (approved):
lgtm 👍🏼

PR #3528, opened 2025-09-30
Tests: port systemd units test to the new test framework

Leon-hk:
Migrates `test_systemd_unit` from the following files:
```
features/gardener/test/test_systemd_units.py
features/khost/test/test_systemd_unit.py
features/firewall/test/test_systemd_units.py
features/vhost/test/test_systemd_unit.py
features/chost/test/test_systemd_unit.py
features/firecracker/test/test_systemd_unit.py
```

Tests run for these flavors:
| Flavor | Included features |
|-------|------------------|
|aws-gardener_prod-arm64 | gardener |
|kvm_dev-arm64| firewall|
|metal-vhost-arm64| vhost, firewall| 
|metal-capi-arm64|chost, khost|

I was **not** able to verify the `firecracker/rngd` test, as I couldn't find a flavor containing the firecracker feature.

Tracks #3181

I think this is a great example to show how the new test framework simplifies a test. Six almost identical files, which import a function defined in a file far away from their own directory, are merged into a single one where every definition is directly visible. Additionally, it can be easily extended for further flavor/unit combinations.

My uncertainties: 
- Had to add `@pytest.mark.root` as it failed with 
```
Failed to start containerd.service: Access denied as the requested operation requires interactive authentication. However, interactive authentication has not been enabled by the calling program.
See system logs and 'systemctl status containerd.service' for details.
```
- I'm not sure if this is the intended use of `@pytest.mark.modify`


tests-ng/test_systemd_unit.py
```diff
@@ -0,0 +1,54 @@
+import pytest
+from plugins.shell import ShellRunner
+from plugins.dpkg import Dpkg
+
+
+def validate_systemd_unit(shell: ShellRunner, systemd_unit, active=True):
+    """ Validate a given systemd unit """
+    cmd = f"systemctl status {systemd_unit}.service"
```
fwilhe: shell calls to systemctl and similar tools should be in the systemd plugin to ensure consistency in how they are invoked.

Leon-hk:
The tests now use the systemd plugin and I removed the firecracker test as the feature will be removed anyways.

tests-ng/test_systemd_unit.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+from plugins.systemd import Systemd
+
+
+# The key is the feature conidition of the test which checks if the systemd unit (value) is working
```
fwilhe:
```suggestion
# The key is the feature condition of the test which checks if the systemd unit (value) is working
```

tests-ng/test_systemd_unit.py
```diff
@@ -0,0 +1,21 @@
...
+    "gardener": "containerd",
+    "vhost": "libvirtd"
+}
+
+
+@pytest.mark.parametrize(
+    "systemd_unit", [pytest.param(units[feature], marks=pytest.mark.feature(feature)) for feature in units]
+)
+@pytest.mark.booted(reason="Test runs systemd")
+@pytest.mark.root(reason="To start the systemd service")
```
fwilhe, line 18: missing the `modifies` marker

tests-ng/test_systemd_unit.py
```diff
@@ -0,0 +1,21 @@
...
+def test_systemd_unit(systemd: Systemd, systemd_unit):
+    systemd.start_unit(systemd_unit)
+    assert systemd.is_active(systemd_unit)
```
fwilhe: I'd prefer to have an assertion message here like I have in the ssh feature test for example

tests-ng/test_systemd_unit.py
Leon-hk: I removed it because the systemd plugin contains
```python
if not allow_system_modifications():
    pytest.skip("starting units is only supported when system state modifications are allowed")
```

But I can put it back in.

fwilhe: I don't understand, this means your tests will always be skipped without the marker if I get this right. Yes please put it in.

fwilhe: this if statement is to skip when you run the tests without the opt-in flag

fwilhe: cf https://github.com/gardenlinux/gardenlinux/blob/6f7a7b1920854423f5f4433ced3c168a1a48ed15/tests-ng/README.md?plain=1#L275

Leon-hk: Ah okay, I see

Leon-hk:
Good that you noticed it! I didn't thought of it and only tested CI flavors.

It seems like `kubelet` is installed only in 
https://github.com/gardenlinux/gardenlinux/blob/d4b453c2ae2e460393e934b526691edeeb98112b/features/khost/exec.late#L10
and is therefore not included in flavors which directly contain chost.

I just copied the test case from [features/chost/test/test_systemd_unit.py](https://github.com/gardenlinux/gardenlinux/blob/main/features/chost/test/test_systemd_unit.py).
Either we remove `chost` from the test case or also install `kubelet` for it.

PR #3529, opened 2025-09-30
Migrate test_su

### LINKED ISSUE #3526: Migrate base/test/test_su.py:: test_su_restriction

commel:
Migrate https://github.com/gardenlinux/gardenlinux/blob/main/features/base/test/test_su.py to the new testing framework.

Fixes #3526 
Refs #3181 

**Definition of Done:**

### DIFF
--- a/tests-ng/test_su.py
+++ b/tests-ng/test_su.py
@@ -1,0 +1,31 @@
+import pytest
+
+from plugins.pam import PamConfig
+
+@pytest.mark.security_id(166)
+@pytest.mark.root
+@pytest.mark.parametrize("pam_config", [("/etc/pam.d/su")], indirect=["pam_config"])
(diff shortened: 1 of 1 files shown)


ByteOtter (commented):
As we just discussed, maybe we should try to fit the [`PamConfig`](https://github.com/gardenlinux/gardenlinux/blob/main/tests-ng/plugins/password_hashes.py) plugin here instead of doing substring matching?
We may need to change some parts of it to make it more general, but we will figure it out.

ByteOtter:
I went ahead and split of your changes to the `pam_config` fixture, `test_password_hashes` tests and the `PamEntry` dataclass into a separete PR along with the refactoring of the `find_entries` function. This one should allow you more control over what `control` fields we are looking for.

https://github.com/gardenlinux/gardenlinux/pull/3565

PR #3566, opened 2025-10-01
tests-ng: Avoid unneeded write to stdout, avoid shell usage

fwilhe:
Minor refactoring:

"echo foo | tee" writes foo to stdout and the provided file, but we only need it inside that file. Also, we don't need to use shell for that as python is capable of writing files natively.

This is what this looked like before:
Screenshot 2025-10-01 at 14 34 41


moritzkirchnersap (approved):
LGTM

PR #3570, opened 2025-10-06
fix: test-ng: more robust check for tofuenv and GITHUB_TOKEN in run_cloud.sh

yeoldegrove:

* more robust check for tofuenv and GITHUB_TOKEN in run_cloud.sh
* Mention cloud provider CLIs in docs.


commel (approved):
LGTM, we tested this out on my machine.

PR #3628, opened 2025-10-08
Pin dependency versions in tests-ng and setup dependabot

fwilhe:
This will make the tests run more predictable as dependency versions don't change unnoticed. Dependabot will help keep them updated.


yeoldegrove (approved):
LGTM

nkraetzschmar (approved):

PR #3632, opened 2025-10-09
docs: ADR to discontinue packages_musthave tests

### LINKED ISSUE #3618: test-ng: should we keep the `test_packages_musthave` tests?
@fwilhe and @nkraetzschmar  suggested during the "Let's plan Test Framework Migration" meeting to discontinue the `test_packages_musthave` tests, as they have several downsides. 
When I worked on migrating the tests, I also noticed another problem.
Florian and I already discussed some of it and decided to discuss in a larger round.

**Pros: Why they should be kept**
- They provide value in finding misconfigured or outdated `pck.include` files, like #3617
- The test can be easily extended by adding the feature into the feature dictionary of the test

**Cons: Why the tests should be removed/replaced by separate feature tests**
- To check if all packages are installed, the tests have to access the `pck.include` and `pck.exclude` files located in the feature folders
    - These files are not contained in the tests tar-ball and have to be explicitly added
    - I've created such a feature in #3621 
    - But this violates the concept of being independent of the current local repository state
- Hardcoding the required packages would create redundancy, which is hard to maintain

Feel free to extend the pros and cons list and to share your opinion!

Leon-hk:
- Discontinue `packages_musthave` tests

Fixes #3618

### DIFF
--- a/docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md
+++ b/docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md
@@ -8,18 +8,22 @@ Accepted
 
 ## Context
 
-`test_packages_musthave.py` checks if all packages contained in the feature's `pkg.include` and not contained in a `pkg.exclude` of any feature of the flavor is installed.
+We are currently migrating from the old feature oriented test framework to a new independent test framework, where all test related files are stored in a separate directory. This results in better readability and should also make clear which issues a test is about. 
+In the old framework, `test_packages_musthave.py` checks if all packages contained in the feature's `pkg.include` and not contained in a `pkg.exclude` of any feature of the flavor is installed. It only includes one line which is an import from a helper file.
 
 ## Decision
 
 The `packages_musthave` will not be migrated to the new test framework:
 
-- As the `packages_musthave` test uses `pkg.include` and `pkg.exclude` files of the currently used repository, the tests are not version independant
-    - The new test framework is capable of testing production systems, which may have older versions
+- As the `packages_musthave` test uses `pkg.include` and `pkg.exclude` files of the currently used repository, the tests are not version independent
+    - The new test framework is capable of testing production systems, which may be built with an older version of a feature than the test framework and may have fewer or other packages in that version, which would fail the test unintentionally
 - The test only has low value, e.g. detecting misspelled or incomplete package names
+- If a package is not found, the build will already fail
 - Important packages can be checked in the feature test
 - If a real value is found, the implementation from #3621 could be used
 
 ## Consequences
 
 - It is no longer checked if all included packages are installed
+- The test distribution artifact does not need to contain the package lists
+- The tests have less potential incompatibilities between versions (different commit between image build and test distribution build)


docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Context
+
+`test_packages_musthave.py` checks if all packages contained in the feature's `pkg.include` and not contained in a `pkg.exclude` of any feature of the flavor is installed.
+
+## Decision
+
+The `packages_musthave` will not be migrated to the new test framework:
+
+- As the `packages_musthave` test uses `pkg.include` and `pkg.exclude` files of the currently used repository, the tests are not version independant
```
fwilhe:
```suggestion
- As the `packages_musthave` test uses `pkg.include` and `pkg.exclude` files of the currently used repository, the tests are not version independent
```
please use a spellchecker, just this one struck my eye

docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md
```diff
@@ -0,0 +1,25 @@
+# 13. Discontinue `packages_musthave` tests
+
+Date: 2025-10-09
+
+## Status
+
+Accepted
+
+## Context
+
```
fwilhe, line 10: maybe use 2 more sentences to describe how we are migrating from the old to the new system and the reasoning (i.e. we want the tests to be easier to understand and each test should be clear about what potential issue they protect against)

docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md
```diff
@@ -0,0 +1,25 @@
...
+`test_packages_musthave.py` checks if all packages contained in the feature's `pkg.include` and not contained in a `pkg.exclude` of any feature of the flavor is installed.
+
+## Decision
+
+The `packages_musthave` will not be migrated to the new test framework:
+
+- As the `packages_musthave` test uses `pkg.include` and `pkg.exclude` files of the currently used repository, the tests are not version independant
+    - The new test framework is capable of testing production systems, which may have older versions
```
fwilhe: not sure if this is clear without the context I and you now have

suggestion: The new test framework is capable of testing production systems, which may be built with an older version of a feature than the test framework and may have fewer or other packages in that version, which would fail the test unintentionally

docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md
```diff
@@ -0,0 +1,25 @@
...
+- The test only has low value, e.g. detecting misspelled or incomplete package names
```
fwilhe, line 20: in this case build time should already fail..

docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md
```diff
@@ -0,0 +1,25 @@
...
+- Important packages can be checked in the feature test
+- If a real value is found, the implementation from #3621 could be used
+
+## Consequences
+
+- It is no longer checked if all included packages are installed
```
fwilhe, line 27: - the test distribution artefact does not need to contain the package lists
- the tests have less potential incompatibilities between versions (different commit between image build and test distribution build)

PR #3634, opened 2025-10-09
Test: Port users test to new test framework

SujanaSubr:
Ports test_users to new test framework
Fixes #
https://github.com/gardenlinux/gardenlinux/issues/3635

**Definition of Done:**

**Release note**:

```feature user

```

### DIFF
--- a/.github/workflows/test_flavor_chroot.yml
+++ b/.github/workflows/test_flavor_chroot.yml
@@ -50,6 +50,11 @@ jobs:
           rm "$CNAME.tar.gz"
 
           make --directory=tests ${{ inputs.flavor }}-${{ inputs.arch }}-chroot-test
+      - name: Download test distribution artifact
+        uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # pin@v5.0.0
+        with:
+          name: test-distribution
+          path: tests-ng/.build
       # chroot.test.xml is written in the entrypoint tests/init
       - name: Rename test results
         if: always()
--- a/features/_tpm2/initrd.include/etc/systemd/system/check-tpm.service
+++ b/features/_tpm2/initrd.include/etc/systemd/system/check-tpm.service
@@ -1,6 +1,7 @@
 [Unit]
 DefaultDependencies=no
 After=tpm2.target
+Wants=tpm2.target
 Before=systemd-repart.service
 
 [Service]
--- a/tests-ng/README.md
+++ b/tests-ng/README.md
@@ -71,6 +71,7 @@ Before running the test framework, make sure the following dependencies are inst
 - `uuid-runtime`
 - `qemu`
 - `qemu-utils`
+- `socat`
 
 If you plan to provision cloud resources, the cloud provider specific CLIs might be useful or even required:
 
--- a/tests-ng/plugins/capabilities.py
+++ b/tests-ng/plugins/capabilities.py
@@ -1,0 +1,36 @@
+import pytest
+from pyprctl import FileCaps
+
+from .find import FIND_RESULT_TYPE_FILE, Find
+from .shell import ShellRunner
+
+
+class Capabilities:
+    def __init__(self, find: Find, shell: ShellRunner):
+        self._find = find
+        self._shell = shell
+
+    def get(self) -> set[str]:
+        capabilities = set()
+
+        self._find.same_mnt_only = True
+        self._find.root_paths = ["/boot", "/etc", "/usr", "/var"]
+        self._find.entry_type = FIND_RESULT_TYPE_FILE
+
+        for file in self._find:
+            try:
+                capability = FileCaps.get_for_file(file)
+            except OSError:
+                # Skip files without capability xattr or unreadable entries
+                continue
+
+            if capability:
+                # getcap style output
+                capabilities.add(f"{file} {str(capability)}")
+
+        return capabilities
+
+
+@pytest.fixture
+def capabilities(find: Find, shell: ShellRunner) -> Capabilities:
+    return Capabilities(find, shell)
--- a/tests-ng/plugins/find.py
+++ b/tests-ng/plugins/find.py
@@ -1,0 +1,74 @@
+import os
+from typing import Iterator, Union
+
+import pytest
+
+FIND_RESULT_TYPE_FILE = "files"
+FIND_RESULT_TYPE_DIR = "directories"
+FIND_RESULT_TYPE_FILE_AND_DIR = "both"
+
+
+class Find:
+
+    def __init__(self) -> None:
+        """
+        Initializes the object with default search parameters.
+
+        Attributes:
+            same_mnt_only (bool): If True, restricts search to the same mount point.
+            root_paths (Union[str, list[str]]): The root directory paths to start the search from.
+                If a list is provided, all paths will be searched.
+            entry_type (str):
+                Specifies the type of entries to search for (files, directories, or both).
+        """
+        self.same_mnt_only: bool = False
+        self.root_paths: Union[str, list[str]] = "/"
+        self.entry_type: str = FIND_RESULT_TYPE_FILE
+
+    def __iter__(self) -> Iterator[str]:
+        return iter(list(self._find()))
+
+    def _find(self) -> Iterator[str]:
+        # Intentionally don't use pathlib/rglob for performance reasons
+        root_paths = self.root_paths
+        if isinstance(root_paths, str):
+            root_paths = [root_paths]
+
+        for root_path in root_paths:
+            root_dev = os.stat(root_path).st_dev
+            for dirpath, dirnames, filenames in os.walk(root_path):
+
+                if self.entry_type in (
+                    FIND_RESULT_TYPE_DIR,
+                    FIND_RESULT_TYPE_FILE_AND_DIR,
+                ):
+                    for dirname in dirnames:
+                        full_path = os.path.join(dirpath, dirname)
+                        if self.same_mnt_only:
+                            try:
+                                if os.stat(full_path).st_dev != root_dev:
+                                    continue
+                            # Skip dead symlinks
+                            except FileNotFoundError:
+                                continue
+                        yield full_path
+
+                if self.entry_type in (
+                    FIND_RESULT_TYPE_FILE,
+                    FIND_RESULT_TYPE_FILE_AND_DIR,
+                ):
+                    for filename in filenames:
+                        full_path = os.path.join(dirpath, filename)
+                        if self.same_mnt_only:
+                            try:
+                                if os.stat(full_path).st_dev != root_dev:
+                                    continue
+                            # Skip dead symlinks
+                            except FileNotFoundError:
+                                continue
+                        yield full_path
+
+
+@pytest.fixture
+def find() -> Find:
+    return Find()
--- a/tests-ng/test_basics.py
+++ b/tests-ng/test_basics.py
@@ -51,9 +51,7 @@ def test_fhs(shell: ShellRunner):
     reason="We can only measure startup time if we actually boot the system"
 )
 @pytest.mark.performance_metric
-@pytest.mark.feature(
-    "server and not azure"
-)  # server installs systemd and azure has notoriously bad startup times
+@pytest.mark.feature("server", reason="server installs systemd")
 def test_startup_time(systemd: Systemd):
     tolerated_kernel = 60.0
     tolerated_userspace = 40.0
--- a/tests-ng/test_capabilities.py
+++ b/tests-ng/test_capabilities.py
@@ -1,0 +1,21 @@
+import pytest
+from plugins.capabilities import Capabilities
+
+# There are two debian packages that provide an 'arping' binary
+# https://packages.debian.org/search?searchon=contents&keywords=arping&mode=path&suite=stable&arch=any
+# /usr/bin/arping is provided by iputils-arping
+# /usr/sbin/arping is provided by arping
+# We only use iputils-arping
+capabilities_allowlist = {"/usr/bin/arping cap_net_raw=ep"}
+
+
+@pytest.mark.root(reason="Need to read files not readable by unprivileged user")
+def test_only_expected_capabilities_are_set(capabilities: Capabilities):
+    actual_capabilities = capabilities.get()
+
+    unexpected = actual_capabilities - capabilities_allowlist
+
+    assert not unexpected, (
+        f"Unexpected capabilities found: {unexpected}. "
+        f"Allowed: {capabilities_allowlist}"
+    )
--- a/tests-ng/util/requirements.txt
+++ b/tests-ng/util/requirements.txt
@@ -1,6 +1,7 @@
 boolean.py==5.0
-pytest==8.4.2
+pyprctl==0.1.3
 pytest-metadata==3.1.1
 pytest-order==1.3.0
+pytest==8.4.2
 python-debian==1.0.1
 validators==0.35.0


tests-ng/test_users.py
```diff
@@ -23,3 +27,22 @@ def test_no_extra_home_directories(expected_users):
     entries = os.listdir("/home")
     unexpected = [e for e in entries if e not in expected_users]
     assert not unexpected, f"Unexpected entries in /home: {entries}"
+
+@pytest.mark.skipif(sudo_binary is None, reason="sudo does not exist")
+@pytest.mark.booted
+@pytest.mark.root
+def test_sudo_permission_for_users(expected_users, shell: ShellRunner):
+    for user in expected_users:
+        output = shell(f"sudo -s sudo -l -U {user}", capture_output=True)
```
fwilhe: would it make sense to use `sudo_binary` here? this should be the full path.. not sure it this makes a real difference, there should not be more than one `sudo` binary, but it feels a bit weird to get the full path and then rely on the shell's PATH here again

tests-ng/plugins/get_uid_range.py
```diff
@@ -0,0 +1,21 @@
+import pytest
```
fwilhe: naming the file `get_uid_range` is too specific imo, plugins are more of a generic thing and this uidrange plugin might provide more features in the future?

consider calling it `uid.py` or similar

tests-ng/plugins/get_uid_range.py
```diff
@@ -0,0 +1,21 @@
...
+
+@pytest.fixture
+def get_uid_range() -> UIDRange:
+    uid_range = UIDRange()
+    output = Path("/etc/login.defs").read_text().strip()
+    for line in output.split("\n"):
+        if not line or line.startswith("#"):
+            continue
+        if line.startswith("UID_MIN"):
+            uid_range.uid_min = int(line.split()[1])
```
fwilhe: sure we don't need error handling here?

tests-ng/test_users.py
```diff
@@ -2,10 +2,14 @@
 import pwd
 import pytest
 import stat
+import shutil
+from plugins.shell import ShellRunner
 
-def test_service_accounts_have_nologin_shell():
+sudo_binary = shutil.which("sudo")
```
fwilhe: having this here top-level makes this being executed when the file is imported as far as I know.. would it not be good to only have this inside a function/singleton of sorts?

tests-ng/test_users.py
```diff
@@ -23,3 +27,22 @@ def test_no_extra_home_directories(expected_users):
     entries = os.listdir("/home")
     unexpected = [e for e in entries if e not in expected_users]
     assert not unexpected, f"Unexpected entries in /home: {entries}"
+
+@pytest.mark.skipif(sudo_binary is None, reason="sudo does not exist")
+@pytest.mark.booted
+@pytest.mark.root
+def test_sudo_permission_for_users(expected_users, shell: ShellRunner):
```
fwilhe: unclear to me what the expected thing is this test ensures. can you rename the method to make it more clear what the expectation is?

tests-ng/test_users.py
```diff
@@ -23,3 +27,22 @@ def test_no_extra_home_directories(expected_users):
...
+    for user in expected_users:
+        output = shell(f"sudo -s sudo -l -U {user}", capture_output=True)
+        output_lines = output.stdout.strip().splitlines()
+        assert (len(output_lines) > 2 and 
+            "may run the following commands on" in output_lines[-2], 
+            f"User: {user} sudo permission doesn't match requirement : {output_lines[-1]}")
```
fwilhe: what is in `output_lines[-1]`? can you put `output_lines[-1]` and `output_lines[-2]` into local variables with speaking names to make it easier to read?

fwilhe: in my sample vm it looks like this

```
sudo -s sudo -l -U fwilhe
Matching Defaults entries for fwilhe on lima-gardenlinux:
    env_reset, mail_badpass, secure_path=/usr/local/sbin\:/usr/local/bin\:/usr/sbin\:/usr/bin\:/sbin\:/bin, use_pty, env_keep+=SSH_AUTH_SOCK

User fwilhe may run the following commands on lima-gardenlinux:
    (ALL) NOPASSWD: ALL
```

what would be expected?

fwilhe: please use long options for better readability

```
  -s, --shell                   run shell as the target user; a command may
  -l, --list                    list user's privileges or check a specific
  -U, --other-user=user         in list mode, display privileges for user
```

tests-ng/plugins/get_uid_range.py
SujanaSubr: naming it as system_id_parser.py so we can use it for getting group id in future if required

tests-ng/test_users.py
```diff
@@ -2,10 +2,14 @@
...
+import shutil
+from plugins.shell import ShellRunner
 
-def test_service_accounts_have_nologin_shell():
+sudo_binary = shutil.which("sudo")
+
+def test_service_accounts_have_nologin_shell(get_uid_range):
     for entry in pwd.getpwall():
-        if entry.pw_uid >= 1000:
+        if get_uid_range.uid_min <= entry.pw_uid <= get_uid_range.uid_max:
```
nkraetzschmar: While this works, couldn't we make this cleaner by making the `uid_range` fixture an object that implements `__contains__`, then we could simply write

```python
if entry.pw_uid in uid_range
```

Which in my opinion looks quite a bit nicer

tests-ng/plugins/get_uid_range.py
nkraetzschmar: How about merging this with the existing `expected_users.py` into a single combined `users.py` that provides fixtures for both the set of expected users and the regular user uid range

tests-ng/plugins/get_uid_range.py
```diff
@@ -0,0 +1,21 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+
+@dataclass
+class UIDRange:
+    uid_min: int = None
+    uid_max: int = None
+
+@pytest.fixture
+def get_uid_range() -> UIDRange:
```
nkraetzschmar: I'd suggest calling this `regular_user_uid_range` or something like this. Since fixtures are passed into test functions as if they were variables, not functions, so naming the fixture like the value it returns, not with `get_` semantics makes the tests using it read cleaner

tests-ng/plugins/get_uid_range.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+from dataclasses import dataclass
+from pathlib import Path
+
+@dataclass
+class UIDRange:
```
nkraetzschmar: This class should implement `__contains__` which should then perform the `uid_min <= uid <= uid_max` check

tests-ng/plugins/get_uid_range.py
```diff
@@ -0,0 +1,21 @@
...
+    uid_max: int = None
+
+@pytest.fixture
+def get_uid_range() -> UIDRange:
+    uid_range = UIDRange()
+    output = Path("/etc/login.defs").read_text().strip()
+    for line in output.split("\n"):
+        if not line or line.startswith("#"):
+            continue
+        if line.startswith("UID_MIN"):
```
nkraetzschmar: this would also match lines for keys that start with `UID_MIN` but have something following it. While such keys *shouldn't* be in `/etc/login.defs` it's still not a good idea to rely on this for parsing logic.

nkraetzschmar: In general I'd suggest moving this to a seperate parse config file, that properly splits every line into key, value pairs and then returns a `dict` representing the config file. Then here you'd just call that parser and get the dict key.... something like:

```python
login_defs_config = parse_config("/etc/login.defs")
uid_range.uid_min = login_defs_config["UID_MIN"]
```

nkraetzschmar: Also instead of setting min and max individually, this might be better done using a constructor for `UIDRange`. Since you made it `@dataclass` there should already be a constructor `__init__(self, min=None, max=None)` auto generated, but maybe a custom constructor that does NOT allow `None` but requires both values to be set and does sanity checks like max > min or similar would be better

tests-ng/test_users.py
nkraetzschmar: finding if there is a sudo binary (and where) would be a job for a fixture in my opinion. then we'd let pytest handle if and when this fixture needs to be evaluated

tests-ng/test_users.py
nkraetzschmar: the double `sudo` should NOT be required. We are running this as root (`@pytest.mark.root`). The double sudo was only in the old tests since they ran as a regular user.

tests-ng/test_users.py
nkraetzschmar: also can we maybe move the sudo output parsing to some function / fixture which then just returns the set of commands that the user can run, so the test logic is a nice clean check that all elements in this list are expected, without any shell output parsing logic in the test.

tests-ng/test_users.py
```diff
@@ -23,3 +27,22 @@ def test_no_extra_home_directories(expected_users):
...
+        output_lines = output.stdout.strip().splitlines()
+        assert (len(output_lines) > 2 and 
+            "may run the following commands on" in output_lines[-2], 
+            f"User: {user} sudo permission doesn't match requirement : {output_lines[-1]}")
+
+@pytest.mark.booted
+@pytest.mark.root
+def test_available_users(expected_users, get_uid_range):
+    for entry in pwd.getpwall():
+        if get_uid_range.uid_min <= entry.pw_uid <= get_uid_range.uid_max:
```
nkraetzschmar: as above, this should be:

```python
if entry.pw_uid in regular_user_uid_range
```

with the proper `__contains__` method implemented in the class returned by the fixture

tests-ng/plugins/get_uid_range.py
SujanaSubr: Will we not need any other functionality later like getting group id, system user id and so on? I thought of keeping it separate to have all the id related things in one file. If it needs to be still merged , I can do that

tests-ng/test_users.py
SujanaSubr: > in my sample vm it looks like this
> 
> ```
> sudo -s sudo -l -U fwilhe
> Matching Defaults entries for fwilhe on lima-gardenlinux:
>     env_reset, mail_badpass, secure_path=/usr/local/sbin\:/usr/local/bin\:/usr/sbin\:/usr/bin\:/sbin\:/bin, use_pty, env_keep+=SSH_AUTH_SOCK
> 
> User fwilhe may run the following commands on lima-gardenlinux:
>     (ALL) NOPASSWD: ALL
> ```
> 
> what would be expected?

I added check for system users as well. If it is system user except root, there is no root access expected. If it is normal user, then root access should be there. It will check for the string and assert accordingly

tests-ng/plugins/system_id_parser.py
```diff
@@ -0,0 +1,44 @@
...
+    uid_min: int = None
+    uid_max: int = None
+    uid_range: range = field(init=False)
+
+    def __post_init__(self):
+        assert self.uid_min is not None, "UID_MIN not found"
+        assert self.uid_max is not None, "UID_MAX not found"
+        assert self.uid_max > self.uid_min, "UID_MAX must be greater than UID_MIN"
+
+        self.uid_range = range(self.uid_min, self.uid_max + 1)
```
nkraetzschmar: Instead of creating a `range` object here and adjusting the max to account for python ranges being exclusive upper bounds, the class could just have min and max fields where the `__contains__` function simply implements the check `self.uid_min <= uid <= self.uid_max` directly, without needing the contains check of a `range` object

tests-ng/plugins/system_id_parser.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __contains__(self, uid: int) -> bool:
+        return uid in self.uid_range
+
+def parse_config():
+    config = {}
+    login_defs_path=Path("/etc/login.defs")
```
nkraetzschmar: maybe make the config path an input to this function passed in from the `regular_user_uid_range` function

tests-ng/plugins/system_id_parser.py
```diff
@@ -0,0 +1,43 @@
+import pytest
+from dataclasses import dataclass, field
+from pathlib import Path
+
+
+@dataclass
+class UIDRange:
+    uid_min: int = None
+    uid_max: int = None
+    uid_range: range = field(init=False)
```
nkraetzschmar: this var is no longer used, can be dropped here as well

tests-ng/test_users.py
```diff
@@ -23,3 +25,25 @@ def test_no_extra_home_directories(expected_users):
     entries = os.listdir("/home")
     unexpected = [e for e in entries if e not in expected_users]
     assert not unexpected, f"Unexpected entries in /home: {entries}"
+
+@pytest.mark.skipif((shutil.which("sudo")) is None, reason="sudo does not exist")
+@pytest.mark.booted
+@pytest.mark.root
+def test_users_sudo_capability(expected_users, shell: ShellRunner):
+    for entry in pwd.getpwall():
+        output = shell(f"sudo --list --other-user={entry.pw_name}", capture_output=True)
```
nkraetzschmar: https://github.com/gardenlinux/gardenlinux/pull/3634#discussion_r2418852340 still needs to be addressed (moving the parsing to a fixture or helper function)

nkraetzschmar (changes requested):

tests-ng/test_users.py
```diff
@@ -30,3 +31,27 @@ def test_no_extra_home_directories(expected_users):
...
+@pytest.mark.root
+def test_users_sudo_capability(expected_users, user: User):
+    for entry in pwd.getpwall():
+        if (entry.pw_name == "root") or (entry.pw_name in expected_users):
+            assert user.is_user_sudo(
+                entry.pw_name
+            ), f"User: {entry.pw_name} doesn't have sudo permission"
+        else:
+            assert (
+                user.is_user_sudo(entry.pw_name) == False
```
fwilhe:
```suggestion
                not user.is_user_sudo(entry.pw_name)
```

tests-ng/plugins/users.py
```diff
@@ -1,11 +1,29 @@
...
+
 users: Set[str] = set()
 
 
+class User:
+    def __init__(self, shell: ShellRunner):
+        self._shell = shell
+
+    def is_user_sudo(self, user):
+        assert shutil.which("sudo") is not None, "sudo command not found"
```
fwilhe: we don't want asserts in plugins.. is it actually a quality to test for that we have sudo? I think there are legit use-cases, where no sudo binary is there (in containers for example)

please don't assert here, rather if no sudo binary is there, return a suitable value

tests-ng/test_users.py
```diff
@@ -30,3 +31,27 @@ def test_no_extra_home_directories(expected_users):
...
+            ), f"System User {entry.pw_name} has sudo permission"
+
+
+@pytest.mark.booted
+@pytest.mark.root
+def test_available_users(expected_users, regular_user_uid_range):
+    for entry in pwd.getpwall():
+        if entry.pw_uid in regular_user_uid_range:
```
fwilhe: I feel like getting this list should be in a plugin, too many details for the test

maybe have some `get_users_in_regular_uid_range()` function, and then you can call this here and assert on the expected list returned?

tests-ng/test_users.py
```diff
@@ -30,3 +31,26 @@ def test_no_extra_home_directories(expected_users):
...
+                entry.pw_name
+            ), f"User: {entry.pw_name} doesn't have sudo permission"
+        else:
+            assert not user.is_user_sudo(
+                entry.pw_name
+            ), f"System User {entry.pw_name} has sudo permission"
+
+
+def test_available_regular_users(get_regular_users, expected_users):
+    allowed_users = ["dev", "nobody"] + list(expected_users)
```
fwilhe, line 51: unused variable. should this be used in the line below instead of expected_users?

fwilhe:
testing a `kvm_dev` image fails

```
======================================================= FAILURES =======================================================
______________________________________________ test_users_sudo_capability ______________________________________________
test_users.py:45: in test_users_sudo_capability
    assert not user.is_user_sudo(
E   AssertionError: System User dev has sudo permission
E   assert not True
E    +  where True = is_user_sudo('dev')
E    +    where is_user_sudo = <plugins.users.User object at 0xffff97dd30e0>.is_user_sudo
E    +    and   'dev' = pwd.struct_passwd(pw_name='dev', pw_passwd='x', pw_uid=1000, pw_gid=1000, pw_gecos='', pw_dir='/home/dev', pw_shell='/bin/sh').pw_name
_____________________________________________ test_available_regular_users _____________________________________________
test_users.py:54: in test_available_regular_users
    assert (
E   AssertionError: Unexpected user account found in /etc/passwd, ['dev']
E   assert not ['dev']
```

tests-ng/test_users.py
```diff
@@ -30,3 +31,26 @@ def test_no_extra_home_directories(expected_users):
...
+def test_users_sudo_capability(expected_users, user: User):
+    for entry in pwd.getpwall():
+        if (entry.pw_name == "root") or (entry.pw_name in expected_users):
+            assert user.is_user_sudo(
+                entry.pw_name
+            ), f"User: {entry.pw_name} doesn't have sudo permission"
+        else:
+            assert not user.is_user_sudo(
+                entry.pw_name
+            ), f"System User {entry.pw_name} has sudo permission"
```
fwilhe:
```suggestion
@pytest.mark.booted
@pytest.mark.root(reason="Using sudo command to check the access")
def test_users_sudo_capability(get_regular_users, expected_users, user: User):
    users_with_sudo_capabilities = set([u for u in get_regular_users if user.is_user_sudo(u)])
    allowed_sudo_users = set(expected_users) | {"root", "dev"}

    unexpected_sudo_users = users_with_sudo_capabilities - allowed_sudo_users

    assert not unexpected_sudo_users, f"Unexpected users with sudo capability: {unexpected_sudo_users}"
```

I would prefer it something like this, I think this is easier to read

fwilhe:
> is it the latest version you checked. It has dev listed, so it shouldn't fail

yes, dev image works now

PR #3637, opened 2025-10-10
tests-ng: Implement ClamAV tests

### LINKED ISSUE #3599: Migrate `feature/clamav` tests
- [x] clamav/test/test_clamav_cron.py::test_clamav_cron <- ../features/clamav/test/test_clamav_cron.py
- [x] clamav/test/test_packages_musthave.py::test_packages_musthave <- helper/tests/packages_musthave.py ❌[Removed](https://github.com/gardenlinux/gardenlinux/blob/main/docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md)

fwilhe:
Fixes #3599


tests-ng/test_clamav.py
```diff
@@ -0,0 +1,25 @@
...
+@pytest.mark.feature("clamav")
+def test_clamav_is_installed(dpkg: Dpkg):
+    assert dpkg.package_is_installed("clamav"), "clamav is not installed"
+
+
+@pytest.mark.parametrize("binary", ["freshclam", "clamscan"])
+@pytest.mark.feature("clamav")
+def test_clamav_binaries_are_available(shell: ShellRunner, binary):
+    binary_path = shutil.which(binary)
+    result = shell(f"{binary_path} --version", capture_output=True)
```
nkraetzschmar, line 17: why go the 2 stage approach here of get binary path and then run? you could just do `shell(f"{binary} --version", capture_output=True)`

tests-ng/test_clamav.py
```diff
@@ -0,0 +1,25 @@
...
+    assert "ClamAV" in result.stdout, f"Expected being able to call {binary_path}"
+
+
+@pytest.mark.root(reason="File is only readable by root")
+@pytest.mark.feature("clamav")
+def test_clamscan_cronjob_is_setup():
+    crontab = Path("/var/spool/cron/crontabs/root").read_text()
```
nkraetzschmar, line 24: crontab? really? we're still using this? shouldn't all of this be a systemd native unit?
We should double check if we actually only have this enabled via `systemd-crontab-generator` compatibility layer in which case we should change it. If we already have a proper systemd unit the test should be adjusted to reflect this.

tests-ng/test_clamav.py
fwilhe: Yes you are right, I somehow thought it would add value to manually check for the binary

tests-ng/test_clamav.py
fwilhe: Yes really that is what is in the feature, but as discussed this has no priority right now.

Issue #3615, opened 2025-10-07
Migrate `feature/vhost` tests


brdanin:
- [ ] vhost/test/test_capabilities.py::test_capabilities <- helper/tests/capabilities.py #3639 
- [x] vhost/test/test_packages_musthave.py::test_packages_musthave <- helper/tests/packages_musthave.py ❌[Removed](https://github.com/gardenlinux/gardenlinux/blob/main/docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md)  
- [x] vhost/test/test_systemd_unit.py::test_systemd_unit[libvirtd] <- ../features/vhost/test/test_systemd_unit.py #3528

Issue #3588, opened 2025-10-07
Migrate `feature/_iso` tests


brdanin:
- [ ] _iso/test/test_capabilities.py::test_capabilities <- helper/tests/capabilities.py
- [x] _iso/test/test_packages_musthave.py::test_packages_musthave <- helper/tests/packages_musthave.py ❌[Removed](https://github.com/gardenlinux/gardenlinux/blob/main/docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md)

Issue #3584, opened 2025-10-07
Migrate `feature/vmware` tests


brdanin:
- [x] vmware/test/test_check_files.py::test_check_file[/etc/cloud/cloud.cfg.d/01_debian-cloud.cfg] <- ../features/vmware/test/test_check_files.py https://github.com/gardenlinux/gardenlinux/pull/3305
- [x] vmware/test/test_check_files.py::test_check_file[/etc/cloud/cloud.cfg.d/99_disable-network-config.cfg] <- ../features/vmware/test/test_check_files.py https://github.com/gardenlinux/gardenlinux/pull/3305
- [x] vmware/test/test_check_files.py::test_check_file[/etc/cloud/cloud.cfg.d/99_enabled-datasources.cfg] <- ../features/vmware/test/test_check_files.py https://github.com/gardenlinux/gardenlinux/pull/3305
- [x] vmware/test/test_check_files.py::test_check_file[/usr/bin/dscheck_VMwareGuestInfo] <- ../features/vmware/test/test_check_files.py https://github.com/gardenlinux/gardenlinux/pull/3305
- [x] vmware/test/test_check_files.py::test_check_file[/usr/lib/python3/dist-packages/cloudinit/sources/DataSourceVMwareGuestInfo.py] <- ../features/vmware/test/test_check_files.py https://github.com/gardenlinux/gardenlinux/pull/3305
- [x] vmware/test/test_content_in_file.py::test_file_content[/etc/cloud/cloud.cfg-growpart] <- ../features/vmware/test/test_content_in_file.py https://github.com/gardenlinux/gardenlinux/pull/3305
- [x] vmware/test/test_content_in_file.py::test_file_content[/etc/cloud/cloud.cfg-ntp] <- ../features/vmware/test/test_content_in_file.py https://github.com/gardenlinux/gardenlinux/pull/3305
- [x] vmware/test/test_content_in_file.py::test_file_content[/etc/cloud/cloud.cfg-resizefs] <- ../features/vmware/test/test_content_in_file.py https://github.com/gardenlinux/gardenlinux/pull/3305
- [x] vmware/test/test_packages_musthave.py::test_packages_musthave <- helper/tests/packages_musthave.py ❌[Removed](https://github.com/gardenlinux/gardenlinux/blob/main/docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md)

PR #3639, opened 2025-10-10
tests-ng: Implement test for capabilities

fwilhe:
This is a test in multiple features. The new test works like the old one but uses more native python API. For features that don't have any expected capabilities defined, the expectation is that none are set.

Fixes https://github.com/gardenlinux/gardenlinux/issues/3588
Fixes https://github.com/gardenlinux/gardenlinux/issues/3615

Tracks https://github.com/gardenlinux/gardenlinux/issues/3584


tests-ng/test_capabilities.py
```diff
@@ -0,0 +1,23 @@
...
+import pytest
+
+from plugins.capabilities import Capabilities
+
+# There are two debian packages that provide an 'arping' binary
+# https://packages.debian.org/search?searchon=contents&keywords=arping&mode=path&suite=stable&arch=any
+# /usr/bin/arping is provided by iputils-arping
+# /usr/sbin/arping is provided by arping
+# We only use iputils-arping
+capabilities_allowlist = set(["/usr/bin/arping cap_net_raw=ep"])
```
ByteOtter: *Nitpick* (feel free to ignore): Instead of casting a list here, it's possible to use the direct set expression here as the set is not empty:

```suggestion
capabilities_allowlist = {"/usr/bin/arping cap_net_raw=ep"}
```

tests-ng/plugins/capabilities.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+from .shell import ShellRunner
+
+class Capabilities:
+    def __init__(self, shell: ShellRunner):
+        self._shell = shell
+
+    def list(self) -> list[str]:
```
ByteOtter: *Nitpick* (feel free to ignore): why not return a `set[str]` here and avoid having to cast 2 times?
```suggestion
    def list(self) -> set[str]:
```

tests-ng/test_capabilities.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import Path
```
ByteOtter: Unused import here.
```suggestion
```

tests-ng/plugins/capabilities.py
fwilhe: I wanted to return a list here because the function is called list, but calling the function 'set' would sound weird and non-descriptive to me. not sure.

tests-ng/plugins/capabilities.py
```diff
@@ -0,0 +1,13 @@
...
+        return self._shell("find /boot /etc /usr /var -type f -exec /usr/sbin/getcap {} \\;", capture_output=True).stdout.splitlines()
```
yeoldegrove: Might I suggest to get rid of the `find` call and doing the file operations in pure python?
You could look into [os.getxattr](https://docs.python.org/3/library/os.html#os.getxattr). Looks like this would allow to read the capabilities with python, too.

yeoldegrove (changes requested):

tests-ng/plugins/capabilities.py
fwilhe: I initially had some python implementation to traverse the directory tree and call getcap on the files, but that caused issues in CI, I'm assuming because it was very slow. I'm sure this can be done in a more efficient way, but my conclusion was that probably nothing I'll do will be faster than this shell script, so I decided to go with that.

I looked for a native python way to get capabilities but did not come across getxattr, that looks interesting for sure.

I don't mind spending some extra time trying to get rid of shell here, but imo this would also be a justified usage as it is a simple command that works quite fast.

tests-ng/plugins/capabilities.py
ByteOtter: How about `get()`? Because you get the capabilities of a file.... would also rhyme with set

tests-ng/plugins/capabilities.py
yeoldegrove: Implementing this yourself with os.getxattr would need a translation from the raw binary format that is returned. e.g. https://github.com/cptpcrd/pyprctl could be used to handle that or one could implement it like this:

```
    def _parse_capabilities(self, file_path: str, cap_data: bytes) -> str | None:
        """Parse capability data and return formatted string like getcap.

        os.getxattr() returns raw binary capability data from the 'security.capability'
        extended attribute. This data contains:
        - Version (4 bytes): Capability format version
        - Effective set (4 bytes): Currently active capabilities (bitmask)
        - Permitted set (4 bytes): Allowed capabilities (bitmask)
        - Inheritable set (4 bytes): Capabilities inherited by child processes (bitmask)

        Reference: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/capability.h
        """
        cap_names = [
            'cap_chown', 'cap_dac_override', 'cap_dac_read_search', 'cap_fowner',
            'cap_fsetid', 'cap_kill', 'cap_setgid', 'cap_setuid', 'cap_setpcap',
            'cap_linux_immutable', 'cap_net_bind_service', 'cap_net_broadcast',
            'cap_net_admin', 'cap_net_raw', 'cap_ipc_lock', 'cap_ipc_owner',
            'cap_sys_module', 'cap_sys_rawio', 'cap_sys_chroot', 'cap_sys_ptrace',
            'cap_sys_pacct', 'cap_sys_admin', 'cap_sys_boot', 'cap_sys_nice',
            'cap_sys_resource', 'cap_sys_time', 'cap_sys_tty_config', 'cap_mknod',
            'cap_lease', 'cap_audit_write', 'cap_audit_control', 'cap_setfcap',
            'cap_mac_override', 'cap_mac_admin', 'cap_syslog', 'cap_wake_alarm',
            'cap_block_suspend', 'cap_audit_read', 'cap_perfmon', 'cap_bpf',
            'cap_checkpoint_restore'
        ]

        try:
            if len(cap_data) >= 16:  # Capability v3 format
                version, effective, permitted, inheritable = struct.unpack('<IIII', cap_data[:16])
            elif len(cap_data) >= 12:  # Capability v1 format
                version, effective, permitted = struct.unpack('<III', cap_data[:12])
                inheritable = 0
            else:
                return None

            capabilities = []
            for i, cap_name in enumerate(cap_names[:32]):  # Only handle first 32 capabilities
                cap_flags = []
                # If effective is set, it implies permitted
                if effective & (1 << i):
                    cap_flags.append('e')
                    cap_flags.append('p')
                elif permitted & (1 << i):
                    cap_flags.append('p')
                if inheritable & (1 << i):
                    cap_flags.append('i')

                if cap_flags:
                    capabilities.append(f"{cap_name}={''.join(cap_flags)}")

            return f"{file_path} {' '.join(capabilities)}" if capabilities else None

        except (struct.error, IndexError):
            return None

    def list(self) -> list[str]:
        """List files with capabilities in /boot, /etc, /usr, and /var directories."""
        capabilities = []
        for directory in ['/boot', '/etc', '/usr', '/var']:
            if os.path.exists(directory):
                for root, dirs, files in os.walk(directory):
                    for file in files:
                        file_path = os.path.join(root, file)
                        try:
                            cap_data = os.getxattr(file_path, 'security.capability')
                            formatted = self._parse_capabilities(file_path, cap_data)
                            if formatted:
                                capabilities.append(formatted)
                        except OSError:
                            continue
        return capabilities
```

Pulling in pyprctl would be easier as we could directly use https://pyprctl.readthedocs.io/en/latest/#pyprctl.FileCaps.get_for_file which does most of the mapping magic.

yeoldegrove: For completeness sake, here his how an implementation could look like with `pyprctl`:

```
    def _get_capabilities(self, file_path: str) -> str | None:
        """Get file capabilities using pyprctl library.

        Uses pyprctl.FileCaps.get_for_file() to retrieve file capabilities
        and formats them to match getcap output format.
        """
        try:
            file_caps = FileCaps.get_for_file(file_path)
            if not file_caps or not file_caps.effective:
                return None

            # Format capabilities like getcap output
            capabilities = []
            for cap_name in file_caps.effective:
                cap_flags = []
                if file_caps.effective and cap_name in file_caps.effective:
                    cap_flags.append('e')
                if file_caps.permitted and cap_name in file_caps.permitted:
                    cap_flags.append('p')
                if file_caps.inheritable and cap_name in file_caps.inheritable:
                    cap_flags.append('i')

                if cap_flags:
                    capabilities.append(f"{cap_name}={''.join(cap_flags)}")

            return f"{file_path} {' '.join(capabilities)}" if capabilities else None

        except Exception:
            return None

    def list(self) -> list[str]:
        """List files with capabilities in /boot, /etc, /usr, and /var directories.
        """
        capabilities = []
        for directory in ['/boot', '/etc', '/usr', '/var']:
            if os.path.exists(directory):
                for root, dirs, files in os.walk(directory):
                    for file in files:
                        file_path = os.path.join(root, file)
                        try:
                            formatted = self._get_capabilities(file_path)
                            if formatted:
                                capabilities.append(formatted)
                        except OSError:
                            continue
        return capabilities
```

nkraetzschmar: I think I slightly prefer the python approach, assuming this is nicely abstracted away.
We probably have multiple places in various tests that need to do some operation for all files, right? Then a fixture to just get an iterator of all files (matching some pattern, in depth first or last order, some mount point only (analog to `find -xdev`), and similar options). So essentially a find fixture written in python, where the user of this fixture does not need to worry about any tree walk operations but can just do a simple

```python
def test_using_find(find):
    find.same_mnt_only = True
    for file in find:
        assert not file.endswith(".zip")

```

nkraetzschmar:
> I like that suggestion, but will the find / -xdev method not run into the same issue with extracted container images on /usr?

yes, a) and b) are two separate problems. we'll likely need some way to get all mount points we wanna consider plus some explicit list of directories to exclude

tests-ng/plugins/capabilities.py
yeoldegrove: I really did not think about the angle that other tests also want to find files in some way... If we had such a find fixture the test or a plugin would indeed only have to implement the _get_capabilities part (e.g. with pyprctl) and the test code could be pretty slick. 
I really would like that as it would potentially benefit other tests.

fwilhe: @yeoldegrove did you ever run your `_get_capabilities` code or is this untested copilot usage? It does not work and I'm not up for fixing that now. Reverted to using `getcap` for now.

tests-ng/plugins/capabilities.py
```diff
@@ -0,0 +1,36 @@
...
+class Capabilities:
+    def __init__(self, find: Find, shell: ShellRunner):
+        self._find = find
+        self._shell = shell
+
+    def get(self) -> set[str]:
+        capabilities = set()
+
+        self._find.same_mnt_only = True
+        self._find.root_paths = ["/boot", "/etc", "/usr", "/var"]
```
yeoldegrove, line 17: Maybe we could pass this as a parameter to not hide test logic inside the plugin?

We could simply update the test and not the plugin later if these paths need to be extended.

fwilhe: The idea of putting it here was that this is actually not a concern of the test, but an implementation detail. The test wants to search 'everywhere', and the plugin decides what that is. I am open to changing that if we agree that it is better; but really my current view is that here is the right place for that.

Also of course I remember @nkraetzschmar's comment about extracted container filesystems, so we will have to revise this, but I would like to keep it in smaller increments as this PR is already not small and does implement roughly the same test we had before.

nkraetzschmar (approved):

tests-ng/plugins/capabilities.py
nkraetzschmar: yeah I think for now this is fine. eventually I'd like to go the route of making this automatically pick up any "real" mount points (however we choose to define real mount points) and then doe something akin to `find -xdev` on those, but for now this is probably fine.

nkraetzschmar: once this is merged I can probably already do a trial run against gardener using the WIP implementation for #3693 so we at least get a vague idea of how much this is going to fail with on disk container images.

yeoldegrove: Alright. Let's merge this and improve it over time.

PR #3646, opened 2025-10-13
attempting azure test_azure_times.py, please bear with me initially

### LINKED ISSUE #3594: Migrate `feature/azure` tests
- [ ] azure/test/test_azure_times.py::test_startup_time <- ../features/azure/test/test_azure_times.py
- [x] azure/test/test_packages_musthave.py::test_packages_musthave <- helper/tests/packages_musthave.py ❌[Removed](https://github.com/gardenlinux/gardenlinux/blob/main/docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md)

saubhikdattagithub:
attempting to migrate features/azure/test_azure_times.py, please bear with me initially to get in line with expectations :-) 

Fixes #3594

### DIFF
--- a/tests-ng/test_azure_times.py
+++ b/tests-ng/test_azure_times.py
@@ -5,7 +5,7 @@ from plugins.systemd import Systemd
 @pytest.mark.root(reason="To start the systemd service")
 @pytest.mark.modify(reason="Starts systemd service")
 @pytest.mark.performance_metric
-@pytest.mark.feature("Azure instance to test boot time for kernel, initrd and userspace") 
+@pytest.mark.feature("azure", reason="Azure instance to test boot time for kernel, initrd and userspace") 
 def test_startup_time(systemd: Systemd):
     """System startup time must be within tolerated limits."""
     tolerated_kernel = 30.0


features/azure/test/test_azure_times.py
```diff
@@ -3,27 +3,40 @@
 from helper.sshclient import RemoteClient
 
 
+@pytest.mark.booted(reason="Requires a fully booted system with systemd")
```
fwilhe: all new tests are located in the `tests-ng` directory. please create a new file there to implement the test

saubhikdattagithub: Yes, my bad, reworked on the script, kindly review [47932e5](https://github.com/gardenlinux/gardenlinux/pull/3646/commits/47932e5f7e9ffb590f10c21d09109842c55a4586)

tests-ng/test_azure_times.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+from plugins.systemd import Systemd
+
+@pytest.mark.booted(reason="Test runs systemd")
+@pytest.mark.root(reason="To start the systemd service")
+@pytest.mark.modify(reason="Starts systemd service")
+@pytest.mark.performance_metric
+@pytest.mark.feature("Azure instance to test boot time for kernel, initrd and userspace")
```
fwilhe: check other example of the `feature` marker for reference. It should contain a boolean expression, not a prose sentence. You can provide a reason using the reason argument, that is optional.

fwilhe: this is the reason you are getting this error in ci 

> INTERNALERROR> boolean.boolean.ParseError: Invalid symbols sequence such as (A B) for token: "instance" at position: 13


We have a library in place to evaluate if a test should be run based on which features are in the image. there are examples for how to use that in other tests.

saubhikdattagithub: yes, didnt read the mark.features section well in the well explained README.md earlier, added mark.feature=("azure", reason="---") now, check running, will keep an eye

tests-ng/test_azure_times.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+from plugins.systemd import Systemd
+
+@pytest.mark.booted(reason="Test runs systemd")
+@pytest.mark.root(reason="To start the systemd service")
+@pytest.mark.modify(reason="Starts systemd service")
+@pytest.mark.performance_metric
+@pytest.mark.feature("azure", reason="Azure instance to test boot time for kernel, initrd and userspace") 
+def test_startup_time(systemd: Systemd):
```
fwilhe, line 9: [here](https://github.com/gardenlinux/gardenlinux/blob/dfb9183d433c48c5f238305198009acd25a7c195/tests-ng/test_basics.py#L44) we exclude azure for being too slow, but then in this test we have lower thresholds? That does not make sense.

@saubhikdattagithub please sync with @nkraetzschmar on the reasoning why azure was excluded in that other test and what this means for the test you're writing here.

nkraetzschmar: I guess the easiest option would be just to do a trial run with the `not azure` removed in the `test_startup_time` maybe azure boot up is no longer painfully slow, but historically we always had to exclude azure form any boot time tests

saubhikdattagithub: With assistance from @yeoldegrove, the `./test-ng --cloud azure <IMAGE>` is now possible via MacOS to execute tests in Azure.

A local change with `@pytest.mark.feature ("server") # server installs systemd and azure has notoriously bad startup times` in tess-ng/test_basics.py test in Azure passes fine as per screenshot below:

```bash
    tolerated_kernel = 60.0
    tolerated_userspace = 40.0
```

image

image

Kindly verify if this suffices the requirement, if yes, then would proceed with #3663 to merge

fwilhe: I can't tell how reliable this will be on azure, but if it works then we should go ahead with #3663 and merge that, and close this one

Issue #3567, opened 2025-10-01
make containerd tests more resilient ecr pull fails


fwilhe:
Recently we get ci fails with the following error:

```
ERROR test_containers.py::test_basic_container_functionality[public.ecr.aws/docker/library/busybox:unstable-uclibc] - RuntimeError: command ctr image pull public.ecr.aws/docker/library/busybox:unstable-uclibc failed with exit code 1
```

This might be related to rate limiting, or just some general issue with ecr

maybe providing credentials might help (if it is rate limiting)

fwilhe:
looks like aws allows one pull/second w/o authentication https://docs.aws.amazon.com/AmazonECR/latest/public/public-service-quotas.html

yeoldegrove:
Dockerhub also has quotas that I hit regularly hit when running tests-ng locally. We should use a registry that has no quotas or have to come up with a smart idea to prevent hitting the quota.
One idea would be using GHCR.io and GITHUB_TOKEN (could also be used locally). The ugly part is getting this token inside the test-distribution I guess.

vivus-ignis:
Blocked by: https://github.com/gardenlinux/gardenlinux/issues/3705

PR #3660, opened 2025-10-14
3567 make containerd tests more resilient ecr pull fails

vivus-ignis:

Fixes https://github.com/gardenlinux/gardenlinux/issues/3567

This sets up a mocked container registry running in a separate thread for "ctr pull tests". Busybox container tarball (created with docker save from this https://hub.docker.com/layers/library/busybox/1.37.0-uclibc/images/sha256-3ae94dbd6a2350124f7c5199a17d5c3df9ad7913c08c97880c8c7637c872102d) is used for the minimal container to be server by the mocked registry.


tests-ng/plugins/busybox_amd64.tar
nkraetzschmar, line 1: commiting binaries to git should be a no go

nkraetzschmar (changes requested):

tests-ng/plugins/busybox_amd64.tar
vivus-ignis: then we would have to download them from a well-known secure location without throttling (as this very change tries to fix the throttling problem when pulling images from public registries). and this means our tests rely on one more external resource, making the whole system more fragile and unnecessarily complex. 

> commiting binaries to git should be a no go

nkraetzschmar:
> This sets up a mocked container registry running in a separate thread for "ctr pull tests"

I don't think that is a particularly great idea..... or at the very least that should be done via a proper registry running in a container of its own, not re-implementing a container registry in python.....

Obviously then this just moves the problem of where to fetch the initial registry container

fwilhe:
I am getting this error when trying to run it locally

```
./test-ng .build/kvm-chost-arm64-2030.0.0-local.raw
Building test distribution...
🛠️  building python runtime
/bin/sh: command substitution: line 0: syntax error near unexpected token `)'
/bin/sh: command substitution: line 0: `podman pull -q --arch=amd64 busybox:1.37.0-uclibc)'
make: *** [.build/busybox_amd64.tar] Error 1
```

fwilhe:
@nkraetzschmar @vivus-ignis 

Did you align on the goal? My proposal would be to proceed with this PR for now, but create an issue for the advanced configuration option to do that later. This would allow us to make this PR more flexible, swap out the images and also provide registry credentials which would be useful for integration scenarios where you want to test a certain registry.

fwilhe:
On hold as discussed

PR #3663, opened 2025-10-16
check server boot time for azure

saubhikdattagithub:
Check boot time of Azure as recommended here https://github.com/gardenlinux/gardenlinux/pull/3646#discussion_r2432664842

Fixes #3646


tests-ng/test_basics.py
```diff
@@ -52,8 +52,8 @@ def test_fhs(shell: ShellRunner):
 )
 @pytest.mark.performance_metric
 @pytest.mark.feature(
-    "server and not azure"
-)  # server installs systemd and azure has notoriously bad startup times
+    "server"
+)  # server installs systemd and azure has notoriously bad startup times #
```
fwilhe: the comment is outdated, and also it should not be in a comment but provided as a reason like this

```
@pytest.mark.feature("server", reason="server installs systemd")
```

Issue #3001, opened 2025-04-29
Define and document information management strategy


pnpavlov:
**Goal:** Provide in markdown format a information management strategy highlighting what content must be stored in what repository. Keep the amount of repositories to minimal. Garden Linux Maintainers must understand: 
- where to publish documentation related to Garden Linux (e.g. GL end-user features description, developer docu, LF, company internal data)
- where to publish documentation related to individual packages
- where to publish infrastructure related information
- where to store public and private publishing docu

**Tasks:**
- Prepare a proposal.
- Present it to the team to collect feedback, make sure everyone understands the required steps.
- Publish the final document in repository with adequate access based on the defined strategy.

tmang0ld:
Will be handle in issue [4332](https://github.com/gardenlinux/gardenlinux/issues/4332)

PR #3664, opened 2025-10-16
test-ng: add Developer Guidelines

### LINKED ISSUE #3582: Develop and present next gen test development guidelines
Goal is to develop and present to the next generation test framework test contributors a reusable easy to follow guideline including how to structure tests, how to re-use common functionality and how to review pull requests

yeoldegrove:

This is the state of discussion and feedback in https://github.com/gardenlinux/gardenlinux/issues/3582.
Adds "Garden Linux Tests Next Generation - Developer Guide" to the tests-ng directory.

Fixes #3582

### DIFF
--- a/tests-ng/DEVELOPER.md
+++ b/tests-ng/DEVELOPER.md
@@ -44,11 +44,14 @@ The following principles guide all test development in Garden Linux:
 - Document (`reason=`) why root access is required
 - Prefer unprivileged testing when possible
 
-5. **Only require a booted system when needed**
+5. **Target appropriate test environments and platforms**
 
-- Use `@pytest.mark.booted` only when a running system is required
-- Document why a booted system is necessary
-- Prefer chroot testing for filesystem-only tests
+- Use `@pytest.mark.booted` to mark tests that require a full booted system (such as QEMU or cloud VMs) to function correctly.
+- Use `@pytest.mark.feature` to restrict tests to only those environments or platforms where they are intended to run, especially if they would fail elsewhere.
+- Whenever possible, design your tests so that they work across _all_ supported environments and platforms. Only exclude an environment or platform if there’s a strong, well-documented reason to do so.
+- If possible, still add a minimal test for excluded platforms
+- Document (`reason=`) why a test must run (or be excluded) in certain environments or platforms
+- For a list of all available test environments (like chroot, QEMU, cloud, and OCI), see [Test Environment Details](../README.md#test-environment-details).
 
 6. **Use abstractions judiciously to hide implementation details**
 
@@ -314,6 +317,44 @@ Limits test execution based on feature conditions:
                      reason="We want no authorized_keys for unmanaged users")
 def test_users_have_no_authorized_keys(expected_users):
     # Test implementation
+```
+
+### Common Filtering Patterns
+
+**Environment-specific filtering:**
+
+```python
+# Only run on booted systems (QEMU, cloud)
+@pytest.mark.booted(reason="Requires running system")
+
+# Skip in container environments (OCI)
+@pytest.mark.feature("not container", reason="Container environment not suitable")
+```
+
+**Platform-specific filtering:**
+
+```python
+# Cloud provider specific tests
+@pytest.mark.feature("aws", reason="AWS-specific functionality")
+@pytest.mark.feature("azure", reason="Azure-specific functionality")
+@pytest.mark.feature("gcp", reason="GCP-specific configuration")
+@pytest.mark.feature("ali", reason="Alibaba Cloud-specific configuration")
+
+# Other platform-specific tests
+@pytest.mark.feature("openstack", reason="OpenStack-specific configuration")
+@pytest.mark.feature("vmware", reason="VMware-specific configuration")
+```
+
+**Complex environment combinations:**
+
+```python
+# Multiple conditions
+@pytest.mark.feature("(gardener or chost) and not _pxe",
+                     reason="containerd needed but not working in PXE environment")
+
+# Feature combinations
+@pytest.mark.feature("ssh and not (ali or aws or azure or openstack)",
+                     reason="We want no authorized_keys for unmanaged users")
 ```
 
 #### `@pytest.mark.performance_metric`


tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,490 @@
...
+- **Pattern**: Yield fixtures that prepare resources and explicitly clean up after tests
+- **Usage**: Used as pytest fixtures with `yield` for cleanup
+- **Examples**: `service_ssh`, `service_containerd`
+
+**Key distinction**: Unlike regular fixtures that provide data, handlers manage stateful resources that need explicit cleanup.
+
+```python
+# Example handler usage - setup/teardown pattern
+@pytest.fixture
+def service_ssh(systemd: Systemd):
```
ByteOtter: Maybe this goes too far, but sometimes it may be a good thing to set up a naming convention for handlers to distinguish them from normal features?
Something like `service_*` or `*_handler()`?

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,490 @@
...
+
+**Guidelines:**
+
+- Follow [PEP 8 – Style Guide for Python Code](https://peps.python.org/pep-0008/)
+- Use type hints for function parameters and return values
+- Prefer descriptive variable names
+- Use docstrings for complex functions
+
+> [!TIP]
+> Run `make -f tests-ng/dev.makefile format` locally before committing
```
ByteOtter, line 426: I'd be pretty much in favor to introduce `.pre-commit` to help with that. Open for discussions.

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,490 @@
...
+- **Document why** external dependencies are necessary
+- **Consider maintenance cost** of external dependencies
+
+### Current Dependencies
+
+See `tests-ng/util/requirements.txt` for current dependencies. Each dependency should be justified.
+
+### Adding New Dependencies
+
+When adding new dependencies:
```
ByteOtter, line 491: With external dependencies we also need to make sure that the versions we are using a safe to use and actively maintained.
In one of my projects, we use a `cargo-audit` CI to help us with that, maybe something similar exists for python and would be helpful with that?

yeoldegrove: https://github.com/gardenlinux/gardenlinux/blob/8d15eb4f9c14870273e43f812758fb97c7646973/.github/dependabot.yml#L20-L26 should already handle updates to our dependencies.

Maybe adding a comment for maintained dependencies is enough, then?

tests-ng/DEVELOPER.md
yeoldegrove: If you have a snippet to introduce this, we could add it here I guess. If it is too complex, we need to put it somewhere else.

tests-ng/DEVELOPER.md
yeoldegrove: Not sure how strict we should be here. I guess our view on that could change as we add more handlers. but naming those `*_handler()` sounds good to me currently as we could easily distinguish those functions from the ones that are in the plugins.

tests-ng/DEVELOPER.md
ByteOtter: Ah, right ... completely forgot about our bot. Yeah should be enough for python. A comment about a dependency receiving regular maintenance should suffice then :)

tests-ng/DEVELOPER.md
ByteOtter: I agree. If we change our minds later it's easy enough to fix. But for now I think this would be beneficial for us and outside contributors.

tests-ng/DEVELOPER.md
ByteOtter: it's a config file for the repo and a tool the contributor has to have installed. Maybe this is worth going into depth about in a follow up PR about general quality enforcement though.

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,490 @@
...
+- **Only require a booted system when needed**
+
+  - Use `@pytest.mark.booted` only when a running system is required
+  - Document why a booted system is necessary
+  - Prefer chroot testing for filesystem-only tests
+
+- **Use abstractions judiciously to hide implementation details**
+
+  - Leverage plugins for infrastructure concerns (parsing files, accessing data, establishing connections)
+  - Use handlers for setup/teardown operations
```
moritzkirchnersap: The comment lower seems like "handlers" is a common name for "yield fixtures" - is that the case? Please add (yield fixtures) here as well if that's the case.
Do you have an online resource? I'd like to read about it.

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,490 @@
...
+```python
+@pytest.fixture
+def service_ssh(systemd: Systemd):
+    """Fixture for SSH service management with cleanup."""
+    service_active_initially = systemd.is_active("ssh")
+
+    if not service_active_initially:
+        systemd.start_unit("ssh")
+
+    yield "ssh"  # Provide service name to test
```
fwilhe: what is that string used for? is this actually relevant? should be mentioned here. 'provide service name to test' leaves open questions imo

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,490 @@
...
+Handlers are pytest fixtures that manage test setup and teardown:
+
+- **Purpose**: Setup/teardown of test state (connections, services, environments)
+- **Pattern**: Yield fixtures that prepare resources and explicitly clean up after tests
+- **Usage**: Used as pytest fixtures with `yield` for cleanup
+- **Examples**: `service_ssh`, `service_containerd`
+
+**Key distinction**: Unlike regular fixtures that provide data, handlers manage stateful resources that need explicit cleanup.
+
+```python
```
fwilhe: this example is repeating from above?

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,490 @@
...
+@pytest.mark.performance_metric
+@pytest.mark.booted(reason="We can only measure startup time if we actually boot the system")
+def test_startup_time(systemd: Systemd):
+    # Test implementation
+```
+
+### Parametrization
+
+Use `@pytest.mark.parametrize` for testing multiple scenarios:
```
fwilhe: I don't think this code sample is good, and also this section should reflect the tradeoff discussion we had for parameterised tests.

in short: use, but not over-use parameters

the goal is not to have super generic test functions. if a test would be duplicated to be almost identical, and the differences can be easily extracted into a parameter, the do that, but if you are creating something complex just for the sake of avoiding a bit code duplication, then maybe code duplication **in tests** is acceptable

tests-ng/DEVELOPER.md
yeoldegrove: @fwilhe We could simply leave out explaining this. But as the nvme handler is already using this feature, maybe we should very briefly explain it.

What do you think, is this sufficient?

```
    yield "ssh"  # This returns "ssh" to the test as the fixture's value. This can be use to parametrize tests.
```

tests-ng/DEVELOPER.md
yeoldegrove: It was meant to have a different purpose as both examples are in different sections.

My current idea would be moving the "Framework Structure" before the "Test Writing Best Practices" and move all examples to later one.
I'll try that one out and come back for feedback.

tests-ng/DEVELOPER.md
yeoldegrove: @fwilhe What is your take on enforcing/suggesting `*_handler()` naming in the "Test Organization and Naming" section? If you don't object, I would add a small section.

tests-ng/DEVELOPER.md
yeoldegrove: Reworked in fa7957ccca581eafb5a13ad4247d40680d26727d

tests-ng/DEVELOPER.md
yeoldegrove: Reworked in 2ef25bee304af13866591d0aecbe7bbaddc47eba

tests-ng/DEVELOPER.md
yeoldegrove: Reordered this in 2ac24a1b431b95ae7ed268fab635f7a69c01c9b8

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,446 @@
...
+
+  - Leverage plugins for infrastructure concerns (parsing files, accessing data, establishing connections)
+  - Use handlers for setup/teardown operations
+  - Keep test logic visible and maintain Arrange-Act-Assert structure
+  - Avoid over-abstraction that requires reading multiple plugins to understand a test
+
+- **Be mindful about external dependencies**
+  - Prefer Python standard library over third-party packages
+  - Only add PyPI dependencies when there's clear benefit
+  - Document why external dependencies are necessary
```
vivus-ignis: It makes sense to mention how do we deal with external resources (like container registries, S3 buckets etc). 
For example, everything we use for a cloud test, should be deployed and later cleaned up by terraform. 
This also concerns mocks, like in this PR: https://github.com/gardenlinux/gardenlinux/pull/3660. I personally think that mocks are preferable to using external resources (because we are not testing external resources per se).

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,446 @@
...
+See `tests-ng/util/requirements.txt` for current dependencies. Each dependency should be justified.
+
+### Adding New Dependencies
+
+When adding new dependencies:
+
+1. **Justify the need** - Why can't the standard library solve this?
+2. **Document the benefit** - What does this library provide?
+3. **Consider alternatives** - Are there lighter alternatives?
+4. **Update requirements.txt** - Include version constraints
```
vivus-ignis: This should be probably rephrased to "the package version should be explicitly pinned".

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,446 @@
...
+- **[Black](https://black.readthedocs.io/en/stable/)**: Automatic code formatting
+- **[isort](https://pycqa.github.io/isort/)**: Import statement sorting
+- **[Pyright](https://github.com/microsoft/pyright/blob/main/docs/getting-started.md)**: Static type checking
+
+**Guidelines:**
+
+- Follow [PEP 8 – Style Guide for Python Code](https://peps.python.org/pep-0008/)
+- Use type hints for function parameters and return values
+- Prefer descriptive variable names
+- Use docstrings for complex functions
```
ByteOtter: I think this wording is too lenient here. There are countless reasons why we should follow PEP 8 and python best practices as much as possible. So I would like to rephrase this more strongly.


```suggestion
1. All code contributions must follow the [PEP 8 – Style Guide for Python Code](https://peps.python.org/pep-0008/)
2. The use of type hints is mandatory for all new additions to avoid typing related bugs and increase readability
3. Use descriptive variable and function names
4. Use docstrings to summarize functions and classes. The extent of which depends on the size, scope and complexity of the function or class.
5. Initialize all variables with a sensible default value instead of using `None` (unless the value and its type hint explicitly make it noneable)
```

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,446 @@
...
+## Python Best Practices
+
+### Code Style and CI Enforcement
+
+The project enforces code quality through CI linting (see `.github/workflows/lint_tests-ng.yml`):
+
+- **[Black](https://black.readthedocs.io/en/stable/)**: Automatic code formatting
+- **[isort](https://pycqa.github.io/isort/)**: Import statement sorting
+- **[Pyright](https://github.com/microsoft/pyright/blob/main/docs/getting-started.md)**: Static type checking
+
```
ByteOtter, line 411: I would like to point people at this point to enable the following in their local IDE or text editor:

- A language server (pylance, Jedi, etc) available through their IDE's language plugins
- The Black formatter extension (available for PyCharm, VSCode, Codium, NeoVim, Vim, Atom, Emacs (I think), etc.)
- The isort extension (available for PyCharm, VSCode, Codium, NeoVim, Vim, etc.)
- The pyright type checker (available for PyCharm, VSCode, Codium, NeoVim, Vim, etc.)

And to enable " apply formatting (and type checking) on save".
Happy to share my Codium and NeoVim configs for that or write a guide somewhere else.

I think this would make all our lives easier.

```suggestion
These tools are available for most IDEs and a wide range of text editors.

In order to improve your development experience, and to allow us to process your PRs faster, we suggest your enable them with your editor or IDE.

Please refer to your editor's documentation for clues on how to set them up.

We highly suggest you configure your IDE or text editor to automatically apply formatting and type checking on save.
```

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,446 @@
...
+### Error Handling
+
+- Use descriptive assertion messages
+- Handle expected failures gracefully
+- Use appropriate exception types
+
+```python
+def test_config_file_exists():
+    config_path = Path("/etc/ssh/sshd_config")
+    assert config_path.exists(), f"SSH config file not found: {config_path}"
```
ByteOtter: I personally think this is a good example for a short test, but not a good one for error handling.

I'd go with something more like this (doesn't have to be that explicit, we can also just go with catching `OSError as exc` and printing `exc` with the error message:


```suggestion
def read_file(path: str) -> str:
    """Read and return the contents of a file safely."""
    try:
        with open(path, "r", encoding="utf-8") as file:
            return file.read()
    except FileNotFoundError:
        print(f"Error: File not found — {path}")
    except PermissionError:
        print(f"Error: Permission denied when accessing {path}")
    except OSError as exc:
        print(f"Error: Could not read file {path}: {exc}")
```

Also I would like to print to `stderr` instead of `stdout` in such cases, but python is not great with this, so in the end, as long as we log errors it doesn't really matter.

ByteOtter (commented):
Some more thoughts regarding code style. Just my opinion, but happy do discuss.

tests-ng/DEVELOPER.md
yeoldegrove: Addressed in 7e082b01ca999008939b994a46f4c2e8aec7f14d

tests-ng/DEVELOPER.md
yeoldegrove: I guess as long as discussions and improvements in this domain are ongoing (like https://github.com/gardenlinux/gardenlinux/issues/3705) we can skip documenting this. As soon as we have a common understanding, we should add it.
@vivus-ignis @fwilhe Are you ok with skipping this for now? If not, please phrase something to add here, please.

vivus-ignis: Ok to skip.

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,461 @@
...
+    # ...too much unrelated logic for one test
+```
+
+**Guideline:** Prefer clarity and intent over DRY-ness in tests. Parametrize when it makes tests simpler, not just shorter.
+
+### Missing Markers (TODO)
+
+#### `@pytest.mark.security_id`
+
+TODO: Explain what this marker is good for.
```
fwilhe, line 399: @Akendo can you provide a useful paragraph for this?

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,461 @@
...
+  - Prefer chroot testing for filesystem-only tests
+
+- **Use abstractions judiciously to hide implementation details**
+
+  - Leverage plugins for infrastructure concerns (parsing files, accessing data, establishing connections)
+  - Use handlers for setup/teardown operations
+  - Keep test logic visible and maintain Arrange-Act-Assert structure
+  - Avoid over-abstraction that requires reading multiple plugins to understand a test
+
+- **Be mindful about external dependencies**
```
Akendo: Newline is missing.

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,461 @@
...
+- [External Dependencies](#external-dependencies)
+- [Resources](#resources)
+
+## Test Development Principles
+
+The following principles guide all test development in Garden Linux:
+
+### Core Principles
+
+- **Be easy to read and understand** without extensive knowledge of Garden Linux internals
```
Akendo: Please add numbers so we can reference them more effectively.

tests-ng/DEVELOPER.md
```diff
@@ -0,0 +1,461 @@
...
+  - Ensure tests clean up after themselves.
+    - If new functionality is added, check if `tests-ng/plugins/sysdiff.py` collects modifications.
+
+- **Only run as root when needed**
+
+  - Use `@pytest.mark.root` only when root privileges are absolutely necessary
+  - Document (`reason=`) why root access is required
+  - Prefer unprivileged testing when possible
+
+- **Only require a booted system when needed**
```
Akendo: Why? The dichotomy of tests we compose will aim for different stages of the testing phase we have. This is a goal that would mislead developers to conclude that they have to develop tests with chroot in mind. 

This might work for basic modification of the file system, but I would make the argument that this set of tests should be successful at any stage of the testing. It should be stated that the test should be separated based on the change it conducted on the underlying filesystem and that this needs to pass the chroot stage. 

However, I imagine that a developer would try to integrate tests to fulfill this here. I am not sure if that would be good, since my understanding of how this rule was composed is that I have to invoke the target testing binary within the chroot to ensure that its test is correctly conducted. This is a nightmare from the developer's perspective, and I think this is not the way to go.

tests-ng/DEVELOPER.md
Akendo: Yes, we might have to rethink how we're planning to separate our testing sets based on the target we need to test. But I might get something in here in a moment.

Akendo:
Foremost, great work! 

But going through this document, I feel like this is more a mix of different types of documents meshed. I’m using the nomenclature of Diataxis. But what I get here is a document that aims to provide a reference. However, the document does not provide good guidance on how to compose a test. Who is the audience for this document? Who should read this? I will not read this when I start the writing test, sorry, but no sorry.

And that makes sense, since the goal of this document is not to accomplish a goal but to inform the developer about the rules he should consider before committing. Hence, this document is great when I already have a good understanding of what I'm aiming for and need some support to see if that fits, but even then I wouldn’t go through this document to ‘check’ my test against. You also don't read through the constitution and then make corrections in your behavior, unless you’re a lawyer, maybe.

In short, reading this document overburdens me and feels more like an obstacle to starting testing.

However, we still need these documents. There is just not a guide. Here are some suggestions on how to improve:

Split the document into smaller parts. For instance, put the chapter “Test Development Principles” into a separate document. It contains our high-level values we might aim for during the development of our tests, things to aim for.

Then split the document at the “Framework Structure.” This is another reference since it tries to explain how the testing framework was organized. That's a great document that we can use for reference! It's just not a document that can live outside an established context.

tests-ng/DEVELOPER.md
yeoldegrove: fixed in 30b4085ae317355cabd95bc75964fec07f2c04d4

tests-ng/DEVELOPER.md
yeoldegrove: fixed in 30b4085ae317355cabd95bc75964fec07f2c04d4

tests-ng/DEVELOPER.md
yeoldegrove: The **Only require a booted system when needed** heading is a bit misleading and misses some context.

Added 6fb1019b6df8e35e19f85cd6795bff739a4cbeaf

I think we meant the same thing here and rephrased it to:

---

5. **Target appropriate test environments and platforms**

- Use `@pytest.mark.booted` to mark tests that require a full booted system (such as QEMU or cloud VMs) to function correctly.
- Use `@pytest.mark.feature` to restrict tests to only those environments or platforms where they are intended to run, especially if they would fail elsewhere.
- Whenever possible, design your tests so that they work across *all* supported environments and platforms. Only exclude an environment or platform if there’s a strong, well-documented reason to do so.
- If possible, still add a minimal test for excluded platforms.
- Document (`reason=`) why a test must run (or be excluded) in certain environments or platforms.
- For a list of all available test environments (like chroot, QEMU, cloud, and OCI), see [Test Environment Details](../README.md#test-environment-details).

---

I also added a new section to explain common filtering patterns:

---

## Markers and Test Configuration

### Common Filtering Patterns

**Environment-specific filtering:**

```python
# Only run on booted systems (QEMU, cloud)
@pytest.mark.booted(reason="Requires running system")

# Skip in container environments (OCI)
@pytest.mark.feature("not container", reason="Container environment not suitable")
```

**Platform-specific filtering:**

```python
# Cloud provider specific tests
@pytest.mark.feature("aws", reason="AWS-specific functionality")
@pytest.mark.feature("azure", reason="Azure-specific functionality")
@pytest.mark.feature("gcp", reason="GCP-specific configuration")
@pytest.mark.feature("ali", reason="Alibaba Cloud-specific configuration")

# Other platform-specific tests
@pytest.mark.feature("openstack", reason="OpenStack-specific configuration")
@pytest.mark.feature("vmware", reason="VMware-specific configuration")
```

**Complex environment combinations:**

```python
# Multiple conditions
@pytest.mark.feature("(gardener or chost) and not _pxe",
                     reason="containerd needed but not working in PXE environment")

# Feature combinations
@pytest.mark.feature("ssh and not (ali or aws or azure or openstack)",
                     reason="We want no authorized_keys for unmanaged users")
```

#### `@pytest.mark.performance_metric`

Marks performance tests that can be skipped under emulation:

```python
@pytest.mark.performance_metric
@pytest.mark.booted(reason="We can only measure startup time if we actually boot the system")
def test_startup_time(systemd: Systemd):
    # Test implementation
```

yeoldegrove:
@Akendo 

> But going through this document, I feel like this is more a mix of different types of documents meshed. I’m using the nomenclature of Diataxis. But what I get here is a document that aims to provide a reference. However, the document does not provide good guidance on how to compose a test. Who is the audience for this document? Who should read this? I will not read this when I start the writing test, sorry, but no sorry.
> 
The audience here is of course test developers as the file name suggests and as it is also stated in the first sentence. We currently have not decided on any other way of suggesting/categorizing audiences.

> And that makes sense, since the goal of this document is not to accomplish a goal but to inform the developer about the rules he should consider before committing. Hence, this document is great when I already have a good understanding of what I'm aiming for and need some support to see if that fits, but even then I wouldn’t go through this document to ‘check’ my test against. You also don't read through the constitution and then make corrections in your behavior, unless you’re a lawyer, maybe.
> 
The goal is to guide you through the process of developing and maintaining tests by giving you the basic boundaries. IMHO Didactically the table of contents fits here (but of course can be extended).

> In short, reading this document overburdens me and feels more like an obstacle to starting testing.
> 
I am sorry you feel that way. For me, not having any document is more of an obstacle.

> However, we still need these documents. There is just not a guide. Here are some suggestions on how to improve:
> 
> Split the document into smaller parts. For instance, put the chapter “Test Development Principles” into a separate document. It contains our high-level values we might aim for during the development of our tests, things to aim for.
> 
> Then split the document at the “Framework Structure.” This is another reference since it tries to explain how the testing framework was organized. That's a great document that we can use for reference! It's just not a document that can live outside an established context.
>

My initial idea actually was to rework `tests-ng/README.md` together with this new `tests-ng/DEVELOPER.md` document and even thought about adding a `tests-ng/ARCHITECTURE.md` ...
But I discarded that idea in favor of letting the people that already wrote tests:
1. first commit to a common understanding
2. a certain set of content
3. extend it with content that is still missing
 
I think this worked out quite well.

Documentation always is a moving target and for sure the next step would be splitting/creating those documents.
As long as we don't have a new documentation structure where we distinguish by Diataxis (also see https://github.com/gardenlinux/gardenlinux/issues/3001), it makes no sense to me to rework this again.

After we have established a new documentation structure, this could even be a good candidate for a prototype/template documentation but until then I want this merged the way it is structured right now.

Issue #3669, opened 2025-10-16
test-ng: rethink test_proc.py for container use case


yeoldegrove:
**What would you like to be added**:

Rethink test_proc.py for container use case.

**Why is this needed**:

A container does have /proc available but is not booted and hence not covered by the test yet.

PR #3679, opened 2025-10-20
test-ng: rework OCI tests to use container base image instead of bare flavors

### LINKED ISSUE #3577: Re-evaluate integration of OCI tests into bare_flavours
Merging of the bare flavours tests into test-ng for `.oci` artifacts seems conceptually inconsistent with how this test suite is intended to function. Bare flavours tests are not a tests-ng–based platform - it’s designed for a different testing context - so integrating it here feels misplaced.

We should absolutely have OCI test coverage, but that should primarily target the base image, not the bare flavours. Ideally, this would:

- Import the container image
- Bind-mount the test suite (rather than building another layer on top)
- Run the standard tests.

Mixing the tests-ng tests directly with bare flavours tests feels like an ad-hoc workaround just to maintain a single entry point, rather than a clean integration. This probably needs a clearer separation of purpose and test strategy

yeoldegrove:

Currently the OCI tests in `test-ng` integrated the **bare** flavors, which don’t align with `test-ng`’s pytest-driven model (they don’t run `pytest`). This mismatch complicates execution, reporting, and maintenance. We’re switching to **container base image** tests, which are fully runnable under `pytest` and fit `test-ng` paradigms.

* **Remove:**
  * Integration of OCI **bare** test flavors from `test-ng`.
* **Add:**
  * Integration of **container base image** OCI tests runnable via `pytest`.
* **Change:**
  * temporarily adapt `test_proc.py` for OCI containers. Follow up issue opened https://github.com/gardenlinux/gardenlinux/issues/3669
* **Github Actions:**
  * Our CI will still run **bare** test flavors and will be extended to also test the **container base image**. 
  * https://github.com/gardenlinux/gardenlinux/issues/3680 is tackled by 929fcf26eb64b930111cc1901616e85282cf1cbe which simplifies the `includes` matrix generation and fixes `matrix:` because it's fromJson() evaluation cannot be skipped by setting the matrix to an empty string or array hence we set a minimal skip flavor matrix and evaluate it in each test_flavor_* workflow.

**Notes for reviewers**:

Successful CI runs are here:
- `test_types: oci`: https://github.com/gardenlinux/gardenlinux/actions/runs/18708310630
- `test_types: chroot,oci,qemu`: https://github.com/gardenlinux/gardenlinux/actions/runs/18708314042
- `test_types: chroot,oci,qemu,cloud`: https://github.com/gardenlinux/gardenlinux/actions/runs/18719981214
- `test_types: oci,bare`: https://github.com/gardenlinux/gardenlinux/actions/runs/18708317440
- `test_types: cloud`: https://github.com/gardenlinux/gardenlinux/actions/runs/18719990835

Fixes #3577

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -108,7 +108,7 @@ jobs:
 
             const testsEnvsEnabled = gitHubLib.getTestEnvironmentsEnabled("${{ inputs.test_types }}");
 
-            const TESTS = [
+            const testDefinitions = [
               {
                 key: "chroot",
                 outTests: "chroot_tests",
@@ -147,31 +147,26 @@ jobs:
             ];
             // `matrix:` and it's fromJson() evaluation cannot be skipped by setting the matrix to an empty string or array
             // hence we set a minimal skip flavor matrix and evaluate it in each test_flavor_* workflow.
-            const SKIP_MATRIX = '{"include":[{"arch":"amd64","flavor":"skip"}]}';
+            const skipFlavorMatrix = '{"include":[{"arch":"amd64","flavor":"skip"}]}';
 
             // Process all test types, but only enable requested ones
-            for (const t of TESTS) {
-              console.log(`Processing test type: ${t.key}`);
+            for (const t of testDefinitions) {
               if (!testsEnvsEnabled.includes(t.key)) {
                 core.setOutput(t.outTests, false);
-                core.setOutput(t.outMatrix, SKIP_MATRIX);
-                console.log(`${t.key}: disabled`);
-                console.log(`${t.key}: matrix: ${JSON.stringify(SKIP_MATRIX)}`);
+                core.setOutput(t.outMatrix, skipFlavorMatrix);
+                console.debug(`${t.key}: disabled`);
                 continue;
               }
-              console.log(`Test type ${t.key} is not disabled`);
 
               const matrix = gitHubLib.intersectFlavorsMatrix(t.inputMatrix, t.supported);
               const isEnabled = !gitHubLib.isMatrixEmpty(matrix);
               core.setOutput(t.outTests, isEnabled);
-              console.log(`${t.key}: isEnabled: ${isEnabled}`);
+              console.log(`${t.key}: ${isEnabled ? 'enabled' : 'disabled'} (${isEnabled ? matrix.length : 0} flavors)`);
 
               if (isEnabled) {
                 core.setOutput(t.outMatrix, matrix);
-                console.log(`${t.key}: matrix: ${JSON.stringify(matrix)}`);
               } else {
                 core.setOutput(t.outMatrix, SKIP_MATRIX);
-                console.log(`${t.key}: matrix: ${JSON.stringify(SKIP_MATRIX)}`);
               }
             }
   ## test-ng start


.github/workflows/tests.yml
```diff
@@ -100,71 +108,70 @@ jobs:
...
-              );
-
-              const isEnabled = !gitHubLib.isMatrixEmpty(matrix);
-              core.setOutput("chroot_tests", isEnabled);
-
-              if (isEnabled) {
-                core.setOutput("chroot_test_flavors_matrix", matrix);
-              } else {
-                core.setOutput("chroot_test_flavors_matrix", '{"include":[]}');
+            const TESTS = [
```
NotTheEvilOne: JavaScript is not e.g. Python. Please do not upper-case variable names :)

.github/workflows/tests.yml
```diff
@@ -100,71 +108,70 @@ jobs:
...
-            // qemu_test_flavors_matrix
-            if (testsEnvsEnabled.includes("qemu")) {
-              const matrix = gitHubLib.intersectFlavorsMatrix(
-                ${{ inputs.flavors_matrix }},
-                ${{ needs.qemu_test_flavors_supported_matrix.outputs.matrix }}
-              );
+            ];
+            // `matrix:` and it's fromJson() evaluation cannot be skipped by setting the matrix to an empty string or array
+            // hence we set a minimal skip flavor matrix and evaluate it in each test_flavor_* workflow.
+            const SKIP_MATRIX = '{"include":[{"arch":"amd64","flavor":"skip"}]}';
```
NotTheEvilOne: Are you sure we can not use `{"include":[]}` here? An empty (but still valid) matrix is gracefully handled by GitHub tooling as far as I know.

.github/workflows/tests.yml
```diff
@@ -100,71 +108,70 @@ jobs:
...
-              } else {
-                core.setOutput("qemu_test_flavors_matrix", '{"include":[]}');
+            // Process all test types, but only enable requested ones
+            for (const t of TESTS) {
+              console.log(`Processing test type: ${t.key}`);
+              if (!testsEnvsEnabled.includes(t.key)) {
+                core.setOutput(t.outTests, false);
+                core.setOutput(t.outMatrix, SKIP_MATRIX);
+                console.log(`${t.key}: disabled`);
+                console.log(`${t.key}: matrix: ${JSON.stringify(SKIP_MATRIX)}`);
```
NotTheEvilOne: I think this is spurious from testing as a line saying "test: disabled" is enough for logging purposes. Alternatively you may want to use `console.debug()` here.

.github/workflows/tests.yml
```diff
@@ -100,71 +108,70 @@ jobs:
...
               }
-            }
-
-            // platform_test_flavors_matrix
-            if (testsEnvsEnabled.includes("cloud")) {
-              const matrix = gitHubLib.intersectFlavorsMatrix(
-                ${{ inputs.flavors_matrix }},
-                ${{ needs.platform_test_flavors_supported_matrix.outputs.matrix }}
-              );
+              console.log(`Test type ${t.key} is not disabled`);
```
NotTheEvilOne: Additional debug outputs follow I think. For a non-disabled test environment it should be enough to output the matrix generated or even only "enabled" as the matrix is seen in the GitHub actions afterwards.

.github/workflows/test_flavor_cloud_ng.yml
```diff
@@ -41,6 +41,7 @@ jobs:
   test:
     name: Test-NG flavor in Cloud ${{ inputs.flavor }} (${{ inputs.arch }})
     runs-on: "ubuntu-24.04"
+    if: ${{ inputs.flavor != 'skip' }}
```
NotTheEvilOne, line 44: I guess these if-conditions can be removed. See .github/workflows/tests.yml comment about empty matrices.

NotTheEvilOne (changes requested):

.github/workflows/tests.yml
yeoldegrove: @NotTheEvilOne No, tried it out...it gives you this error message when you pass the empty include:

```
Error when evaluating 'strategy' for job 'test_flavors_cloud'.
gardenlinux/gardenlinux/.github/workflows/tests.yml@aaeffbe3313fda358598d16fa2dc42b171e035c4
(Line: 258, Col: 15): matrix must define at least one vector
```

.github/workflows/test_flavor_cloud_ng.yml
yeoldegrove: See my comment about the skip matrix...I really tried it out and no empty string, no empty list, not empty `include` object lets you get past the `fromJson()` parsing.
And if it is successfully parsed, the `test_flavour_cloud_*.yml` is always started.

If you have other ideas, we can try those out. This is what worked with different combinations of test_types (see PR description and logs).

.github/workflows/tests.yml
yeoldegrove: Removed the verbose debugging.

.github/workflows/tests.yml
yeoldegrove: Removed the verbose debugging.

NotTheEvilOne (approved):
LGTM

PR #3683, opened 2025-10-20
Tests: Migrate `feature/metal` tests to new testing framework

### LINKED ISSUE #3610: Migrate `feature/metal` tests
- [x] metal/test/test_kernel_config.py::test_kernel_configs[test_kernel_config_tuple0] <- ../features/metal/test/test_kernel_config.py #3753 
- [x] metal/test/test_kernel_config.py::test_kernel_configs[test_kernel_config_tuple1] <- ../features/metal/test/test_kernel_config.py #3753 
- [x] metal/test/test_kernel_config.py::test_kernel_configs[test_kernel_config_tuple2] <- ../features/metal/test/test_kernel_config.py #3753 
- [x] metal/test/test_packages_musthave.py::test_packages_musthave <- helper/tests/packages_musthave.py ❌[Removed](https://github.com/gardenlinux/gardenlinux/blob/main/docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md)

ByteOtter:

Migrates the `test_kernel_configs` test to the new testing framework and introduces some additional fixtures in a new `kernel_configs.py` plugin.

Fixes #3610 

**Definition of Done:**

**Release note**:

```feature user

```

### DIFF
--- a/tests-ng/plugins/kernel_configs.py
+++ b/tests-ng/plugins/kernel_configs.py
@@ -8,7 +8,6 @@ import pytest
 def kernel_config_paths():
     """Get all available /boot/config-* files."""
     configs = sorted(glob.glob("/boot/config-*"))
-    assert configs, "No kernel config files found under /boot/config-*"
     return configs
 
 
--- a/tests-ng/test_kernel_config.py
+++ b/tests-ng/test_kernel_config.py
@@ -15,13 +15,16 @@ def test_kernel_config_values(
     """Ensure key kernel config options have expected values."""
 
     # Arrange
+    if not kernel_config_dict:
+        pytest.fail("No kernel config files found under /boot/config-*")
+
     relevant_configs = [
         (path, config)
         for path, config in kernel_config_dict.items()
         if arch == "all" or arch in path or arch == system_architecture
     ]
     if not relevant_configs:
-        pytest.skip(f"No kernel configs found for arch {arch}")
+        pytest.skip(f"No kernel configs found for arch '{arch}'")
 
     # Act / Assert
     for path, config in relevant_configs:


tests-ng/plugins/kernel_configs.py
```diff
@@ -0,0 +1,54 @@
+import glob
+import platform
+
+import pytest
+
+
+@pytest.fixture(scope="session")
```
fwilhe, line 7: what is this scope needed for?

tests-ng/plugins/kernel_configs.py
```diff
@@ -0,0 +1,54 @@
+import glob
+import platform
+
+import pytest
+
```
fwilhe, line 5: other plugins are typically implemented as a class, is there a good reason to do it differently here? not exactly sure if that class pattern is really the best, but I would argue that consistency is important here, unless there is a really good reason to do it differently

tests-ng/plugins/kernel_configs.py
```diff
@@ -0,0 +1,54 @@
...
+
+@pytest.fixture(scope="session")
+def kernel_config_paths():
+    """Get all available /boot/config-* files."""
+    configs = sorted(glob.glob("/boot/config-*"))
+    assert configs, "No kernel config files found under /boot/config-*"
```
fwilhe: we should not have asserts in plugins

returning an empty list and asserting in the test would be fine imo

btw we have this test in the old framework to assert that there is only one kernel https://github.com/gardenlinux/gardenlinux/blob/194499e69d622c252bed70c3b00fb7759bce4338/features/cloud/test/test_only_one_kernel.py#L5

tests-ng/test_kernel_config.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+
+
+@pytest.mark.parametrize(
```
fwilhe, line 4: I would argue that this is close to the 'bad' case here https://github.com/gardenlinux/gardenlinux/blob/main/tests-ng/DEVELOPER.md#parametrization

Not exactly sure how I would do it, but this test has way too much logic for my liking, and the parameter structure is quite complex.

tests-ng/plugins/kernel_configs.py
ByteOtter: Ah yes agreed. I'll go ahead and fix this.

tests-ng/plugins/kernel_configs.py
ByteOtter: I want to make sure these fixtures are only created once and last until all tests that want to access these files are done running. So we don't have to access these files over and over and over again.

tests-ng/plugins/kernel_configs.py
ByteOtter: I don't really see a reason to overengineer this into a class to be completely honest. But I can if you'd prefer.

tests-ng/test_kernel_config.py
ByteOtter: I see your point, but I have to disagree. While it's good that tests don't get too big, I don't think we are nearly there with this one.
I personally think it's clear that the test has one job, and I would prefer not dissecting everything into a dozen one-liner helper functions. Not that I'd know what to pull out here in the first place ... I guess I could pull the `relevant_configs` assembly out into a helper like this?

```python
def relevant_kernel_configs(kernel_config_dict, arch, system_architecture):
    return [
        (path, config)
        for path, config in kernel_config_dict.items()
        if arch == "all" or arch in path or arch == system_architecture
    ]
```

When it comes to the parameters, I just took what we already used so I am open to suggestions.

yeoldegrove (changes requested):
@ByteOtter I guess this PR is obsolete with the merged https://github.com/gardenlinux/gardenlinux/pull/3753 ?

PR #3692, opened 2025-10-20
Fix various regressions in test-ng

nkraetzschmar:
This solves the following issues:

- #3689
  **Note**: this is "fixed" by fully reverting https://github.com/gardenlinux/gardenlinux/commit/88aefda3b403de368828984968bd475250c816a3. This should only be a temporary fix until we properly re-work this
- Fixes #3690
- Fixes #3691
  - Disables the pytest cache plugin which tried to write to a read-only directory in qemu, we anyway don't need or use a cache
 - Implements the `security_id` mark to get rid of the warnings

  We had used `security_id` in tests without having the mark implemented in the new test framework. This PR ports the old implementation of the mark into the new framework, in the form of a plugin. This allows us to remove the warnings without commenting out the marks, and should be the way forward.
  
  Port of https://github.com/gardenlinux/gardenlinux/pull/2477
  Related to https://github.com/gardenlinux/gardenlinux/issues/3691
  Fixes https://github.com/gardenlinux/gardenlinux/issues/3707

PR #3694, opened 2025-10-20
Add basic ability to bundle test-ng as a container

### LINKED ISSUE #3693: Make test-ng work in place on live gardener nodes
Bundle test-ng in a container that can be run privileged on a gardener node and will then escape to the host namespaces and run the test suite on the host that way. The test framework dist itself should also be bundled in the container => This is a bit more tricky then just a regular `setns` call as we also need to carry over a detached bind mount fd (obtainable via the `open_tree` syscall of the new kernel mount API) in order to keep access to the contents of the original container once transitioned into the host `mnt_ns`.

nkraetzschmar:
This container uses `setns` to run on the host namespace. Therefore it requires privileged containers with `pid=host`.

This will resolve #3693

**~Still a draft - TODO:~**

- [x] Merge regression fixes https://github.com/gardenlinux/gardenlinux/pull/3692 (especially https://github.com/gardenlinux/gardenlinux/pull/3692/commits/4d86486c0b41a22ea0da5d30b0e64a24fa0ab3d8 is a hard requirement for running on gardener nodes!)
- [x] Add logic to build and **publish** this container into the GitHub actions
- [x] Do an e2e run on actual gardener nodes (so far only tested against privileged pid host containers locally)

### DIFF
--- a/tests-ng/README.md
+++ b/tests-ng/README.md
@@ -352,6 +352,30 @@ cd /run/gardenlinux-tests && sudo ./run_tests --system-booted --allow-system-mod
 - Very fast execution method
 - Limited to Base Image and an unbooted system
 
+### Gardener / Kubernetes Cluster Live Tests
+
+The test framework can be run directly on a live gardener cluster (or any Kubernetes cluster running gardenlinux nodes for that matter). For that simply deploy the following yml:
+
+```yml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-ng
+spec:
+  hostPID: true
+  restartPolicy: Never
+  containers:
+  - name: test-ng
+    image: ghcr.io/gardenlinux/test-ng:nightly
+    securityContext:
+      privileged: true
+```
+
+After this is deployed and the tests ran you can simply get the pod logs to see the test results. If you want to target a specific node to run the tests on you should also pin this pod to that node.
+
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


tests-ng/plugins/features.py
```diff
@@ -5,6 +5,8 @@
 
 booleanAlgebra = boolean.BooleanAlgebra()
 
+features = set([])
```
moritzkirchnersap: Isn't that the same as set() ? 🤔

.github/workflows/publish_oci_containers.yml
```diff
@@ -349,3 +349,103 @@ jobs:
             --container ${{ needs.determine_environment.outputs.repository }} \
             --version ${{ inputs.version }} \
             --manifest_folder manifests
+
```
yeoldegrove, line 352: @NotTheEvilOne Can `gl-oci` do these `podman` operations already?

tests-ng/plugins/features.py
nkraetzschmar: yep, updated to just use `set()`

.github/workflows/publish_oci_containers.yml
nkraetzschmar: I mostly just mirrored the way [.github/workflows/publish_kmodbuild_container.yml](https://github.com/gardenlinux/gardenlinux/blob/3077997cb5a081353ef0d3480f111729aa8aafb6/.github/workflows/publish_kmodbuild_container.yml) works. If `gl-oci` can do this, then we should probably do a separate PR after this to migrate both to it.

But unless this is a really quick change that we can be sure works, I'd prefer not stalling this PR till then.

PR #3699, opened 2025-10-21
feat: test-ng: support multiarch builds for tofu provider azurerm

yeoldegrove:

Support multiarch builds for tofu provider azurerm in test-ng.


saubhikdattagithub (approved):
yes, it now works in macOs, Thanks much @yeoldegrove

PR #3728, opened 2025-10-27
dmesg test

vivus-ignis:
**What this PR does / why we need it**: migration of https://github.com/gardenlinux/gardenlinux/blob/main/features/server/test/test_dmesg.py to the new test framework

Fixes https://github.com/gardenlinux/gardenlinux/issues/3727

### DIFF
--- a/features/server/pkg.include
+++ b/features/server/pkg.include
@@ -14,7 +14,6 @@ lsb-release
 net-tools
 netbase
 nvme-cli
-pkexec
 polkitd
 procps
 quota


tests-ng/test_dmesg.py
```diff
@@ -0,0 +1,20 @@
+import re
+from pathlib import Path
+
+import pytest
+
+
+@pytest.mark.feature("server and not metal", reason="needs server feature")
+def test_dmesg_sysctl_config():
+    config_file = "/etc/sysctl.d/40-allow-nonroot-dmesg.conf"
+    assert Path(config_file).exists()
```
moritzkirchnersap: according to our rules, this should be a different test, right?

Only one assert per test

While I'm not the biggest fan of that rule, I guess we should try to be consistent.

tests-ng/test_dmesg.py
```diff
@@ -0,0 +1,20 @@
...
+    assert re.search(
+        r"(?m)^\s*(?!#)kernel\.dmesg_restrict\s*=\s*0", Path(config_file).read_text()
```
moritzkirchnersap: I'm trying to come up with a better readable without puzzling together the RE solution, but so far I couldn't.

Should we start using configparser and use that in some sort of plugin? We might want to parse more config files than only one, but even then I couldn't come up with a pretty looking test for that. With allow_unnamed_section=True it can [parse strings](https://docs.python.org/3/library/configparser.html#unnamed-sections), but then throws exceptions if you try to .get() a value in a commented out string AFAI saw while trying around right now.


Maybe something like:

```
config_value = 1
config_file_content = Path(config_file).read_text()
for line in config_file_content.splitlines():
    if line.strip().startswith('#'):
        continue
    if 'kernel.dmesg_restrict' in line:
        config_option, config_value = line.strip().split('=')
assert config_value.strip() == '0', "Config option kernel.dmesg_restrict value wasn't set to 0"
```

But then that's merely a suggestion and a question: what do you think about the readability of it? I don't like it too much as there are too many .strip()'s, honestly, but maybe can be improved upon?

tests-ng/test_dmesg.py
```diff
@@ -0,0 +1,20 @@
...
+    )
+
+
+@pytest.mark.feature(
+    "server and booted and not metal", reason="sysctl needs a booted system"
```
yeoldegrove: - `booted` is not a feature, use https://github.com/gardenlinux/gardenlinux/blob/main/tests-ng/DEVELOPER.md#pytestmarkbootedreason.

tests-ng/test_dmesg.py
yeoldegrove: @moritzkirchnersap @vivus-ignis There is something in the making that parses (files) ... It is not ready yet, though. https://github.com/gardenlinux/gardenlinux/pull/3753/commits/91be0e779a77695970f0db33c5936d86b535b076

vivus-ignis: Converted regex to a multiline one with inline comments.

tests-ng/test_dmesg.py
vivus-ignis: Fixed.

tests-ng/test_dmesg.py
vivus-ignis: Fixed.

tests-ng/test_dmesg.py
moritzkirchnersap: Thanks, way better IMHO!

tests-ng/test_dmesg.py
```diff
@@ -0,0 +1,79 @@
...
+@pytest.mark.booted(reason="sysctl needs a booted system")
+def test_dmesg_sysctl_runtime(config_file, is_restricted, sysctl):
+    sysctl_toggle = "1" if is_restricted else "0"
+    assert sysctl["kernel.dmesg_restrict"] == sysctl_toggle
+
+
+@pytest.mark.booted(reason="needs a booted system with dmesg restrictions loaded")
+@pytest.mark.feature("gardener")
+def test_dmesg_call_by_unprivileged_user_works(shell):
+    assert shell("id -u", capture_output=True).stdout.strip() != "0"
```
yeoldegrove: As we don't add the `@pytest.mark.root()` marker here it is clear that this does not run as root and this assertion can be skipped (please double check it works, though).

tests-ng/test_dmesg.py
```diff
@@ -0,0 +1,79 @@
...
+
+    res = shell("dmesg", capture_output=False, ignore_exit_code=True)
+    assert res.returncode == 0
```
yeoldegrove, line 92: 👍

tests-ng/test_dmesg.py
```diff
@@ -0,0 +1,79 @@
...
+    },
+    "server and not gardener": {
+        "file_path": "/etc/sysctl.d/40-restric-dmesg.conf",
+        "restricted": True,
+    },
+    "stig": {"file_path": "/etc/sysctl.d/99-stig.conf", "restricted": True},
+}
+
+# server adds /etc/sysctl.d/40-restric-dmesg.conf, gardener excludes it
+#    and adds /etc/sysctl.d/40-allow-nonroot-dmesg.conf instead
```
yeoldegrove: While I get where you are going with this, e.g. not having multiple very similar tests and drying this out, I would vote to align with https://github.com/gardenlinux/gardenlinux/blob/5c964f41922e379f4e0fd06012f55463911b1968/tests-ng/DEVELOPER.md#parametrization and use clear (but multiple tests) that use `@pytest.mark.feature`. This way one does not have to understand the `CONFIG` map and its logic and can _just read_ the test.

tests-ng/test_dmesg.py
```diff
@@ -0,0 +1,79 @@
...
+    [
+        pytest.param(
+            value["file_path"], value["restricted"], marks=pytest.mark.feature(key)
+        )
+        for key, value in CONFIG.items()
+    ],
+    ids=list(CONFIG.keys()),
+)
+def test_dmesg_sysctl_restrictions_on_accessing_dmesg(config_file, is_restricted):
+    sysctl_toggle = 1 if is_restricted else 0
```
yeoldegrove: I do not 100% get this `is_restricted` variable. Maybe this goes away if we just have clear/multiple tests like outlined above?

yeoldegrove (changes requested):

tests-ng/test_dmesg.py
vivus-ignis: Fixed, unnecessary check removed.

tests-ng/test_dmesg.py
vivus-ignis: Improved naming for consistency.

tests-ng/test_dmesg.py
vivus-ignis: While in general I prefer plain tests by myself, in this case just by reading the CONFIG dict a developer can quickly spot the differences between features that configure dmesg in opposite ways. The alternative would be 3 screens of code and jumping back and forth to find the most important information.
I improved variables naming to have uniform `is_user_restricted` used for dict keys, parameters and variables.

yeoldegrove: @fwilhe What is your take on this?

fwilhe: > The alternative would be 3 screens of code and jumping back and forth to find the most important information.

Not sure if I follow. If I understand it correctly, Eike's suggestion is to turn this one parameterised test into three tests, does not look like each of them would be a screen of code.

I don't have strong feelings in this case, sound like an edge case where I would be willing to accept both solutions, but my gut feeling would be to prefer the three individual tests because reading the parameter and how it extracts information from the dictionary does seem non-trivial to me.

But this would not be something where I insist on a change to approve, I feel.

Does that help?

vivus-ignis: The code is rewritten without parametrization. @yeoldegrove please re-check.

PR #3739, opened 2025-10-28
shell history test

vivus-ignis:
migration of https://github.com/gardenlinux/gardenlinux/blob/main/features/server/test/test_history.py to the new test framework
Fixes https://github.com/gardenlinux/gardenlinux/issues/3729

### DIFF
--- a/tests-ng/test_dmesg.py
+++ b/tests-ng/test_dmesg.py
@@ -1,4 +1,3 @@
-import re
 from pathlib import Path
 
 import pytest


tests-ng/test_history.py
```diff
@@ -0,0 +1,58 @@
...
+@pytest.mark.feature("server", reason="needs server feature")
+def test_history_profile_d_contains_required_configuration():
+    config_contents = Path(CONFIG_FILE).read_text()
+
+    patterns = [
+        re.compile(
+            r"""
+            ^\s*    # any whitespace in the beginning of the line is allowed
+            (?!\#)  # fail the search if a comment character is found
+            HISTFILE=/dev/null
```
fwilhe: I would prefer to abstract those regexes into a utility function like `has_line(contents, 'HISTFILE=/dev/null')`, so the whole test would be more expressive

```python
assert has_line(contents, 'HISTFILE=/dev/null')
assert has_line(contents, 'readonly HISTFILE')
```

etc

fwilhe: not even sure if regex is needed here.. sure the line could have whitespaces between the words ans still be valid shell, but maybe this could be done in a more simple way, like splitting the input line on whitespace and checking that all the words appear in that line in the right order

vivus-ignis: Replaced regex with file_content plugin call.

tests-ng/test_history.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@pytest.mark.feature("server")
+def test_history_profile_d_file_exists():
+    assert Path(CONFIG_FILE).exists()
+
+
+@pytest.mark.feature("server")
+def test_history_profile_d_contains_required_configuration(file_content):
+    assert file_content.check_lines(
```
fwilhe, line 15: i was not aware we have this function already. looks good to me.

fwilhe (approved):

Issue #3436, opened 2025-09-11
Create FIPS 140-3 Garden Linux base image


pnpavlov:
# EIPC: Create FIPS 140-3 Garden Linux base image

## Overview

As first step of preparing for [FIPS 140-3](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.140-3.pdf) certification, Garden Linux should offer for non-productive usage an image based on the latest nightly release with FIPS enabled. This would enable preview and further validation by our adopters, before the formal application procedure starts.

## Objectives
- Build Garden Linux image for development and evaluation purposes with FIPS enabled.
- Enable FIPS configuration for the 4 packages in scope: Kernel Crypto API, OpenSSL, GnuTLS, Libgcrypt

## Acceptance Criteria
- [x] FIPS enabled the packages in scope
   - [x] Kernel Crypto API
   - [x] OpenSSL
   - [x] GnuTLS
   - [x] Libgcrypt

## Deliverables
- [ ] Garden Linux image with FIPS mode on that is able to join a Kubernetes cluster as Gardener Worker Node.

## User Stories
- [x] Create Garden Linus FIPS feature
- [x] Prepare Kernel Crypto API with FIPS on
- [x] Prepare OpenSSL with FIPS on
- [x] Prepare GnuTLS with FIPS on
- [x] Prepare Libgcrypt with FIPS on
- [x] Prepare new image with Gardener and FIPS feature on
- [x] Test starting the image standalone
- [x] Implement a test in the feature that validates if FIPS is enabled.
- [ ] Test using the image in Gardener context (worker node)

## Risks

Akendo:
I've checked out the different subsystems and think we can ship them in a single profile. There are some different subjects to discuss , but for the moment I want to document the current working state.

It's possible to set the kernel into the FIPS module by adding the `fips=1` to the boot parameter. Once started, the kernel will return the necessary settings:
```
sysctl -a | grep fips
crypto.fips_enabled = 1
crypto.fips_name = Linux Kernel Cryptographic API
crypto.fips_version = 6.12.47-cloud-arm64
```
Important is the status of the procfs:
```
cat /proc/sys/crypto/fips_enabled
1
```

[...]

PR #3744, opened 2025-10-28
Feature/enable FIPS flavors in AWS

Akendo:
In this PR, we add the FIPS flavors to AWS. 

Relates: #3436

### DIFF
--- a/features/ssh/test/test_sshd.py
+++ b/features/ssh/test/test_sshd.py
@@ -44,5 +44,5 @@ import pytest
     ]
 )
 
-def test_sshd(client, expected):
+def test_sshd(client, expected, non_fips):
         sshd(client, expected)
--- a/tests/conftests/features.py
+++ b/tests/conftests/features.py
@@ -12,16 +12,23 @@ def non_dev(testconfig):
         pytest.skip('test not supported on dev feature')
 
 @pytest.fixture
-def non_trustedboot(testconfig):
-    features = testconfig.get("features", [])
-    if "_trustedboot" in features:
-        pytest.skip('test not supported on trustedboot feature')
-
-@pytest.fixture
 def non_ephemeral(testconfig):
     features = testconfig.get("features", [])
     if "_ephemeral" in features:
         pytest.skip('test not supported on empemeral feature')
+
+
+@pytest.fixture
+def non_fips(testconfig):
+    features = testconfig.get("features", [])
+    if "_fips" in features:
+        pytest.skip('test not supported on fips feature')
+
+@pytest.fixture
+def non_trustedboot(testconfig):
+    features = testconfig.get("features", [])
+    if "_trustedboot" in features:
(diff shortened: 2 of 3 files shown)


flavors.yaml
```diff
@@ -32,6 +32,24 @@ targets:
...
+          - _prod
+        arch: amd64
+        build: true
+        test: true
+        test-platform: true
+        publish: true
+      - features:
+          - gardener
+          - _prod
+        arch: arm64
```
fwilhe, line 45: this change removes the non-fips version of the aws amd64 flavour, right? is this intentional?

Akendo: I think the format of the diff makes it seem like this? We add two separate entries here: [one](https://github.com/gardenlinux/gardenlinux/pull/3744/commits/8f21f1a771fd3d6ad08fbcd8c2161ca2042f34c6) and [two](https://github.com/gardenlinux/gardenlinux/pull/3744/commits/08c33e53682691486746230d03b9c606764388a3). The only change that follows is the sorting.

Akendo: When using the [split](https://github.com/gardenlinux/gardenlinux/pull/3744/files?diff=split#r2493765397) view of this PR, it might become more obviously.

yeoldegrove:
@Akendo As you enabled `test` and `test-platform`, please make sure that all tests pass and add some link to a log that proves this.
You can start a manual test on the cloud providers with e.g.:
```
gh workflow run .github/workflows/manual_tests.yml \
  --ref $your_branch \
  -f version=2046.0.0 \
  -f target=nightly \
  -f ignore_workflow_concurrency=true \
  -f test_types=chroot,qemu,cloud \
  -f build=true \ # or false if you want to test a built nightly version
  -f flavors_parse_params_test='--exclude "bare-*" --no-arch --json-by-arch --test --include-only "*fips*'
```

If we simply merge this without testing, the nightly will most likely be broken.

flavors.yaml
fwilhe: ok, seems right

fwilhe (approved):

Akendo:
> @Akendo As you enabled `test` and `test-platform`, please make sure that all tests pass and add some link to a log that proves this. You can start a manual test on the cloud providers with e.g.:
> 
> ```
> gh workflow run .github/workflows/manual_tests.yml \
>   --ref $your_branch \
>   -f version=2046.0.0 \
>   -f target=nightly \
>   -f ignore_workflow_concurrency=true \
>   -f test_types=chroot,qemu,cloud \
>   -f build=true \ # or false if you want to test a built nightly version
>   -f flavors_parse_params_test='--exclude "bare-*" --no-arch --json-by-arch --test --include-only "*fips*'
> ```
> 
> If we simply merge this without testing, the nightly will most likely be broken.

@yeoldegrove 
This does not work well. See [here](https://github.com/gardenlinux/gardenlinux/actions/runs/19331740769).

PR #3489, opened 2025-09-23
Migration of time_config to new testing-framework. refs #3428

commel:
**What this PR does / why we need it**:
Migrates https://github.com/gardenlinux/gardenlinux/blob/main/features/base/test/test_time_config.py to the new test-ng-Framework.

**Which issue(s) this PR fixes**:
Fixes #3428

**Definition of Done:**
- [x] The code is sufficiently documented
- [x] Tests run

**Special notes for your reviewer**:
To be able to migrate platform specific behavior this PR expands the `systemd` and `features` plugins by improving the handling of service files, being able to list installed units in addition to active ones and creating more guardrails for parsing. It also creates a new `Hypervisor` plugin for proper detection of the test suite's host hypervisor (if any) from two sources: `systemd-detect-virt` and DMI data read from `/sys/class/dmi/id/sys_vendor` and `/sys/class/dmi/id/product_name` respectively. Two sources were necessary as some QEMU based test environments for hypervisor images would falsely report running on Azure or AWS, etc leading to our tests attempting to assert the wrong thing (like being able to contact AWS's NTP server, while QEMU only provides the default clock)

PR #3750, opened 2025-10-29
sysctls tests

vivus-ignis:
**What this PR does / why we need it**: migration of https://github.com/gardenlinux/gardenlinux/blob/main/features/server/test/test_kernel_parameter.py to the new test framework

Fixes https://github.com/gardenlinux/gardenlinux/issues/3740

### DIFF
--- a/features/aide/exec.config
+++ b/features/aide/exec.config
@@ -1,21 +1,5 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-# needed by fedramp
-echo -e "\n/var/log/faillog Full" >> /etc/aide/aide.conf
-
-# Initials aide by creating the aide database based on the config under /etc/aide/aide.conf
-# and create a cronjob to check against the database every night. The output of the nightly
-# check will be written to /dev/shm/aide_check.
-
-if [ ! -f /var/lib/aide/aide.db ]; then
-    mkdir -p /var/lib/aide
-    aide --init --config /etc/aide/aide.conf
-    cp /var/lib/aide/aide.db{.new,}
-fi
-
-if [ ! -f /var/spool/cron/crontabs/root ] || ! grep "aide --check" /var/spool/cron/crontabs/root; then
-    cat >> /var/spool/cron/crontabs/root << EOF
-0 10 * * * root /usr/bin/aide --check --config /etc/aide/aide.conf > /dev/shm/aide_check 2>&1
-EOF
-fi
+systemctl enable aide-init.service
+systemctl enable aide-check.timer
--- a/features/aide/file.include/etc/systemd/system/aide-check.service
+++ b/features/aide/file.include/etc/systemd/system/aide-check.service
@@ -1,0 +1,7 @@
+[Unit]
+Description=Run AIDE integrity check
+
+[Service]
+Type=oneshot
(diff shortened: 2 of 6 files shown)


tests-ng/test_kernel_parameters.py
```diff
@@ -0,0 +1,19 @@
...
+@pytest.mark.feature("cloud or openstackbaremetal")
+@pytest.mark.booted(reason="sysctl needs a booted system")
+def test_kernel_parameters_cannot_symlink_what_you_do_not_own(sysctl):
+    assert sysctl["fs.protected_symlinks"] == "1"
+
+
+@pytest.mark.feature("cloud or openstackbaremetal")
+@pytest.mark.booted(reason="sysctl needs a booted system")
+def test_kernel_parameters_randomize_memory_allocation(sysctl):
+    assert sysctl["kernel.randomize_va_space"] == "2"
```
fwilhe, line 19: how about creating a local variable here to explain the meaning of the 'magic value' 2?

https://sysctl-explorer.net/kernel/randomize_va_space/


```python
assert sysctl["kernel.randomize_va_space"] == ENABLED_WITH_HEAP_RANDOMIZATION
```

would read much better imo

fwilhe (approved):
lgtm except for my comment on 'magic numbers'. please add a comment or use a local variable for documentation purposes, then it's good to merge imo

yeoldegrove:
I won't block this, just wanted to raise the discussion again how/when to write tests for different test environments as I had a similar discussion in https://github.com/gardenlinux/gardenlinux/pull/3752 with @Akendo and currently everybody does this a bit different (me included).

I already extended [Core Principle 5](https://github.com/gardenlinux/gardenlinux/blob/682d10c45602a0d5a483fd48d53d67613f5df064/tests-ng/DEVELOPER.md#5-target-appropriate-test-environments-and-platforms) a bit last week. Applying this would mean for me:

If we check sysctl parameters one should test for different test environments like this.

- chroot tests: check file `/etc/sysctl.d/$foo`
  - Not sure if this adds any value, except more test coverage and faster feedback, we need to align on this.
- qemu tests: check `sysctl $foo.$bar`
  - IMHO this is the test we want (like in this PR) as it really checks if what we wrote into a file is applied.

Other tests like kernel cmdline or adding modules to initrd should be tested with the same paradigm.

There might even scenarios where all 3 (chroot/qemu/cloud) tests have the possibility to be tested differently, e.g. @commel and @ByteOtter efforts in https://github.com/gardenlinux/gardenlinux/pull/3489.

tests-ng/test_kernel_parameters.py
yeoldegrove: Not sure if this information is something that should be in the implementation of the feature and not the test :shrug:

yeoldegrove: https://github.com/gardenlinux/gardenlinux/blob/682d10c45602a0d5a483fd48d53d67613f5df064/features/cloud/file.include/etc/sysctl.d/20-cloud.conf#L2 it is in fact stated in the feature, so I would not care.
Guess we have to find some common understanding how much explanation is needed in a test. Adding this as a docstring comment would also work.

Issue #3701, opened 2025-10-22
Develop some tests for FIPS


Akendo:
In this task, we develop some of the tests necessary to test FIPS.

We should have a set of tteststo ensure that all configuration files are present.
This includes the runtime files. Also that the HMAC file is present, the OpenSSL configuration.

Akendo:
PR was merged.

PR #3752, opened 2025-10-29
Feature/3701 tests for fips

Akendo:
In this PR, we add some basic tests for GnuTLS, libgcrypt, and the kernel regarding the FIPS mode. It validates only that the necessary prerequisites are present to enable the FIPS mode. 

Relates: #3701, #3436

### DIFF
--- a/docs/01_developers/bare_container.md
+++ b/docs/01_developers/bare_container.md
@@ -3,10 +3,10 @@
 ## Overview
 Garden Linux offers a range of specialized bare container images, each tailored for specific applications and designed for minimalism and security:
 
-- **Bare-libc**: Ideal for C/C++ applications requiring only essential C runtime libraries.
-- **Bare-python**: Equipped with Python runtime, perfect for Python-based applications.
-- **Bare-sapmachine**: Customized for sapmachine with necessary libraries and binaries.
-- **Bare-nodejs**: Includes Node.js environment, suitable for server-side JavaScript applications.
+- **bare-libc**: Ideal for C/C++ applications requiring only essential C runtime libraries.
+- **bare-python**: Equipped with Python runtime, perfect for Python-based applications.
+- **bare-sapmachine**: Customized for sapmachine with necessary libraries and binaries.
+- **bare-nodejs**: Includes Node.js environment, suitable for server-side JavaScript applications.
 
 These containers are optimized as base images for Dockerfiles. 
 Their development is streamlined by the [unbase_oci](https://github.com/gardenlinux/unbase_oci) tool, which efficiently minimizes image size by removing unnecessary components from base container images.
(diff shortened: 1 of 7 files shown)


Akendo:
To enable FIPS on Garden Linux, we have to perform different tasks. The first task is to set the kernel into the FIPS mode. This will be done by setting the `fips=1` flag. The test ensures that we have the necessary configuration in place. It validates that the system was booted with the runtime indicator presented. 

For GnuTLS and libgcrypt both require an additional setting besides the kernel setting, and that's an empty file in the respective configuration files. Once these files are placed, both libraries will be initialized in FIPS mode. The test ensures that these files are presented.

tests-ng/test_fips.py
```diff
@@ -0,0 +1,87 @@
...
+
+@pytest.mark.feature("_fips")
+def test_kernel_cmdline_fips_file_content():
+    """
+    We have to ensure that the fips=1 was set.
+    """
+    with open("/etc/kernel/cmdline.d/30-fips.cfg") as kernel_cmd_file:
+        assert (
+            kernel_cmd_file.read() == 'CMDLINE_LINUX="$CMDLINE_LINUX fips=1"\n'
+        ), "fips=1 wasn't set in the kernel cmdline"
```
yeoldegrove, line 76: I asked myself the question if just checking content of the very same config files that we put into the image via a feature is a good test.
https://github.com/gardenlinux/gardenlinux/blob/f1850587cbe47f910d0e4061884bb06f6121a4a0/tests-ng/plugins/kernel_cmdline.py adds some real checking of the booted kernel's cmdline which is a better test.

Akendo: I will include this test as well, but I prefer the file check as well, since this way I see something goes wrong within the feature. The main aim at this stage was to have file checks.

Akendo: But Thanks!

Akendo: Also, the test would become a runtime test afterward. But since I validate that the kernel has set the `/proc/sys/crypto/fips_enabled` was set it would be the same. Your code can only be executed with a fully working system and reads a different file from procfs.

yeoldegrove: The 5th [Core Principle](https://github.com/gardenlinux/gardenlinux/blob/9ca161dcc8cbd7fe3f7291bc15ac95a990b593c8/tests-ng/DEVELOPER.md#core-principles) tells us to write tests that can be run across all supported environments and platforms, which includes chroot but also qemu/cloud envs.
Maybe extending this or even adding another core principle to add "runtime tests" (in lack of a better word) or tests that actually check that something that is configured in a config file really has the expected outcome (maybe "integration test"?).

I'll guess we'll have this discussion again at the point where we review all features again and talk about test coverage.
I won't block you here, this was just a comment.

Akendo: That's the point, but since I was blocked by #3796 this is the first set of the tests. I'm not done, but I won't create all test at once.

Akendo: And I appreciate the comments. I should have added that this is just the first set of changes.

PR #3756, opened 2025-10-29
checking sudoers for env_reset

vivus-ignis:
**What this PR does / why we need it**: migration of https://github.com/gardenlinux/gardenlinux/blob/main/features/server/test/test_reset_env_vars.py to the new test framework

Fixes https://github.com/gardenlinux/gardenlinux/issues/3755


moritzkirchnersap (approved):
LGTM

(in the previous version I would have asked to separate the check if the file exists but this version looks good to me)

PR #3787, opened 2025-11-03
Tests: Migrate test_sysrq to test-ng

### LINKED ISSUE #3782: Migrate base/test/test_disable_sysrq.py

commel:
Migrate test https://github.com/gardenlinux/gardenlinux/blob/main/features/base/test/test_disable_sysrq.py to new testing-framework.

Fixes #3782

### DIFF
--- a/tests-ng/test_sysrq.py
+++ b/tests-ng/test_sysrq.py
@@ -1,0 +1,26 @@
+import pytest
+from plugins.sysctl import Sysctl
+
+
+@pytest.mark.booted("sysrq is only relevant in a booted environment")
+def test_disable_sysrq(sysctl: Sysctl):
+    """
+    With the Magic SysRq key a user can debug the kernel and modify the system. This
+    should not be possible in a production environment. Therefore kernel.sysrq should be
+    disabled and not be found in the /proc/sys directory.
+
+    See more: https://en.wikipedia.org/wiki/Magic_SysRq_key
+    """
+    with pytest.raises(KeyError):
+        sysctl["kernel.sysrq"]
+
+
+def test_persistent_disable_sysrq():
+    """
+    To ensure that the Magic Sysrq key is persistenly disabled, it should be turn off
+    in the configuration file.
+
+    See more: https://en.wikipedia.org/wiki/Magic_SysRq_key
+    """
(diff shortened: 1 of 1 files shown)


tests-ng/test_sysrq.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+from plugins.sysctl import Sysctl
+
+
+@pytest.mark.booted("sysrq is only relevant in a booted environment")
+def test_disable_sysrq(sysctl: Sysctl):
+    """kernel.sysrq should be disabled and therefore not be found in the /proc/sys directory"""
```
fwilhe: this is to avoid users from interfering with the system via keypresses, right? maybe the reason for this assertion should be in the function name/docstring

commel: The former test did not explain anything at all. But I guess I can add a description that describes what this might intend.

tests-ng/test_sysrq.py
```diff
@@ -0,0 +1,26 @@
...
+@pytest.mark.booted("sysrq is only relevant in a booted environment")
+def test_disable_sysrq(sysctl: Sysctl):
+    """
+    With the Magic SysRq key a user can debug the kernel and modify the system. This
+    should not be possible in a production environment. Therefore kernel.sysrq should be
+    disabled and not be found in the /proc/sys directory.
+
+    See more: https://en.wikipedia.org/wiki/Magic_SysRq_key
+    """
+    with pytest.raises(KeyError):
```
Akendo, line 14: I feel a bit dumb here. How is this test working? Should we not also check for the state of the procfs?

yeoldegrove (changes requested):
@commel sorry for the parallel work, I already take care of this test in https://github.com/gardenlinux/gardenlinux/pull/3753/files#diff-a3308468cef8d7ea6152a170cd164e5d075404aa275135774683f38cb2adb940 (including extending `sysctl` plugin and a new `kernel_configs` plugin).

Are you good with closing this PR?

tests-ng/test_sysrq.py
commel: The Plugin iterates over the files in `/proc/sys` and fills a dict. If the key is missing the entry in `/proc/sys` is also missing. As the test checks for absence if expect the KeyError to be raised here.

PR #3788, opened 2025-11-03
feat: test-ng: migrate selinux tests

### LINKED ISSUE #3590: Migrate `feature/_selinux` tests
- [ ] _selinux/test/test_kernel_cmdline.py::test_kernel_cmdline[/etc/kernel/cmdline-security=selinux] <- ../features/_selinux/test/test_kernel_cmdline.py
- [x] _selinux/test/test_packages_musthave.py::test_packages_musthave <- helper/tests/packages_musthave.py ❌[Removed](https://github.com/gardenlinux/gardenlinux/blob/main/docs/architecture/decisions/0013-discontinue-packages-musthave-tests.md)

nkraetzschmar:

Migrate the selinux tests to the new test framework. Instead of plainly migrating the old test that just did:

```python
@pytest.mark.parametrize(
    "file,args",
    [
        ("/etc/kernel/cmdline", "security=selinux"),
    ]
)


def test_kernel_cmdline(client, file, args):
    file_content(client, file, args, only_line_match=True)
```

So was looking at the contents of `/etc/kernel/cmdline` this migrated version instead actually reads `/proc/cmdline` from within the running system and additionally also checks for the presence of `/sys/fs/selinux/enforce` to double validate that selinux is actually on and not e.g. overwritten by some other `lsm=` cmdline arg.

> [!NOTE]
> It is purposefully only checking for existence of `/sys/fs/selinux/enforce`, **not** reading it, since (at least for now) we only care about selinux being on, we don't care if we are running in enforcing or permissive mode.

Resolves #3590

### DIFF
--- a/tests-ng/plugins/kernel_cmdline.py
+++ b/tests-ng/plugins/kernel_cmdline.py
@@ -1,0 +1,13 @@
+from typing import List
+
+import pytest
+
+from .booted import is_system_booted
+
+
+@pytest.fixture
+def kernel_cmdline() -> List[str]:
+    if not is_system_booted():
+        pytest.skip("can't get kernel cmdline when not running on a booted system")
+    with open("/proc/cmdline", "r") as f:
+        return f.read().strip().split()
--- a/tests-ng/test_selinux.py
+++ b/tests-ng/test_selinux.py
@@ -1,0 +1,16 @@
+import os
+from typing import List
+
+import pytest
+
+
+@pytest.mark.feature("_selinux")
+@pytest.mark.booted(reason="requires a running kernel to read /proc/cmdline")
+def test_selinux_cmdline(kernel_cmdline: List[str]):
+    assert "security=selinux" in kernel_cmdline
+
+
+@pytest.mark.feature("_selinux")
+@pytest.mark.booted(reason="requires a running kernel to access sysfs")
+def test_selinux_enabled():
+    assert os.path.exists("/sys/fs/selinux/enforce"), "SELinux not enabled"


tests-ng/plugins/kernel_cmdline.py
```diff
@@ -0,0 +1,13 @@
+from typing import List
+
+import pytest
+
+from .booted import is_system_booted
+
```
yeoldegrove, line 6: Some plugins already define `@dataclass`es. Maybe it is worth looking into that.

nkraetzschmar:
> Looks like the tests actually fail for a few features

yep, that's because the `_pxe` based artifacts are currently starting with a fixed cmdline, so the test is correctly failing due to an invalid cmdline

#3801 will fix that, so we should probably wait for that to be merged before we merge in this

tests-ng/plugins/kernel_cmdline.py
nkraetzschmar: Not sure I quite follow.... You mean we should return some dataclass instead of a plain `List[str]`? I can't quite see how that would provide real benefit here since we are not returning some struct like data. Or am I missing something?

yeoldegrove: You're right, the list we return here is fine. Just had that in mind as I was working on more complex plugins over the last few days.

yeoldegrove (changes requested):
@nkraetzschmar As https://github.com/gardenlinux/gardenlinux/pull/3801 is now merged, if you rebase this on main and it builds, you'll get my ✔️ .

PR #3797, opened 2025-11-04
Set OPENSSL_MODULES to load fips.so

Akendo:
While developing a test for GnuTLS to ensure that we have the correct HMAC computed, I noticed a 'bug' in OpenSSL/Python/Testing Framework. It's an oddity. The problem is that the Python version can't load the `fips.so`, by setting this as an environment variable, it resolves the issue.


yeoldegrove (approved):
As long as adding env vars like this is an exception, I am fine with it. If those pile up, we should think about another solution.

Akendo:
> As long as adding env vars like this is an exception, I am fine with it. If those pile up, we should think about another solution.

You're correct. Technically, this should be fixed upstream. However, there were already some PRs closed that tried to address FIPS.

PR #3798, opened 2025-11-04
Update test-ng Python env to 3.14

Akendo:
Updating to the latest version of Python for test-ng.


fwilhe:
> @Akendo What is the reason for this?

@yeoldegrove is there a reason not to? In general I think it is good to keep up to date, maybe one would want to wait for some patch releases, but in general I would want to automate keeping python updated in a dependabot-like fashion if possible.

CI failures seem unrelated (sysdiff), so i would assume we could merge this without any disruption?

PR #3799, opened 2025-11-04
Tests: Migrate duplicate_uids to test-ng

### LINKED ISSUE #3784: Migrate base/test/test_duplicate_uids.py

commel:
Migrates https://github.com/gardenlinux/gardenlinux/blob/main/features/base/test/test_duplicate_uids.py to new test-ng.

Fixes #3784

I've added a negative test.

### DIFF
--- a/tests-ng/test_duplicate_uids.py
+++ b/tests-ng/test_duplicate_uids.py
@@ -1,0 +1,25 @@
+from typing import Dict, List
+
+from plugins.password_shadow import passwd_entries
+from plugins.utils import check_for_duplicates
+
+
+def test_duplicate_uids(passwd_entries: List[Dict[str, str | List[str]]]):
+    """
+    Check if any duplicate userids are defined in /etc/passwd.
+    """
+    duplicates = check_for_duplicates(entries=passwd_entries, field="uid")
+    assert (
+        len(duplicates) == 0
+    ), f"uids {duplicates} are used multiple times in /etc/passwd"
+
+
+def test_duplicate_uids_detection(passwd_entries: List[Dict[str, str | List[str]]]):
+    """
+    Appending an existing entry to the list to check if the duplication check finds this.
+    """
+    passwd_entries.append(passwd_entries[0])
+
+    duplicates = check_for_duplicates(entries=passwd_entries, field="uid")
+    assert len(duplicates) == 1, f"duplicate uid is not found successfully"
+    assert duplicates[0] == passwd_entries[0]["uid"]
--- a/tests-ng/plugins/utils.py
+++ b/tests-ng/plugins/utils.py
@@ -77,3 +77,18 @@ def parse_etc_file(
         entries.append(entry)
 
     return entries
+
+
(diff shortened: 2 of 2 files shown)


tests-ng/plugins/getent.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class GetEnt:
+
+    def __init__(self, shell: ShellRunner):
+        self._shell = shell
+
+    def passwd(self) -> list[PasswdEntry]:
+        result = self._shell(
+            cmd="getent passwd", capture_output=True, ignore_exit_code=True
```
yeoldegrove: As we do not currently have another source of users as local ones, maybe we can just name this class `Passwd` and parse the file? Otherwise I really like the approach.

commel: Sure, I will join these and modify it to return a dataclass.

yeoldegrove (approved):
Adding the `_check_for_duplicated` is fine here as it is nothing that makes sense in a plugin.
In the long run we should return better structured data from a User dataclass that returns the values we need so we don't need such helper functions anymore.

tests-ng/test_duplicate_uids.py
```diff
@@ -0,0 +1,38 @@
...
+    Appending an existing entry to the list to check if the duplication check finds this.
+    """
+    passwd_entries.append(passwd_entries[0])
+
+    duplicates = _check_for_duplicates(passwd_entries=passwd_entries)
+    assert len(duplicates) == 1, f"duplicate uid is not found successfully"
+
+
+def _check_for_duplicates(
+    passwd_entries: List[Dict[str, str | List[str]]],
```
fwilhe: does this need to be a union type?

tests-ng/test_duplicate_uids.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    Appending an existing entry to the list to check if the duplication check finds this.
+    """
+    passwd_entries.append(passwd_entries[0])
+
+    duplicates = _check_for_duplicates(passwd_entries=passwd_entries)
+    assert len(duplicates) == 1, f"duplicate uid is not found successfully"
+
+
+def _check_for_duplicates(
```
fwilhe: typically we would want to move utility functions to the plugins/utils.py file.. I'm wondering if this should not be written more generically, then it could also be used in general cases where 'list does not contain duplicates' is a requirement

yeoldegrove: :+1: for moving it to utils.py
@commel will work on https://github.com/gardenlinux/gardenlinux/issues/3814 next which might remove the need for this function

tests-ng/test_duplicate_uids.py
commel: At the moment yes, but this will be changed with #3814.

tests-ng/test_duplicate_uids.py
commel: Moved to utils and made the function more general purpose.

PR #3808, opened 2025-11-05
Test: Migrate test_groups.py to new testing framework.

### LINKED ISSUE #3786: Migrate base/test/test_groups.py

commel:
Migrate test_groups.py to new testing framework.

Fixes #3786

### DIFF
--- a/tests-ng/test_groups.py
+++ b/tests-ng/test_groups.py
@@ -43,7 +43,7 @@ def test_group_root_has_no_users(group_entries: List[Group]):
     ), f"user group root is not empty as expected. Was: {group.user_list}"
 
 
-@pytest.mark.feature("not _dev and not pythonDev")
+@pytest.mark.feature("not _dev and not pythonDev and not container")
 def test_group_wheel_has_no_users(group_entries: List[Group]):
     """
     Check if parameterized groups have the expected users.


tests-ng/test_groups.py
```diff
@@ -0,0 +1,41 @@
...
+        actual_user_list = grp["user_list"].strip().split(",")
+    else:
+        raise ValueError(f"user_list is of unhandled type")
+
+    # python split on an empty array will return [''] instead of []
+    actual_user_list = [a for a in actual_user_list if a.strip() != ""]
+
+    assert (
+        actual_user_list == user_list
+    ), f"assigned users from group {grp["group_name"]} dont match expected {user_list}. Was: {actual_user_list}"
```
yeoldegrove: I know it's duplicated code, but can we split this into another test?

tests-ng/test_groups.py
```diff
@@ -0,0 +1,41 @@
...
+    assert (
+        len(matching_groups) == 1
+    ), f"No exact match found for group {group_name}. Found {len(matching_groups)}, expected only 1"
+
+    # we have only group available
+    grp = matching_groups[0]
+
+    if isinstance(grp["user_list"], list):
+        # list, should not happen here, as group_entries will split by : by default and not comma as expected
+        actual_user_list = grp["user_list"]
```
yeoldegrove: If this should not happen, why have an if condition for it?

yeoldegrove (changes requested):

yeoldegrove:
@commel Seems as if you have to exclude the `container` feature from some tests (and maybe think about if those need extra tests).

tests-ng/test_groups.py
commel: Because the datatype suggests that a `str` or a `List[str]` can be returned, so I am covering the possible scenarios. Atm user_id splitting is not implemented, but when that changes the test would not fail automatically.

tests-ng/test_groups.py
commel: So one test to check if the groups `root` and `wheel` are there and another one to test if the lists are empty?

tests-ng/test_groups.py
nkraetzschmar: I think the `List[str]` return type is only for the `"rest"` key (not sure if having the return of `parse_etc_file` this way is that great of a design decision - but that's a different discussion).

So I'd say a simple assert that `grp["user_list"]` is a string would be better here.

tests-ng/test_groups.py
```diff
@@ -0,0 +1,41 @@
...
+    Check if parameterized groups have the expected users.
+    """
+    # find the parameterized group in the dataset
+    matching_groups = [
+        group_entry
+        for group_entry in group_entries
+        if group_entry["group_name"] == group_name
+    ]
+    assert (
+        len(matching_groups) == 1
```
nkraetzschmar: Instead of checking that the group we care about is mentioned once and only once, it might be better to split this into two tests: one that first checks that the `"group_name"` of all entries is unique (completely unparameterized, just looking at the entire set).

Then a second one that simply gets the first group of this name without any care about potential duplicates and asserts the extra conditions on the parameterized groups.

At least to me that would make the tests more readable, by having mostly just assertions in the test and not conditionals / multiple exit points within the same test.

tests-ng/test_groups.py
commel: We discussed internally to return more readable dataclasses, but that should happen in a different MR. Will change the if to an assert.

tests-ng/test_groups.py
yeoldegrove: The later one. IMHO parameterize is fine for group_name,user_list.

tests-ng/test_groups.py
```diff
@@ -0,0 +1,67 @@
...
+    duplicate_group = group_entries[0]
+    group_entries.append(duplicate_group)
+
+    duplicates = check_for_duplicates(group_entries)
+    assert (
+        len(duplicates) == 1
+    ), f"Group #{duplicate_group.groupname} should be found as duplicate, but was not."
+
+
+@pytest.mark.parametrize("group, user_list", [("root", [])], indirect=["group"])
```
fwilhe: 'indirect' is new to me https://docs.pytest.org/en/stable/example/parametrize.html#apply-indirect-on-particular-arguments

to be honest I'm not sure if this suits our goal of making tests easy to read. I recognise it is a clever concept, but the tests should be easy to read and not clever.

commel:
> I don't want to block a merge of this, but I feel a bit uneasy with the usage of advanced pytest constructs here. I think this goes a bit against the goal of making tests easy to read. @yeoldegrove @commel did you discuss this already?

It has been used in the tests already a couple of times. Imho makes the test more readable by removing the call of the configuration method within the tests itself.

fwilhe (approved):
LGTM, thanks for simplifying it @commel

PR #3816, opened 2025-11-06
docs: document decision to enforce single platform by default in Garden Linux Builder

fwilhe:
Documenting the decision we agreed on in a meeting today.

### DIFF
--- a/docs/architecture/decisions/0020-enforce-single-platform-by-default-in-builder.md
+++ b/docs/architecture/decisions/0020-enforce-single-platform-by-default-in-builder.md
@@ -27,3 +27,6 @@ Garden Linux builds are defined by a set of *features*, with `platform` being a 
 - This is a user-facing change and may cause issues for known and unknown users of the build infrastructure, requiring communication and migration support.
 - Consumers relying on multiple platforms or `cnames` may be affected and need to update their workflows.
 - Documentation and tooling must be updated to reflect the new enforcement and flag usage.
+- A new `platform-variant` flag will be introduced in the manifests to support platforms that require different deployment behaviors based on their features.
+- OpenStack images will be the first to utilize the `platform-variant` flag.
+- GLCI will need to be updated to support and use the new flag.


docs/architecture/decisions/0020-enforce-single-platform-by-default-in-builder.md
```diff
@@ -0,0 +1,30 @@
...
+1. The builder will be changed so that, by default, it enforces the presence of exactly one platform feature in a build.
+
+2. If zero or multiple platforms are specified, the build will fail.
+
+3. The previous behavior (allowing multiple platforms) will remain available via an explicit opt-in flag.
+
+## Consequences
+
+- Multiple features in the `gardenlinux/gardenlinux` repo need to be changed.
+- The builder needs to be changed (for example the `bootstrap` phase relies on the current behavior).
```
nkraetzschmar: That's not quite accurate. The bootstrap in the builder works fine. The dedicated GitHub actions bootstrap stage relies on the assumption to be able to build with zero platforms (to build a simple `base` only target). The bootstrap path that's implicitly called by the builder should continue to work just fine.

docs/architecture/decisions/0020-enforce-single-platform-by-default-in-builder.md
```diff
@@ -0,0 +1,30 @@
...
+- Builds with zero or multiple platforms will fail unless the opt-in flag is used.
+- The change will affect the "Common Name" (cname) of many image builds, as features like 'metal' will no longer be considered platforms. This impacts both command line arguments for the builder and output artifact filenames.
+- This is a user-facing change and may cause issues for known and unknown users of the build infrastructure, requiring communication and migration support.
+- Consumers relying on multiple platforms or cnames may be affected and need to update their workflows.
+- The builder will support platform variants, requiring additional branching logic.
```
nkraetzschmar: This isn't really a builder thing (at least not the way I understood this). The variant logic can trivially be implemented in a feature by just writing fields into `/etc/os-release`.

docs/architecture/decisions/0020-enforce-single-platform-by-default-in-builder.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Decision
+
+1. The builder will be changed so that, by default, it enforces the presence of exactly one platform feature in a build.
+
+2. If zero or multiple platforms are specified, the build will fail.
+
+3. The previous behavior (allowing multiple platforms) will remain available via an explicit opt-in flag.
+
+## Consequences
```
Gchbg, line 21: I would add something like this, feel free to reword:
`A new flag will be added to the manifests (`platform-variant`) for platforms that need to be deployed differently based on their features.`
`The OpenStack images will be the first to make use of this flag.`
`GLCI needs to be adapted to make use of the flag.`

Gchbg (changes requested):

docs/architecture/decisions/0020-enforce-single-platform-by-default-in-builder.md
fwilhe: addressed

Issue #3781, opened 2025-11-03
Enable Lima feature for releases


pnpavlov:
Currently, the lima images are [built in a separate pipeline](https://github.com/gardenlinux/gardenlinux/blob/main/.github/workflows/lima_upload_oidc.yml) and have preview/proof of concept status

To make them a proper part of the Garden Linux product, the images should be built and published like any other images. To do this, we need to add them to the [flavors.yaml](https://github.com/gardenlinux/gardenlinux/blob/main/flavors.yaml), not sure what else would need to be done, the images will then be published by glci like any other Garden Linux image.

Target picture:

Automated publishing of disk images (qcow2) and yaml manifests for each release and the nightlies.
To be defined and mapped to a images.gardenlinux.org domain.

User visible artefacts: Under (to be decided: the exact subdomain) images.gardenlinux.org, we expose the following manifests:

- `gardenlinux.yaml` to point to latest stable release always (same logic as for container images, always the latest minor of our latest major version). This is a moving target, so having sha checksums in here would not be helpful as users would have to prune the cache manually.
- `gardenlinux-nightly.yaml` pointing to latest nightly build that was completed. This is a moving target, so having sha checksums in here would not be helpful as users would have to prune the cache manually. This should overwrite the previous yaml manifest/disk image daily, there is no specific need to keep each nightly release in the lima variant.
- `gardenlinux-2070.0.0.yaml` pointing to that specific version. Those images never change so the yaml manifest needs to include the sha sum of the disk images. Each minor version gets its own yaml manifest and disk images.

For each manifest, there is an amd64 and an arm64 qcow disk image at the same location. For me (@fwilhe), it does not matter if this is hosted on gcp or on s3, all that matters is that we get a nice and stable url shema similar to https://images.gardenlinux.io/gardenlinux.yaml

Decision: no gardenlinux-nightly-number.yaml to not blow up the number of artefacts produced. (comment @fwilhe: I am not sure if our publishing pipeline supports this mode)

Consequence: This will increase the number of images on s3 and github releases.

SujanaSubr:
https://github.com/gardenlinux/gardenlinux/pull/3868

SujanaSubr:
Added a script to generate lima manifest which can be piped to limactl command

PR #3818, opened 2025-11-06
Add Lima image build

SujanaSubr:
Adds Lima image in flavors.yml to build with nightly and for release
Fixes #

**Definition of Done:**

**Release note**:

```feature user

```


SujanaSubr:
> @SujanaSubr Nice, the tests actually pass already for tests-ng. The old test suite also only needs a few adaptions or skips (from what I can see from the log, not looked into details, yet). I would vote for letting the tests enabled and fix them.

Yes , checking that already

fwilhe:
Awesome, thanks for working on this @SujanaSubr 

> Nice, the tests actually pass already for tests-ng

Whoop whoop 🎉 

> I would vote for letting the tests enabled and fix them.

Makes sense as long as the effort is reasonable. Let us know if there are blockers @SujanaSubr

flavors.yaml
```diff
@@ -661,13 +661,17 @@ targets:
   - name: lima
     category: hypervisor
     flavors:
-      - features: []
+      - features:
+          - gardener
+          - _prod
```
fwilhe: please don't do that, the lima feature is self sufficient and should neither include gardener nor prod

features/lima/pkg.exclude
```diff
@@ -0,0 +1 @@
+python3-cffi-backend
```
fwilhe: why is this needed?

flavors.yaml
SujanaSubr: ok i will remove it. _prod should still be there right?

features/lima/pkg.exclude
SujanaSubr: It is an orphaned package , test fails becoz of this . i do not see any related package which depends on it. So removed it to see if this will be excluded. Also It is just a trial to see what happens. I think not ready for review yet. i created PR to see the test case pass/fail

fwilhe: got it, thanks

flavors.yaml
fwilhe: No, I never tested it with prod and I don't think there is value in that. As mentioned, the lima feature is supposed to be self sufficient, it can be combined with features such as python or sapmachine, but it should not be combined with prod or similar features, they might very well break it.

yeoldegrove:
> > @SujanaSubr I am not 100% sure yet which purpose the published lima images shall fulfill... Is this documented somewhere already? If the purpose is somewhat gardener related, could you also try to enable the `gardener` and `_prod` features for the `lima` flavors?
> 
> No, this is not Gardener related. We only want/need the `lima` feature in those. This is the related issue: #3781
> 
Thanks for clarifying that.

> Essentially, it was @pnpavlov 's request to treat lima as a proper feature of Garden Linux, and publish images for all releases. For that I think we need to get rid of the special workflow we have for this image and add it to flavours.yaml.
> 
This makes it a "flavor", and not only a "feature" anymore ;)

> In the end, not all features of our publishing process are actually needed here, not sure if we can disable those, such as distributing the images to all the clouds, etc.. What we need in the end is only that those images are hosted somewhere (currently on gcp, aws is fine too) with a predictable url schema.
> 
> If there is a better way to do this, let me know @yeoldegrove. I know I have gaps in my knowledge of the publishing workflow.
If we want to treat lima as any other flavor, then this is build, tested and published (to S3 and ghcr.io).
In addition to that, custom yaml files as mentioned in #3781 are not part of this process and we need to think of another way to publish those. This could be a workflow that runs after the publishing (triggered automatically) or just some script that builds you the yaml from e.g. glrd data (this is maybe easier to achieve).

SujanaSubr:
I removed script which is used for fetching and starting lima image. This will be added as new PR once we have nightly image available for lima

PR #3819, opened 2025-11-06
cleanup: retire bootstrap stage

nkraetzschmar:
Timing with bootstrap stage:

- bootstrap build: 30s
- cache upload: 23s
- cache download: 1s
- build (kvm as example): 3m30s

Timing without bootstrap stage:

- build (kvm): 3m14s (of which 8s are spend on bootstrap)

So from this it looks like since the switch to `mmdebstrap` there really is little to no benefit of running the bootstrap phase as a shared action and it's more efficient to just quickly re-build the bootstrap for each target.

Related ADR https://github.com/gardenlinux/gardenlinux/pull/3816 => this is a pre-requisite before we can update to the new builder


NotTheEvilOne (approved):
Good finding :)

LGTM

PR #3827, opened 2025-11-07
pam faillock test

vivus-ignis:
**What this PR does / why we need it**: migration of https://github.com/gardenlinux/gardenlinux/blob/main/features/server/test/test_pam_faillock.py & https://github.com/gardenlinux/gardenlinux/blob/main/features/server/test/test_pam.py to the new test framework

Fixes https://github.com/gardenlinux/gardenlinux/issues/3826

### DIFF
--- a/tests-ng/test_pam_faillock.py
+++ b/tests-ng/test_pam_faillock.py
@@ -73,7 +73,4 @@ def test_common_password_pwhistory_pam_faillock(pam_config):
         module_contains="pam_pwhistory.so",
         arg_contains=["use_authtok", "remember=5", "retry=5"],
     )
-    assert (
-        len(results) == 1
-    ), "Last 5 user passwords should not be reused"
-
+    assert len(results) == 1, "Last 5 user passwords should not be reused"


vivus-ignis:
> We actually have a plugin to parse pam entries: https://github.com/gardenlinux/gardenlinux/blob/e2f3103dfbbd43e2e6a355f13d8271c7b556c33e/tests-ng/plugins/pam.py @ByteOtter As you initially wrote this, do you see a benefit of using the pam plugin instead of the new file parsing plugin?

Not sure how another abstraction helps here. Besides that, pam module looks too specialized and obscures the goal of a test.

vivus-ignis:
> @yeoldegrove Yes, I think any PAM related tests should use the PAM plugin as it offers much more structured parsing of the individual fields and guardrails agains comments, whitespaces or any kind of abnormalities in the files.

I'm not getting this fascination with abstractions. I don't think that there is need even for file_content fixture.
But how something like this

```
    candidates = pam_config.find_entries(
        type_="password",
        control_contains={"success": "*", "default": "ignore"},
        match_all=True,
    )
    assert (
        len(candidates) == 1
    ), f"Expected exactly one password entry, found {len(candidates)}: {candidates}"
 ```
 
 is better than something like this
 
 ```
     file_content.check_lines(
        "/etc/pam.d/common-auth",
        [
            "auth required pam_faillock.so preauth silent audit deny=5 unlock_time=900",
            "auth [default=die] pam_faillock.so authfail silent audit deny=5 unlock_time=900",
        ],
    )
 ```
 
Abstractions are good when they make understanding and managing something easier, I'm afraid it's not the case with the pam fixture. The goal of tests-ng is to test distribution build for correctness. Why we need these massive fixtures that need to be tested by themselves?

vivus-ignis:
> I know other test also still do this (needs cleanup) but we declared parameterize somewhat an anti-pattern if it is not of clear use. https://github.com/gardenlinux/gardenlinux/blob/0e5025ced68a6712e4a86234c07b583ab14bc045/tests-ng/DEVELOPER.md#parametrization

I don't see how I can get rid of `@pytest.mark.parametrize` decorator with the way `pam_config` is defined. `pam_config` function requires an argument.

yeoldegrove:
https://github.com/gardenlinux/gardenlinux/blob/0e5025ced68a6712e4a86234c07b583ab14bc045/tests-ng/plugins/pam.py#L99-L128

The example mentions how to use this. Did not try it out, though.

PR #3829, opened 2025-11-07
requisites for features/cis

### LINKED ISSUE #3598: Migrate `feature/cis` tests
# Goal: Migrate feature/cis tests

## Tasks 
- [ ] Get the leatest CIS Debian 12 benchmark files
- [ ] Prepare a proposal of which need to be fixed and which should be skipped (with reason for skipping)

saubhikdattagithub:
- CIS compliance with agreed compliances (`auditd, sshd, sysctl, selinux=permissive`)
- package `cis-hardening `would be used and config files would be created during the image build
- Readme: https://github.com/saubhikdattagithub/public/blob/main/cis-gl.md

Fixes #3598

image


features/cis/exec.late
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+# 1. trigger the conf files of cis-hardening
+/opt/cis-hardening/bin/hardening.sh --create-config-files-only --allow-unsupported-distribution &>/dev/null
+
+# 2. Disable the checks not needed/taken care alternately
```
nkraetzschmar: might look nicer to have a set of config files only here, and then use a loop where the `sed -i 's|audit|disabled|'` and `/opt/cis-hardening/etc/conf.d/` are part of the loop body to avoid repeating yourself

features/cis/exec.late
```diff
@@ -0,0 +1,62 @@
...
+sed -i 's|audit|disabled|' /opt/cis-hardening/etc/conf.d/syslog-ng_remote_host.cfg
+sed -i 's|audit|disabled|' /opt/cis-hardening/etc/conf.d/remote_syslog-ng_acl.cfg
+sed -i 's|audit|disabled|' /opt/cis-hardening/etc/conf.d/enable_cron.cfg
+sed -i 's|audit|disabled|' /opt/cis-hardening/etc/conf.d/disable_ipv6_router_advertisement.cfg
+sed -i 's|audit|disabled|' /opt/cis-hardening/etc/conf.d/disable_ip_forwarding.cfg
+sed -i 's|audit|disabled|' /opt/cis-hardening/etc/conf.d/disable_source_routed_packets.cfg
+#sed -i 's|audit|disabled|' /opt/cis-hardening/etc/conf.d/check_distribution.cfg
+sed -i 's|audit|disabled|' /opt/cis-hardening/etc/conf.d/run_shm_nodev.cfg
+sed -i 's|audit|disabled|' /opt/cis-hardening/etc/conf.d/run_shm_noexec.cfg
+sed -i 's|^\(EXCEPTIONS=\"[^\"]*\)\"|\1 /usr/lib/dbus-1.0/dbus-daemon-launch-helper /usr/lib/polkit-1/polkit-agent-helper-1 /usr/bin/mailq /usr/sbin/nullmailer-queue"|' /opt/cis-hardening/etc/conf.d/find_suid_files.cfg
```
nkraetzschmar:
```
sed -i '/^EXCEPTIONS=/ s|"$/| /usr/lib/dbus-1.0/dbus-daemon-launch-helper /usr/lib/polkit-1/polkit-agent-helper-1 /usr/bin/mailq /usr/sbin/nullmailer-queue"|'
```

might be a cleaner option here to:
- first match lines starting with `EXCEPTIONS=`
- then only replace the final `"` on those lines with the added stuff

Results in the same, just might be easier to read

features/cis/exec.late
```diff
@@ -0,0 +1,62 @@
...
+sed -i 's|^\(EXCEPTIONS=\"[^\"]*\)\"|\1 /usr/libexec/systemd-cron/crontab_setgid"|' /opt/cis-hardening/etc/conf.d/find_sgid_files.cfg
```
nkraetzschmar: same here

features/cisAudit/exec.late
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env bash
```
nkraetzschmar: why move this from `exec.config` to `exec.late`?

features/cisOS/file.include/etc/selinux/config
```diff
@@ -3,7 +3,7 @@
 # enforcing - SELinux security policy is enforced.
 # permissive - SELinux prints warnings instead of enforcing.
 # disabled - No SELinux policy is loaded.
-SELINUX=enforcing
+SELINUX=permissive
```
nkraetzschmar, line 6: is this correct? sure currently `enforcing` breaks everything, but don't we need enforcing for cis and should fix the underlying issues instead?

features/cisPackages/pkg.include
```diff
@@ -2,5 +2,6 @@ git
 libpam-pwquality
 libpam-modules-bin
 logrotate
-syslog-ng
 tcpd
+apparmor
```
nkraetzschmar: apparmor? doesn't that go against the assumption that this uses selinux?

features/firewall/test/test_nft_config.py
```diff
@@ -1,8 +0,0 @@
-from helper.utils import execute_remote_command
```
nkraetzschmar, line 1: why delete tests in the old test framework for non CIS features? if they fail we should eval why and if we can feature gate them by adding a test exclude/skip file in cis. (Obviously all temporary till we fully migrate to the new farmework)

features/firewall/test/test_packages_musthave.py
```diff
@@ -1 +0,0 @@
-from helper.tests.packages_musthave import packages_musthave as test_packages_musthave
```
nkraetzschmar, line 1: as above

features/firewall/test/test_systemd_units.py
```diff
@@ -1,16 +0,0 @@
-import pytest
```
nkraetzschmar, line 1: as above

features/ssh/file.include/etc/ssh/sshd_config
```diff
@@ -1,3 +1,4 @@
+# Config from CIS feature
```
nkraetzschmar: this modifies it globally in the ssh feature though, not CIS only..... is that what we want? maybe this should be modified by `exec.config` inside cis features :thinking:

tests-ng/test_cis.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+
+@pytest.mark.feature("cis")
+@pytest.mark.root(reason="CIS audit requires root privileges")
+@pytest.mark.booted(reason="Must be run on a booted system")
+def test_debian_cis_audit(shell):
```
nkraetzschmar, line 8: this should probably be marked modifying as well. while maybe the tests only do non modifying stuff, I'm not sure we sholud run some arbitrary shell script that will be hard to audit for truly being non modifying in envs like live gardener nodes.

nkraetzschmar: or otherwise the script should be wrapped in something like landlock or run as non root user

nkraetzschmar (changes requested):

features/cisOS/file.include/etc/selinux/config
saubhikdattagithub: Yes, we discussed this in the same round on 23.Oct and 29.Oct, it was agreed for `SELINUX=permissive` for now with CIS compliance checks. 
Effort on `SELINUX=enforcing` would be a parallel effort for sometime later stage which would be hosted separately

features/cis/exec.config
```diff
@@ -0,0 +1,175 @@
...
+done
+
+# 2. Add exceptions to the suid and sgid marked daemons
+declare -A EXCEPTIONS_MAP=(
+  [find_suid_files.cfg]="/usr/lib/dbus-1.0/dbus-daemon-launch-helper /usr/lib/polkit-1/polkit-agent-helper-1 /usr/bin/mailq /usr/sbin/nullmailer-queue"
+  [find_sgid_files.cfg]="/usr/libexec/systemd-cron/crontab_setgid"
+)
+
+for cfg in "${!EXCEPTIONS_MAP[@]}"; do
+  sed -i "s|^EXCEPTIONS=\"\(.*\)\"$|EXCEPTIONS=\"\1 ${EXCEPTIONS_MAP[$cfg]}\"|" \
```
nkraetzschmar, line 70: I like the use of associative arrays here :+1:

but I still think the sed expression could be improved to be more legible as already mentioned in https://github.com/gardenlinux/gardenlinux/pull/3829#discussion_r2522794828

features/cis/exec.config
```diff
@@ -0,0 +1,175 @@
...
+    "/opt/cis-hardening/etc/conf.d/$cfg"
+done
+
+# 3. Logrotate permissions
+sed -i 's|create|create 0640 root root|' /etc/logrotate.conf
+
+# 4. Append sshd_config
```
nkraetzschmar: Is just appending correct here? How does sshd handle cases where previous parts of the config might have already defined the keys we set here?

features/cisAudit/exec.late
nkraetzschmar: Unresolved this, as I don't think it should be resolved without at least documenting the decision why this should be moved here in the PR

nkraetzschmar (changes requested):

features/cisAudit/exec.late
saubhikdattagithub: With the `cisAudit/exec.config`, the auditd rules wont apply which gets applied as expected and fine with `cisAudit/exec.late
`
- reproduce
- `mv features/cisAudit/exec.late features/cisAudit/exec.config`
 - `./build kvm-cis_dev`
 - within the container, `/opt/cis-hardening/bin/hardening.sh --audit --allow-unsupported-distribution` would result in 
 
 
image
image
image

features/cis/exec.late
saubhikdattagithub: Ah missed this, Okay

features/cis/exec.config
saubhikdattagithub: Yes missed this comment, noted, will do

features/cis/exec.config
saubhikdattagithub: - Ideally, `features/cisSshd/file.include/etc/ssh/sshd_config `should serve the purpose
- but since `features/ssh` is called by default, not sure how to force/replace the `/etc/ssh/sshd_config` with `features/cisSshd/file.include/etc/ssh/sshd_config`
- would need guidance here if we can override a file which comes with `base` or any `mandatory` package and we need to replace/amend/overwrite at later user space level

features/cisAudit/exec.late
nkraetzschmar: In that case please identify what other features `exec.config` or `exec.late` conflicts with running these steps in `cisAudit/exec.late`, cause functionally running it in either one should be identical, so if that is not he case, then moving them to late is likely just hiding some other underlying issue and we should identify that instead.

features/cis/exec.config
nkraetzschmar: I think as long as the feature `cisSshd` defines `ssh` as one of its depends/includes then the order of include files from ssh first and then overwrite with files from `cisSshd` should be enforced by the builder.

features/cisAudit/exec.late
saubhikdattagithub: Yes, `features/log` conflicted the audit.rules placement with `features/cisAudit`. Maintaining what `features/log` does, `exec.config` does the job with a new file /etc/audit/rules.d/99-cis.rules being placed by cisAudit against utilizing the `exec.late` doing it previously. Agreed and implemented. Kindly review

features/cis/exec.config
saubhikdattagithub: yes, observed as described

features/ssh/test/test_sshd.py
```diff
@@ -1,48 +1,40 @@
-from helper.tests.sshd import sshd
```
fwilhe, line 1: why reformat this file in this pr? not needed and make a anyway large diff even larger

features/cisSysctl/file.include/etc/sysctl.d/99-cis.conf
```diff
@@ -1,17 +1,18 @@
...
-net.ipv4.conf.default.log_martians=1
-net.ipv4.conf.all.rp_filter=1
-net.ipv4.conf.default.rp_filter=1
-net.ipv6.conf.all.disable_ipv6=1
-net.ipv6.conf.default.disable_ipv6=1
-net.ipv6.conf.lo.disable_ipv6=1
-net.ipv4.conf.all.send_redirects=0
-net.ipv4.conf.default.send_redirects=0
-fs.suid_dumpable=0
+net.ipv4.conf.default.accept_redirects = 0
```
fwilhe: why reformatting?

features/cisOS/file.include/etc/selinux/config
fwilhe: Please properly document this decision. might be a comment, might be an ADR, but 'someone said in a meeting' is no documentation.

features/ssh/test/test_sshd.py
saubhikdattagithub: Yes, admitted. The file is reverted as original with `non_cis` addition as required

features/cisSysctl/file.include/etc/sysctl.d/99-cis.conf
saubhikdattagithub: Yes admitted. Reformatted the file as needed. VS/terminal compatibility issue

features/cisOS/file.include/etc/selinux/config
saubhikdattagithub: @fwilhe Ok, ADR is proposed here pls https://github.com/gardenlinux/gardenlinux/pull/4035

features/cisSysctl/file.include/etc/sysctl.d/99-cis.conf
```diff
@@ -11,7 +11,8 @@ net.ipv4.conf.all.rp_filter=1
 net.ipv4.conf.default.rp_filter=1
 net.ipv6.conf.all.disable_ipv6=1
 net.ipv6.conf.default.disable_ipv6=1
-net.ipv6.conf.lo.disable_ipv6=1
+net.ipv6.conf.lo.disable_ipv6=0
```
fwilhe, line 14: why are we doing this? does not look like a setting we should disable.

features/cisSysctl/file.include/etc/sysctl.d/99-cis.conf
```diff
@@ -11,7 +11,8 @@ net.ipv4.conf.all.rp_filter=1
...
 net.ipv4.conf.all.send_redirects=0
 net.ipv4.conf.default.send_redirects=0
+net.ipv4.conf.default.accept_source_route=0
```
fwilhe, line 17: rationale for this is unclear to me.

features/cisSshd/exec.config
```diff
@@ -1,5 +1,102 @@
...
 set -Eeuo pipefail
 
-# 5.2.1: Define sshd conf permission
+# Define sshd conf permission
 chmod 600 /etc/ssh/sshd_config
+
+# Append sshd_config
+cat << 'EOF' >> /etc/ssh/sshd_config
+
+#### CIS-HARDENING + CUSTOM ####
```
fwilhe: what does 'custom' mean?

it looks like a lot of this config is duplicated from what we have anyway in any image with the ssh feature. do we really need this?

features/cisAudit/file.include/etc/audit/rules.d/99-cis.rules
```diff
@@ -0,0 +1,246 @@
+# https://www.stigviewer.com/stig/canonical_ubuntu_20.04_lts/2023-09-08/finding/V-238256
```
fwilhe: out of curiosity, is this file taken from my stig feature? just wondering if this is the right config for CIS. also the stigviewer comments might not make sense in this context, they were intended to proof that we implement (most of) the ubuntu STIG at some point in time, but STIG and CIS are different, so I don't see the value of those references here

features/cis/exec.config
```diff
@@ -0,0 +1,173 @@
...
+  check_distribution.cfg
+  run_shm_nodev.cfg
+  run_shm_noexec.cfg
+)
+
+for cfg in "${CONFIGS[@]}"; do
+    sed -i 's|audit|disabled|' "/opt/cis-hardening/etc/conf.d/$cfg"
+done
+
+# 2. Add exceptions to the suid and sgid marked daemons
```
fwilhe: counting is off, we had '2.' already

features/cis/exec.config
```diff
@@ -0,0 +1,173 @@
...
+    "/opt/cis-hardening/etc/conf.d/$cfg"
+done
+
+# 3. Logrotate permissions
+sed -i 's|create|create 0640 root root|' /etc/logrotate.conf
+
+# 4. Append sshd_config
+cat << 'EOF' >> /etc/ssh/sshd_config
+
+#### CIS-HARDENING + CUSTOM ####
```
fwilhe: why set ssh config here and in the cisSshd feature? should only be done once I assume

features/cisSysctl/file.include/etc/sysctl.d/99-cis.conf
saubhikdattagithub: https://github.com/gardenlinux/gardenlinux/issues/3859 - had this issue and it was confirmed we need IPV6 (Jan/Andre) - BTW, we are disabling disable_IPV6 = i.e enable IPV6

features/cisSysctl/file.include/etc/sysctl.d/99-cis.conf
saubhikdattagithub: we would not need source routed V4 packets to comply with CIS

features/cisSshd/exec.config
saubhikdattagithub: Yes, agree removed term "CUSTOM here" It meant EARLIER WORK+CURRENT REQUIREMENTS initially since i took over the task from a point.

For sshd_config however, to keep it from ONE-UNIFORM-PLACE, the entire sshd_config was added here while the feature /ssh is what `cis` depends on was added in `cis/info.yaml` which controls the overlap as understood (assisted by Niko above)

features/cisAudit/file.include/etc/audit/rules.d/99-cis.rules
saubhikdattagithub: Yes, this stigviewer part was already part of earlier work in cisAudit (ALREADY WORK IN PROGRESS before I got involved) and didnt give a thought as the auditd rules were correctly appended + my work consisted of additional requirements below without stigviewer. Corrected it now without the #stigviewer

features/cis/exec.config
saubhikdattagithub: Yes, corrected.

features/cis/exec.config
saubhikdattagithub: Yes, that was the remains of the confusion of sshd_config = who controls features/ssh or features/cisSshd, Checking and Fixing now...

features/cisSysctl/file.include/etc/sysctl.d/99-cis.conf
fwilhe: oh, the double negative. read it the wrong way. makes sense, thanks for clarifying

PR #3845, opened 2025-11-10
tests-ng server feature: enabled services

vivus-ignis:
**What this PR does / why we need it**: migration of https://github.com/gardenlinux/gardenlinux/blob/main/features/server/test/test_systemctl.py to the new test framework

Fixes https://github.com/gardenlinux/gardenlinux/issues/3757


yeoldegrove (approved):
Extending this test is fine with me for now but I really don't like this test (as the TODOs suggest) and it needs to be rethinked.

PR #3846, opened 2025-11-10
test-ng: keep python distribution up2date by opening PRs if a new version is out

### LINKED ISSUE #3844: test-ng: keep python distribution up2date

We're pulling our python distribution from test-ng from https://github.com/astral-sh/python-build-standalone/releases.
We should have a mechanism of being notified of new releases in a dependabot like fashion, by opening a PR and approving this manually. Dependabot cannot be used as our version file is custom.

To decide:
- Which python version to use?
  - same as latest GL nightly?
  - always the latest available?
- How to implement the update
  - bash
    - minimal dependencies
  - or python?
    - requires an additional `requirements-*.txt` file as dependencies differ from tests and dev env
- How to authenticate to Github for PR
  - allow github actions to create PRs (currently disabled)
  - create github actions application that has permissions to create PRs. Get a tmp token from it and create PR.
- How to open PR?
  - use `gh` cli
  - use https://github.com/marketplace/actions/create-pull-request

**Why is this needed**:

Having an up2date python version is good from a security point of view.

yeoldegrove:

Keep test-ng's python distribution up2date by opening PRs if a new version is out.
This does not update the major.minor python version (still to be done manually), but parses `tests-ng/util/python.env.sh` for `PYTHON_VERSION_SHORT` and updates `PYTHON_VERSION`, `RELEASE_DATE` and `PYTHON_ARCHIVE_CHECKSUM_(AMD|ARM)64`.

Fixes #3844

**Notes for reviewers**:

This is an example PR that would be created:
- https://github.com/gardenlinux/gardenlinux/pull/4000


.github/workflows/test_update_python_runtime.yml
```diff
@@ -0,0 +1,81 @@
...
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      # Generate a token that is allowed to create pull requests
+      - uses: actions/create-github-app-token@67018539274d69449ef7c02e8e71183d1719ab42 # pin@v2.1.4
+        id: generate-token
+        with:
+          app-id: ${{ secrets.APP_BOT_PR_APP_ID }}
+          private-key: ${{ secrets.APP_BOT_PR_PRIVATE_KEY }}
```
fwilhe: where do those secrets come from? do they need to be maintained/rotated?

.github/workflows/test_update_python_runtime.yml
```diff
@@ -0,0 +1,81 @@
...
+              echo "EOF"
+            } >> $GITHUB_OUTPUT
+
+            release_date=$(grep '^export RELEASE_DATE=' tests-ng/util/python.env.sh | sed -n 's/^export RELEASE_DATE="\(.*\)"/\1/p')
+            echo "release_date=$release_date" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create Pull Request
+        if: steps.changes.outputs.has_changes == 'true'
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # pin@v7.0.8
```
fwilhe: I would try to keep the number of third-party (non github) maintained actins low, this can also be done using the `gh` cli tool like in [this example](https://github.com/gardenlinux/package-linux/blob/5186e9f6ee148f0cef60a071bdad41e96bac5870/.github/workflows/pr-if-new-kernel.yml#L39)

tests-ng/util/update_runtime.sh
```diff
@@ -0,0 +1,325 @@
+#!/usr/bin/env bash
```
fwilhe: isn't it a bit ironic to have this non trivial bash program to keep python updated? any reason not do implement this in python?

.github/workflows/test_update_python_runtime.yml
yeoldegrove: These credentials are bound to a "Github App" called "gardenlinux-bot-pr" and the private key does not expire (and cannot expire). The token - that is created using this credentials - is short lived.
- https://github.com/organizations/gardenlinux/settings/installations/94024387
- https://github.com/organizations/gardenlinux/settings/apps/gardenlinux-bot-pr

The documentation is found inside the "create-pull-request" action:
- https://github.com/peter-evans/create-pull-request/tree/271a8d0340265f705b14b6d32b9829c1cb33d45e/?tab=readme-ov-file#token
- https://github.com/peter-evans/create-pull-request/blob/271a8d0340265f705b14b6d32b9829c1cb33d45e/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens

The Github App has it's secrets set in a special environment called "app_bot_pr" (which is restricted to `main` branch).

tests-ng/util/update_runtime.sh
yeoldegrove: Also asked me that question but went with bash as currently everything "build" related in `tests-ng/util` is bash.
Not sure if it would be way less "non trivial" in python :shrug: but would not be opposed to re-write it.
@nkraetzschmar What would be your opinion on that?

.github/workflows/test_update_python_runtime.yml
yeoldegrove: This could be made obsolete if we allowed Github Actions to create pull requests in the "gardenlinux/gardenlinux" (like in e.g. the gardenlinux/package-linux repo).

image

.github/workflows/test_update_python_runtime.yml
yeoldegrove: I would say that https://github.com/marketplace/actions/create-pull-request is well maintained but yes, we could also build this ourselves like you suggested.
If you insist on that, I can change it.

tests-ng/util/update_runtime.sh
yeoldegrove: Currently we do not have a mechanism to keep track of python dependencies (e.g. PyGithub) that are not supposed to be in the packaged distribution (except for `requirements-dev.txt`), a `requirements-gh.txt` or similar could be introduced.

yeoldegrove: Just checked, could provide a ~220 lines python script instead of the ~320 lines bash script, but the "issue" of having an additional requirements file (different purpose) remains.

PR #3867, opened 2025-11-12
Fix: Remove content from `/etc/hostname` for container builds

### LINKED ISSUE #3314: /etc/hostname differs for container flavor
**What happened:**
- `/etc/hostname` changes when building the flavors `container-arm64` and `container-amd64` two times each

**How to reproduce it**:
- Build a container flavor and save the .tar file, delete the `.build` directory and build the same container flavor again. In the untarred file systems, `/etc/hostname` has a different value
 
**TODO**
- Decide if the hostname can and should be generated at first boot up or if `/etc/hostname` should be added to the whitelist of allowed differences

Leon-hk:
- Container builds contain random generated hostnames in `/etc/hostname`, as they do not contain the `server` feature which removes the content of `/etc/hostname` in it's `exec.post` script
- Random generated values should not be shipped in an operating system image, as they should be unique
- Moves the `/etc/hostname` content removal from the `server` feature to the `base` feature, which is included by `server` and by `container`

Fixes #3314

### DIFF
--- a/features/base/exec.post
+++ b/features/base/exec.post
@@ -31,4 +31,5 @@ fi
 
 find "$rootfs/var/log/" "$rootfs/var/cache/" -type f -delete
 
-echo > "$rootfs/etc/hostname"
+rm "$rootfs/etc/hostname"
+echo -n > "$rootfs/etc/hostname"


features/base/exec.post
```diff
@@ -30,3 +30,6 @@ rm "$rootfs/etc/.resolv.conf.systemd-resolved.bak"
 fi
 
 find "$rootfs/var/log/" "$rootfs/var/cache/" -type f -delete
+
+rm "$rootfs/etc/hostname"
+echo -n > "$rootfs/etc/hostname"
```
vivus-ignis, line 35:
```suggestion
echo > "$rootfs/etc/hostname"
```

vivus-ignis (changes requested):

features/base/exec.post
NotTheEvilOne, line 1: Feature `base` describes itself as "This feature installs the `base` layer for Garden Linux.". I'm not sure if removing the `/etc/hostname` created during build of a flavor is a good idea for that kind of feature. It may become an unexpected side effect for other features.

NotTheEvilOne (changes requested):

features/base/exec.post
Leon-hk: The `base/exec.post`script is executed as the second latest script, right before `server/exec.post`, so other features are not affected by this change. And if we ship all gardenlinux flavors with an empty `/etc/hostname`, I think this can also count to the `base` layer as mentioned in the description.
We could also leave it in the `server` feature and copy it to the `container` feature. I have done this in the first commit of this PR, but it creates duplicated code.

NotTheEvilOne: Any opinion @gehoern?

features/base/exec.post
```diff
@@ -30,3 +30,5 @@ rm "$rootfs/etc/.resolv.conf.systemd-resolved.bak"
 fi
 
 find "$rootfs/var/log/" "$rootfs/var/cache/" -type f -delete
+
+echo > "$rootfs/etc/hostname"
```
nkraetzschmar: I think you'll need to do it the same as in the `server/exec.post` used to do it via

```
rm "$rootfs/etc/hostname"
echo -n > "$rootfs/etc/hostname"
```

Reason: `/etc/hostname` *may* be a symlink, in that case you'll write to the symlink destination, not replace the file with an empty one.

Leon-hk: Makes sense, changed it back to include the `rm` and `-n` flag. Thanks for the hint!

nkraetzschmar (approved):

PR #3868, opened 2025-11-12
Script to fetch Lima manifest

SujanaSubr:
Adds python script which fetched lima image using glrd 
Fixes #

**Definition of Done:**

**Release note**:

```feature user

```

### DIFF
--- a/.github/workflows/build_bare_flavor.yml
+++ b/.github/workflows/build_bare_flavor.yml
@@ -24,7 +24,7 @@ jobs:
     permissions:
       actions: write
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # pin@v5.0.1
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # pin@v6.0.0
         with:
           submodules: true
       - name: Set build reference
--- a/.github/workflows/build_flavor.yml
+++ b/.github/workflows/build_flavor.yml
@@ -41,11 +41,11 @@ jobs:
       USE_KMS: ${{ inputs.signing_env == '' && 'false' || 'true' }}
     environment: ${{ inputs.signing_env }}
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # pin@v5.0.1
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # pin@v6.0.0
         with:
           submodules: true
       - name: Install python-gardenlinux-lib
-        uses: gardenlinux/python-gardenlinux-lib/.github/actions/setup@hotfix-disabled-python-cache-in-setup-action
+        uses: gardenlinux/python-gardenlinux-lib/.github/actions/setup@0.10.6
       - name: Set build reference
         run: |
           echo "${{ inputs.commit_id }}" | tee COMMIT
@@ -58,7 +58,7 @@ jobs:
       - name: Configure aws credentials for kms signing
         id: aws_auth
         if: ${{ env.USE_KMS == 'true' }}
-        uses: aws-actions/configure-aws-credentials@00943011d9042930efac3dcd3a170e4273319bc8 # pin@v4
+        uses: aws-actions/configure-aws-credentials@61815dcd50bd041e203e49132bacad1fd04d2708 # pin@v4
         with:
           role-to-assume: ${{ secrets.aws_kms_role }}
           role-session-name: ${{ secrets.aws_oidc_session }}
@@ -71,7 +71,7 @@ jobs:
         run: make ${{ inputs.flavor }}-${{ inputs.arch }}-build
       - name: Determine CNAME
         id: cname
-        uses: gardenlinux/python-gardenlinux-lib/.github/actions/features_parse@hotfix-disabled-python-cache-in-setup-action
+        uses: gardenlinux/python-gardenlinux-lib/.github/actions/features_parse@0.10.6
         with:
           flags: --cname ${{ inputs.flavor }}-${{ inputs.arch }} cname
       - name: Set CNAME
--- a/.github/workflows/build_flavors_matrix.yml
+++ b/.github/workflows/build_flavors_matrix.yml
@@ -21,7 +21,7 @@ jobs:
     outputs:
       matrix: ${{ steps.matrix.outputs.matrix }}
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # pin@v5.0.1
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # pin@v6.0.0
         with:
           submodules: true
       - id: matrix
--- a/.github/workflows/build_kmodbuild_container.yml
+++ b/.github/workflows/build_kmodbuild_container.yml
@@ -20,7 +20,7 @@ jobs:
     permissions:
       actions: write
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # pin@v5.0.1
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # pin@v6.0.0
         with:
           submodules: true
       - id: build_container_cache
@@ -33,7 +33,7 @@ jobs:
       - if: ${{ steps.build_container_cache.outputs.cache-hit == 'true' }}
         name: Determine CNAME
         id: cname
-        uses: gardenlinux/python-gardenlinux-lib/.github/actions/features_parse@hotfix-disabled-python-cache-in-setup-action # pin@0.10.5
+        uses: gardenlinux/python-gardenlinux-lib/.github/actions/features_parse@0.10.6 # pin@0.10.6
         with:
           flags: --cname container-${{ matrix.arch }} cname
       - if: ${{ steps.build_container_cache.outputs.cache-hit == 'true' }}
--- a/.github/workflows/build_platform_test_images.yml
+++ b/.github/workflows/build_platform_test_images.yml
@@ -114,11 +114,11 @@ jobs:
     permissions:
       actions: write
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # pin@v5.0.1
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # pin@v6.0.0
         with:
           submodules: true
       - name: Install python-gardenlinux-lib
-        uses: gardenlinux/python-gardenlinux-lib/.github/actions/setup@hotfix-disabled-python-cache-in-setup-action # pin@0.10.5
+        uses: gardenlinux/python-gardenlinux-lib/.github/actions/setup@0.10.6 # pin@0.10.6
       - name: Set build reference
         run: |
           version="$(./bin/garden-version "$version")"
@@ -158,7 +158,7 @@ jobs:
     permissions:
       actions: write
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # pin@v5.0.1
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # pin@v6.0.0
         with:
           submodules: true
       - name: Set build reference
@@ -216,7 +216,7 @@ jobs:
     permissions:
       packages: write
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # pin@v5.0.1
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # pin@v6.0.0
         with:
           submodules: true
       - name: Load images (${{ matrix.platform }})
@@ -284,7 +284,7 @@ jobs:
       commit_id: ${{ steps.version_reference.outputs.commit_id }}
       version: ${{ steps.version_reference.outputs.version }}
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # pin@v5.0.1
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # pin@v6.0.0
         with:
           submodules: true
       - name: Prepare reference from GLRD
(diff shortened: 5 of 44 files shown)


bin/generate-lima-yaml.py
NotTheEvilOne, line 1: IMHO this is a candidate for a directory named `hack/bin`. `/bin` should be reserved for "binaries" required for the GardenLinux repository to do it's work. We could move `start-vm` there as well. Any opinions @gehoern, @pnpavlov?

bin/start-lima.py
```diff
@@ -0,0 +1,115 @@
...
+    "mountTypesUnsupported": ["9p"],
+    "ssh": {
+        "loadDotSSHPubKeys": True, 
+        "forwardAgent": True
+    },
+    "mounts": [{"location": "~", "writable": False}],
+}
+
+def get_image_path(version, arch):
+    script_dir = os.path.dirname(os.path.realpath(__file__))
```
NotTheEvilOne: Using `pathlib.Path` here would make code more readable :)

bin/generate-lima-yaml.py
```diff
@@ -0,0 +1,115 @@
...
+    try:
+        result = subprocess.run(command, capture_output=True, text=True, check=True)
+        output = json.loads(result.stdout)
+        image_name = f"openstack-gardener_prod-{arch}"
+        image_path = output["releases"][0]["flavors"][image_name]["image"]
+        return image_path
+    except subprocess.CalledProcessError as e:
+        print(f"Command failed with error: {e.stderr}")
+        raise
+    except KeyError as e:
```
NotTheEvilOne: If the expected JSON schema is more complex you may want to consider using `jsonschema` or similar for that purpose.

features/lima/README.md
```diff
@@ -4,17 +4,30 @@
...
 
-Garden Linux images for Lima are published at https://images.gardenlinux.io
+Garden Linux images for Lima are published with nightly or release
 
 How to use the pre-built image:
 
-1. Create and start the VM
+1. Prerequisities
+    - We use glrd to get image link, this needs 
+        - sudo access
```
yeoldegrove: This is only needed on mac I guess, on Linux this runs on user permissions. I would remove it or mention the mac part.

@SujanaSubr You could even try removing sudo from `bin/glrd` ... I have no idea how permissions regarding podman work on mac.

features/lima/README.md
```diff
@@ -4,17 +4,30 @@
...
+        - podman
+        - access to gardenlinux-glrd.s3.eu-central-1.amazonaws.com
 
+2. Create and start the VM
+
+Lima image can be started with start-lima.py inside bin folder
+```
+# for the latest nightly image, use:
+python3 bin/start-lima.py --name gardenlinux --version nightly --arch <arch>
```
yeoldegrove: I would remove the `python3` call and maybe reference what is supported by arch or mention the `--help` flag.

bin/generate-lima-yaml.py
yeoldegrove: I guess for convenience reason, placing it next to `bin/start-vm` is indeed nice.

bin/start-lima.py
```diff
@@ -0,0 +1,116 @@
...
+        raise
+    except json.JSONDecodeError as e:
+        print(f"Error decoding JSON: {e}")
+        raise
+
+def generate_yaml(image_path, name):
+    yaml_file = "gardenlinux.yaml"
+
+    yaml_config_data["images"][0]["location"] = image_path
+    yaml_config_data["images"][0]["arch"] = "amd64"
```
fwilhe: getting this error

```
FATA[0001] the YAML is invalid, saved the buffer as "lima.REJECTED.yaml": field `arch` must be one of [x86_64 aarch64 armv7l ppc64le riscv64 s390x]; got "amd64"
```

bin/generate-lima-yaml.py
fwilhe: Agree with @yeoldegrove, let's have it here

bin/glrd
```diff
@@ -2,7 +2,7 @@
 cmd=$(basename "${0}")
 args="$*"
 if [ -z "${GARDENLINUX_BUILD_CRE}" ]; then
-    GARDENLINUX_BUILD_CRE="sudo podman"
+    GARDENLINUX_BUILD_CRE="podman"
```
fwilhe, line 5: is this change related to this pr? why was there sudo in the first pace?

features/lima/README.md
```diff
@@ -4,22 +4,38 @@
...
+Garden Linux images for Lima are published with nightly or release
 
-How to use the pre-built image:
+**How to use the pre-built image:**
 
-1. Create and start the VM
+1. Prerequisities
+    - We use glrd to get image link, this needs 
+        - podman
+        - access to gardenlinux-glrd.s3.eu-central-1.amazonaws.com
```
fwilhe: how am I as a unser supposed to know if I have access to this? it is public, and the user does not need to know about this url, right?

features/lima/README.md
```diff
@@ -4,22 +4,38 @@
...
 
-Garden Linux images for Lima are published at https://images.gardenlinux.io
+Garden Linux images for Lima are published with nightly or release
 
-How to use the pre-built image:
+**How to use the pre-built image:**
 
-1. Create and start the VM
+1. Prerequisities
+    - We use glrd to get image link, this needs
```
fwilhe: the user should not have to know what glrd is at this point.. just say that podman is a requirement.. but that sounds a bit odd.. not sure if I'm happy with that

bin/start-lima.py
```diff
@@ -0,0 +1,116 @@
...
+        arch = "amd64"
+    else:
+        arch = args.arch
+
+    if args.name is None:
+        name = "gardenlinux"
+    else:
+        name = args.name
+    
+    if args.version is None:
```
fwilhe: instead of requiring this argument, i would vote for making `latest` the default. I think this is a sensible default, if you want to use a specific version, or a nightly, it makes sense that you need to diverge from that.

bin/start-lima.py
```diff
@@ -0,0 +1,116 @@
...
+        print("Error: Provide version value\n")
+        parser.print_help()
+        sys.exit(1)
+    else:
+        version = args.version
+        print(f"Starting Lima Image {version}")
```
fwilhe: terminology: 'lima image' does not make much sense imo.. it is a vm that is started from an image.
do we need to log this at all? if so, it should be reworded.
"Starting Garden Linux VM with version {version}" or similar

bin/start-lima.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+
+    parser.add_argument(
+        "--name",
+        const=None,
+        help="Provide name of the VM. Default:gardenlinux"
+    )
+
+    args = parser.parse_args()
+    if args.arch is None:
```
fwilhe: argparse has support for default values, please use that, cf https://docs.python.org/3/library/argparse.html#default

fwilhe: this will make those if-blocks redundant and simplify the code

bin/start-lima.py
```diff
@@ -0,0 +1,116 @@
...
+        "--version",
+        const=None,
+        help="Provide daily to start latest night build, latest to start latest release image \
+                or version number to start particular version of GL"
+    )
+
+    parser.add_argument(
+        "--arch",
+        const=None,
+        help="Provide architecture amd64 or arm64. Default:amd64"
```
fwilhe: needs to use the terminology used in lima, cf https://lima-vm.io/docs/reference/limactl_start/

I am fine with special case handling so that the user can also provide `amd64` or `arm64` for convenience

bin/start-lima.py
```diff
@@ -0,0 +1,116 @@
...
+
+    parser.add_argument(
+        "--name",
+        const=None,
+        help="Provide name of the VM. Default:gardenlinux"
+    )
+
+    args = parser.parse_args()
+    if args.arch is None:
+        arch = "amd64"
```
fwilhe: arch does not need to have a default value.. use the native architecture as a default value, that that is what lima does anyway

bin/start-lima.py
fwilhe: cf docs here https://lima-vm.io/docs/reference/limactl_start/

bin/glrd
SujanaSubr: I am using glrd in the script which forces sudo access but actually sudo is not required as per Eike's comment hence updated it with this PR. I verified and works fine without sudo

features/lima/README.md
SujanaSubr: When i tried in aws machine, glrd did not work because of some namespace issue , hence added this here so we ca check before running the script with ping command whether there is access

bin/start-lima.py
SujanaSubr: Currently i pass only specific arch image in the script, thats why it is necessary here. If we should definitely remove it, then i will adjust the script to include both image. Otherwise i will leave it as it is

bin/start-lima.py
```diff
@@ -0,0 +1,136 @@
...
+    # Get Latest nightly version
+    command = [glrd_path, "--latest", "--type", "nightly", "--output-format", "json"]
+    result = subprocess.run(command, capture_output=True, text=True, check=True)
+    latest_nightly_version = str(json.loads(result.stdout)["releases"][0]["version"]["major"]) + "." + str(json.loads(result.stdout)["releases"][0]["version"]["minor"])
+
+    if version == "nightly":
+        command = [glrd_path, "--latest", "--type", version, "--output-format", "json"]
+    elif version == "latest":
+        command = [glrd_path, "--latest", "--output-format", "json"]
+    elif (int(version.split(".")[0]) == latest_gl_major_version and int(version.split(".")[1]) > latest_gl_minor_version) or \
```
fwilhe: those conditions are hard to read

some suggestions:

- move `int(version.split(".")[0])` and `int(version.split(".")[1])` into a variable with a speaking name
- use helper functions to provide meaningful names for expressions
- maybe move error-case handling into an earlier if block. currently, error cases and working cases are mixed which makes code flow harder to follow. exit as early as possible if something goes wrong

bin/start-lima.py
```diff
@@ -0,0 +1,136 @@
...
+    },
+    "mounts": [{"location": "~", "writable": False}],
+}
+
+def get_image_path(version, arch):
+    script_dir = Path(__file__).resolve().parent
+    # Construct the path to the glrd script
+    glrd_path = str(script_dir / "glrd")
+
+    gl_minor_version_supporting_lima = [8, 16]
```
fwilhe: what are those numbers? I see they are used below but I don't really see what this does

bin/start-lima.py
```diff
@@ -0,0 +1,136 @@
...
+    
+    # Get Latest version supported by GL
+    command = [glrd_path, "--latest", "--output-format", "json"]
+    result = subprocess.run(command, capture_output=True, text=True, check=True)
+    latest_gl_major_version = json.loads(result.stdout)["releases"][0]["version"]["major"]
+    latest_gl_minor_version = json.loads(result.stdout)["releases"][0]["version"]["minor"]
+
+    # Get Latest minor version for 1592
```
fwilhe: suggestion: let's start with a clean slate and make this simpler. we don't need to support 1592 and 1877. let's start with nightly now, and let's support the next major version we will release. this special case handling for existing versions is not nice and not needed imo

bin/start-lima.py
```diff
@@ -0,0 +1,136 @@
...
+        raise
+    except KeyError as e:
+        print(f"Error parsing JSON: Missing key {e}")
+        raise
+    except json.JSONDecodeError as e:
+        print(f"Error decoding JSON: {e}")
+        raise
+
+def generate_yaml(image_path, name, arch):
+    yaml_file = "gardenlinux.yaml"
```
fwilhe: this leave the file on disk locally, which is not nice. I would prefer if this is in a temporary file that gets deleted when it is not needed anymore. bash has `mktemp`, I'm sure python has something similar for dedicated temporary files. the file should be cleaned up when it's not needed anymore, maybe with an optional flag to keep it if desired (for debugging purposes)

bin/start-lima.py
```diff
@@ -0,0 +1,136 @@
...
+    parser.add_argument(
+        "--version",
+        default='latest',
+        help="By default starts latest release version. \
+            Provide specific version or 'nightly' to start nightly build. Default:latest"
+    )
+
+    parser.add_argument(
+        "--arch",
+        default='amd64',
```
fwilhe: not sure why amd64 needs to be the default

we can get the user's native architecture via `platform.machine()`, it would be better UX to use that imo. Is there a reason we can't use that?

```python
>>> import platform
>>> platform.machine()
'aarch64'
```

features/lima/README.md
```diff
@@ -4,22 +4,37 @@
 This feature flag produces an image suitable for using with [lima](https://lima-vm.io)
 </website-feature>
 
-Garden Linux images for Lima are published at https://images.gardenlinux.io
```
fwilhe, line 7: for the docs, we also have the mention in the main README.md of the project, and we have a dedicated page in the docs folder.
Let's take care of the docs when the implementation is in a ready state

bin/start-lima.py
SujanaSubr: Not required anymore removed

bin/start-lima.py
```diff
@@ -0,0 +1,116 @@
...
+
+def generate_yaml(image_path, name, arch):
+    yaml_config_data["images"][0]["location"] = image_path
+    yaml_config_data["images"][0]["arch"] = "x86_64" if arch == "amd64" else "aarch64"
+
+    with tempfile.NamedTemporaryFile(suffix=".yaml", delete=True) as tmp:
+        with open(tmp.name, "w") as f:
+            yaml.dump(yaml_config_data, f, default_flow_style=False, sort_keys=False)
+        yaml_file = tmp.name
+        subprocess.run(["limactl", "start", yaml_file, "--name", name], check=True)
```
fwilhe: we need to provide the `--arch x86_64` or `--arch aarch64` flag here if we're running on the non-native architecture, I think

I am getting this error currently on my m1 mac

> FATA[0021] [skipped to download: "https://gardenlinux-github-releases.s3.amazonaws.com/objects/lima-amd64-2059.0.0-5178fb7c/lima-amd64-2059.0.0-5178fb7c.qcow2": unsupported arch: "x86_64"]


adding the flag here fixes that.

fwilhe: for reference, this is what I did

```diff
-        subprocess.run(["limactl", "start", yaml_file, "--name", name], check=True)
+        subprocess.run(["limactl", "start", yaml_file, "--arch", "x86_64", "--name", name], check=True)
```

of course, this may not hard code the architecture, did that just for testing. should be 'x86_64' or 'aarch64' based on what it really is

SujanaSubr: As per our latest discussion, removed this part from script. Now script generates only yaml file. It doesn't start VM

PR #3869, opened 2025-11-12
Fix/refactor find to work with Mac

Akendo:
In this PR, we add a bit of logic to detect `find` independent of the underlying operating system. 

When running the test-ng with the `--cloud` parameter, on a Mac this fails:

```
/src/gardenlinux/gardenlinux4/tests-ng/util/tf ~/src/gardenlinux/gardenlinux4/tests-ng
######################################################################## 100.0%
gpgv: Signature made Thu Nov  6 14:23:13 2025 CET
gpgv:                using RSA key E3E6E43D84CB852EADB0051D0C0AF313E5FD9F80
gpgv: Good signature from "OpenTofu (This key is used to sign opentofu providers) <core@opentofu.org>"
Installing OpenTofu v1.10.7
Downloading release tarball from https://github.com/opentofu/opentofu/releases/download/v1.10.7/tofu_1.10.7_darwin_arm64.zip
Downloading SHA hash file from https://github.com/opentofu/opentofu/releases/download/v1.10.7/tofu_1.10.7_SHA256SUMS
Downloading SHA hash signature file from https://github.com/opentofu/opentofu/releases/download/v1.10.7/tofu_1.10.7_SHA256SUMS.gpgsig
Archive:  /var/folders/yx/h2rdqf_d54d_hs2v_8cm_lnw0000gn/T/tofuenv_download.XXXXXX.nVFTzxLx6p/tofu_1.10.7_darwin_arm64.zip
  inflating: ~/src/gardenlinux/gardenlinux4/tests-ng/util/tf/.tofuenv/versions/1.10.7/CHANGELOG.md
  inflating: ~/src/gardenlinux/gardenlinux4/tests-ng/util/tf/.tofuenv/versions/1.10.7/LICENSE
  inflating: ~/src/gardenlinux/gardenlinux4/tests-ng/util/tf/.tofuenv/versions/1.10.7/README.md
  inflating: ~/src/gardenlinux/gardenlinux4/tests-ng/util/tf/.tofuenv/versions/1.10.7/tofu
Installation of tofu v1.10.7 successful. To make this your default version, run 'tofuenv use 1.10.7'
~/src/gardenlinux/gardenlinux4/tests-ng
find: -printf: unknown primary or operator
cat: ~/src/gardenlinux/gardenlinux4/tests-ng/util/tf/.tofuenv/version: No such file or directory
Version could not be resolved (set by ~/src/gardenlinux/gardenlinux4/tests-ng/util/tf/.tofuenv/version or tofuenv use <version>)
⚙️   cleaning up cloud resources
cat: ~/src/gardenlinux/gardenlinux4/tests-ng/util/tf/.tofuenv/version: No such file or directory
Version could not be resolved (set by ~/src/gardenlinux/gardenlinux4/tests-ng/util/tf/.tofuenv/version or tofuenv use <version>)
```

The main problem is here:

```
find: -printf: unknown primary or operator
```

Since by default Mac has a different version of the `find` command that does not support the `-printf` parameter. This code refactors to allow the use of the `find` command with parameters that are supported on GNU `find` and Mac `find`.

Additionally, we clean up the code to make it more readable, with some more notes on it.

### DIFF
--- a/tests-ng/util/install_tofu.sh
+++ b/tests-ng/util/install_tofu.sh
@@ -8,66 +8,49 @@ function die {
     exit 1
 }
 
-
-function test_for_supported_os(){
-	case "$(uname -s)" in
-        Darwin)
-            host_os=darwin
-            ;;
-        Linux)
-            host_os=linux
-            ;;
-        *)
-            die "Host operating system '$host_os' not supported"
-            ;;
-	esac
+function test_for_supported_os() {
+    case "$(uname -s)" in
+    Darwin)
+        host_os=darwin
+        ;;
+    Linux)
+        host_os=linux
+        ;;
+    *)
+        die "Host operating system '$host_os' not supported"
+        ;;
+    esac
 }
 
-function test_for_supported_cpu_architecture(){
-	case "$(uname -m)" in
-        x86_64)
-            host_arch=amd64
-            ;;
-        aarch64 | arm64)
-            host_arch=arm64
-            ;;
-        *)
-            die "Host architecture '$host_arch' not supported"
-            ;;
-	esac
+function test_for_supported_cpu_architecture() {
+    case "$(uname -m)" in
+    x86_64)
+        host_arch=amd64
+        ;;
+    aarch64 | arm64)
+        host_arch=arm64
+        ;;
+    *)
+        die "Host architecture '$host_arch' not supported"
+        ;;
+    esac
 }
 
-function find_local_version_of_tofuenv_and_return_version_string(){
-    # This find will detect the folder from the local tofuenv with it's installation and 
-    # extract the version field out. The result will be sorted. The highest version should 
-    # be s on top. Then we cut to a single parameter, otherwise the tofuenv use will fail.
-    # shellcheck disable=SC2046
-    basename $(\
-        find "$tf_dir/.tofuenv/versions"\
-            -mindepth 1 \
-            -maxdepth 1 \
-            -type d \
-            -print \
-            | sort -r \
-            | head -1
-        )
-}
-
-function install_custom_azure_resource_manager(){
+function install_custom_azure_resource_manager() {
     # We have to install the custom Resource Manager to use Secure Boot.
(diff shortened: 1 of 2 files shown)


tests-ng/util/install_tofu.sh
```diff
@@ -1,9 +1,34 @@
...
+    case "$(uname -o)" in
+        Darwin)
+            FIND="/opt/homebrew/bin/gfind"
+            RETRY="/opt/homebrew/bin/retry"
+            test -x $FIND || die "Can't find find. Please install it with 'brew install findutils'"
+            test -x $RETRY  || die "Can't find find. Please install it with 'brew install findutils'"
+            ;;
+        GNU/Linux)
+            FIND="/usr/bin/find"
+            RETRY="/usr/bin/retry"
```
yeoldegrove: Binaries might not always be in `/usr/bin/` and e.g. might break some development setups.
Please use:

```
command -v find
command -v retry
```

yeoldegrove: And do we really need this for retry, too? It is in my path on my system. Not sure how this behaves on MacOS.

Akendo: The problem is that you will `find` find. It's the BSD variant of [MacOS](https://manp.gs/mac/1/find) and that does not have the necessary `printf` option. Either we rewrite the command with a `printf` option or factor it out. But I'm not super into learning BSD `find` from macOS. 

But I get that this might break something. Maybe we can use the `command -v find` for the Linux version?

yeoldegrove: We could simply switch to some command that does not need `find .. -printf` and does work on all supported operating systems.

`command -v find` was supposed for the linux version.

Akendo: I refactor the `find`, I dislike the solution, but I guess it will do.

Akendo:
@yeoldegrove, I've updated the PR with the description from the ticket issue.

Issue #3870, opened 2025-11-12
FIPS: Enable OpenSSH to use right KEX and Cipher suit for FIPS


Akendo:
We discovered a problem before merging https://github.com/gardenlinux/gardenlinux/pull/3744, the CI system reported a problem when connecting to the target system with OpenSSH. The problem seems to be that the conenction become stall.

<img width="1956" height="583" alt="Image" src="https://github.com/user-attachments/assets/da977a42-924a-4613-8819-d3f5aa659ed6" />

The code does not time out, and it did run for 6 hours until the workflow timeout was reached. The main problem, when adding the debug was that the connection failed. 

```
debug1: SSH2_MSG_KEXINIT sent
debug1: SSH2_MSG_KEXINIT received
debug1: kex: algorithm: sntrup761x25519-sha512
debug1: kex: host key algorithm: ssh-ed25519
debug1: kex: server->client cipher: chacha20-poly1305@openssh.com MAC: <implicit> compression: none
debug1: kex: client->server cipher: chacha20-poly1305@openssh.com MAC: <implicit> compression: none
debug1: expecting SSH2_MSG_KEX_ECDH_REPLY
Connection closed by 3.121.114.171 port 2

```

Debugging it on the machine showed the following error message:

```
Nov 12 14:20:05 ip-10-0-1-61 sshd-session[2319]: Connection from 165.1.187.218 port 9164 on 10.0.1.61 port 22 rdomain ""
Nov 12 14:20:06 ip-10-0-1-61 sshd-session[2319]: ssh_dispatch_run_fatal: Connection from 165.1.187.218 port 9164: invalid argument [preauth]
Nov 12 14:20:06 ip-10-0-1-61 sshd[1056]: srclimit_penalise: ipv4: new 165.1.187.218/32 deferred penalty of 1 seconds for penalty: connections without attempting authentication
```

The problem is that sshd is not configured to use the correct ciphers and key exchange algorithm.

PR #3871, opened 2025-11-12
Enabled OpenSSH KEX and Cipher for FIPS.

Akendo:
In this PR, we enable the FIPS option of OpenSSH to enable ciphers and key exchange algorithms.

Solves: #3870 
Relates to #3744 ,#3436

### DIFF
--- a/tests-ng/test_ssh.py
+++ b/tests-ng/test_ssh.py
@@ -50,11 +50,14 @@ required_sshd_config = {
 }
 
 
+FIPS_REASON = "FIPS uses different values for the KEX and Cipher."
+
+
 @pytest.mark.booted(reason="Calling sshd -T requires a booted system")
 @pytest.mark.root(reason="Calling sshd -T requires root")
 @pytest.mark.feature("ssh")
 @pytest.mark.parametrize("sshd_config_item", required_sshd_config)
-@pytest.mark.feature("not _fips", reason="FIPS uses different values for the KEX and Cipher.")
+@pytest.mark.feature("not _fips", reason=FIPS_REASON)
 def test_sshd_has_required_config(sshd_config_item: str, sshd: Sshd):
     actual_value = sshd.get_config_section(sshd_config_item)
     expected_value = required_sshd_config[sshd_config_item]


features/_fips/exec.config
```diff
@@ -36,3 +36,7 @@ sed -i \
 # Disabled HTTPS for the apt mirror otherwise we can't keep using
 # apt.
 sed -i 's/https/http/' /etc/apt/sources.list.d/gardenlinux.sources
+
+# Enable FIPS KexAlgorithms and Ciphers for OpenSSH Server.
+echo "KexAlgorithms ecdh-sha2-nistp256,ecdh-sha2-nistp384,ecdh-sha2-nistp521,diffie-hellman-group-exchange-sha256" >> /etc/ssh/sshd_config
+echo "Ciphers aes256-gcm@openssh.com,aes256-ctr,aes128-gcm@openssh.com,aes128-ctr"  >> /etc/ssh/sshd_config
```
moritzkirchnersap, line 42: Do we know if we need HostkeyAlgorithms as well?

Would be great to have this in an integration test, no?

Akendo: You're correct. This is the first part of the PR. It's rather minimal to unblock the PR https://github.com/gardenlinux/gardenlinux/pull/3744, the `HostkeyAlgorithms` is for the SSH client and as of now shouldn't bother us. We do have a test for SSH, but that requires some adjustment. I have disabled it for now for the FIPS feature. Once the basic sets of tests for FIPS (https://github.com/gardenlinux/gardenlinux/pull/3752) are merged, I can the SSH test going.

yeoldegrove, line 42: Just out of curiosity (not blocking this) is appending to the file the way to go here? Usually we include files on a per file basis, which should be possible for sshd with `Include`.

tests-ng/test_ssh.py
```diff
@@ -54,6 +54,7 @@
 @pytest.mark.root(reason="Calling sshd -T requires root")
 @pytest.mark.feature("ssh")
 @pytest.mark.parametrize("sshd_config_item", required_sshd_config)
+@pytest.mark.skipif(feature="_fips", reason="FIPS uses different values")
```
yeoldegrove: We're currently not using `@pytest.mark.skipif` and it is not part of the [common filtering patterns](https://github.com/gardenlinux/gardenlinux/blob/0e5025ced68a6712e4a86234c07b583ab14bc045/tests-ng/DEVELOPER.md#common-filtering-patterns) currently. We use `@pytest.mark.feature("not _fips")` (which is a custom plugin) throughout many tests currently

Akendo: What's the reason for not using the `pytest` define way of skipping tests?

yeoldegrove: Pytest does not know what a "feature" is, so we needed to extend it like we did in `tests-ng/plugins/features.py`.

Akendo: That's curious, since my test was skipping. 🤔 By your logic, the test must have failed with an error message. I might have to check if my assumption is correct here; when different features are selected, might  just be working by chance then.

Akendo: Anyway, I did add this the way you recommend. /close

features/_fips/exec.config
Akendo: The pattern you describe here is a modification introduced by Debian to many packages. This is not enabled for OpenSSH. The only file that will be included for configuration is the [`/etc/ssh/sshd_config`](https://manpages.debian.org/trixie/openssh-server/sshd_config.5.en.html#FILES)

yeoldegrove: Even the manpage you linked talks about `Include`, though: https://manpages.debian.org/trixie/openssh-server/sshd_config.5.en.html#Include

Akendo: TIL. We might can do this then later on.

Akendo: Missed that, thanks for point it out.

yeoldegrove (approved):

features/_fips/exec.config
Akendo: I'm going to put this in later, when I refactor it with the tests.

PR #3875, opened 2025-11-13
docs: add ADR - 26. Use Parser Plugins for Files and Command Output in Test Framework

### LINKED ISSUE #3865: test-ng: ADR - When to use parsing abstractions for files and command output
**Description:**

Create an Architecture Decision Record (ADR) documenting the decision on when to use parsing abstractions (plugins) versus direct file/command parsing in the test framework.

**Tasks:**

- [ ] Create ADR file in `docs/architecture/decisions/` (next number in sequence)
- [ ] Document the context and problem statement
- [ ] Document the decision and rationale
- [ ] Document consequences (positive and negative)
- [ ] Include examples and decision framework
- [ ] Reference related PRs and discussions

**Acceptance Criteria:**

- Decision is clearly documented with rationale
- Examples illustrate the decision framework
- ADR is linked from DEVELOPER.md

yeoldegrove:

This ADR will document the decisions take on when to Use Parsing Abstractions for Files and Command Output in Test Framework.

Currently it includes 2 commits, a broader scoped one and a reworked version more focusing on the usage of parsing plugins.

Fixes #3865

### DIFF
--- a/docs/architecture/decisions/0026-test-ng-when-to-parsers.md
+++ b/docs/architecture/decisions/0026-test-ng-when-to-parsers.md
@@ -1,10 +1,10 @@
-# 25. Use Parser Plugins for Files and Command Output in Test Framework
+# 26. Use Parser Plugins for Files and Command Output in Test Framework
 
 **Date:** 2025-11-19
 
 ## Status
 
-Draft
+Accepted
 
 ## Context
 


fwilhe (changes requested):
Don't merge as is, please. I am not sure if this is actually helpful for alignment. Let's gather feedback from the team before assuming that silence is agreement.

docs/architecture/decisions/0022-when-to-use-parsing-abstractions.md
```diff
@@ -0,0 +1,81 @@
...
+
+Key questions that arose:
+
+- When should we create or use parsing plugins versus parsing directly in tests?
+- When is it acceptable to read files directly or parse command output inline?
+- How do we balance abstraction benefits (consistency, reusability) with test readability?
+
+## Decision
+
+We will use parsing abstractions (implemented by plugins) when they provide clear benefits in terms of:
```
Akendo: First of all, thank you for writing this document. ADR is a form of work I enjoy the least and I can appreciate everyone doing it as we.

The main concern this ADR has for me is that this is not what the main discussion was about. We didn't have a discussion about abstraction, we had a discussion about parsing. This document feels misleading and does not provide any clarity.

This might be my bad, since I'm the initiator of this change and did not do a good job of communicating this. But abstraction is the process of trying generalizing a problem domain. And since the term is overloaded I feel like we should avoid it here.

The key question here when to consider a plugin and when not and I think this is at best a rule "Case-by-case judgment without guidelines". 

The goal of this framework is to simplify the process of writing and conducting tests to allow external entries with little understanding of the system to comprehend the test at hand. By this logic, the degree of abstraction necessary follows this main use case. 

That leads to the issue that we have to inspect the different systems we aim for testing and see what's good to use. The main reason we need this lies in the problem domain: we have to understand the target state of a system that might include aspects of a configuration that might digress from our assumption of what to expect. 

In short, it's not a question of rahter we abstract or not. It's a question of whether the inquiry in question can be conducted with a simple readline or not. 

So please, let us sit down next week an revise this document.

fwilhe:
Trying to summarise my issues with this PR

It is a lot of text, but still quite abstract. The root of this discussion was `check_lines` and that it does way more than its name suggests. Not really sure where/if this is reflected in this document.

Maybe this should also be excluded from this PR as it is not really on the topic of parsing, but then it feels like we ditched one very important core aspect of this whole conversation.

In my understandings, the decisions in ADRs should be relatively short and concrete. I know this is a slippery slope, and you can did out my original ADRs on the test framework that don't follow this ideal very well.

Suggested path forward: Create multiple, smaller and more focussed ADRs, one per aspect

- Config file parsers
- Own parser implementation vs third party dependency 
- 'has line' checks (taking extra whitespace into account)

there might be more

those adrs should document the decision in sone specific, actual case and then be precedent for similar decisions in the future, they should not be very abstract documents like this, imo.

docs/architecture/decisions/0022-when-to-use-parsing-abstractions.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Date:** 2025-11-13
+
+## Status
+
+Draft
+
+## Context
+
+In the new Garden Linux tests framework, we need to parse various file formats and command outputs to verify system configuration and behavior. The framework provides plugins for parsing structured formats (JSON, YAML, INI, TOML) and domain-specific formats (PAM configs, SSH configs), but there has been inconsistency in when these abstractions should be used versus direct file/command parsing.
```
Akendo: Please write that this ADR is part of a bigger task to consider. Link the target issue.

ByteOtter (approved):
Looks good, thanks for the effort 👍🏼

docs/architecture/decisions/0025-test-ng-when-to-parsers.md
```diff
@@ -0,0 +1,87 @@
...
+
+> [!NOTE]
+> The parser plugins are planned to undergo refactoring to improve their API and usability.
+> This is tracked in https://github.com/gardenlinux/gardenlinux/issues/3866
+
+## Decision
+
+### Use the default parser plugins for file and command output parsing scenarios
+
+The default parser plugin should be the preferred approach for parsing files and command output because it:
```
fwilhe: you used 'default parser plugin' already in the context, but I think it would be good to mention the exact file or class name just to make it unambiguous

docs/architecture/decisions/0022-when-to-use-parsing-abstractions.md
```diff
@@ -0,0 +1,81 @@
+# 22. When to Use Parsing Abstractions for Files and Command Output in Test Framework
```
fwilhe: assuming this is the old version of the file that will be deleted before merge?

yeoldegrove: Yepp. Left it in for now in case somebody is interested in the differences.

docs/architecture/decisions/0025-test-ng-when-to-parsers.md
yeoldegrove: Thought about that, but as  is subject to change, I reverted that idea.
We could:
- Leave the ADR open until we aligned on the refactoring and adapt it later.
- Declare an ADR to be something that does not talk about those specifics. The `DEVELOPER.md`, it's core principles and it's examples section could be another place to mention those specifics.

fwilhe (changes requested):
Not requesting changes, but blocking a merge to allow for more reviews/approvals.

Deadline is Wednesday, 17:00 CET.

By then I'll mege this with `0022-test-ng-when-to-parsers.md` only (deleting the other file)

fwilhe (approved):
LGTM

Issue #3874, opened 2025-11-13
test-ng: less verbose test output


yeoldegrove:
**What would you like to be added**:

Currently these tests produce very verbose test output:
- test_no_failed_units
- test_dmesg_gardener_call_by_unprivileged_user_allowed

```
======================================================== PASSES ========================================================
_________________________________________________ test_no_failed_units _________________________________________________
------------------------------------------------- Captured stdout call -------------------------------------------------
Waiting for systemd to report the system is running took 0.01 seconds, with state 'running' and return code '0'.
________________________________ test_dmesg_gardener_call_by_unprivileged_user_allowed _________________________________
------------------------------------------------- Captured stdout call -------------------------------------------------
[    0.000000] Linux version 6.12.57-cloud-amd64 (debian-kernel@lists.debian.org) (x86_64-linux-gnu-gcc (Debian 15.2.0-7) 15.2.0, GNU ld (GNU Binutils for 
--> whole dmesg output follows
```

This output should be removed.

**Why is this needed**:

Test logs are not cluttered and important output if better visible.

PR #3877, opened 2025-11-13
do not log dmesg output and debug messages

vivus-ignis:

Test logs are not cluttered and important output if better visible.

Fixes https://github.com/gardenlinux/gardenlinux/issues/3874

### DIFF
--- a/tests-ng/test_basics.py
+++ b/tests-ng/test_basics.py
@@ -1,11 +1,8 @@
-import logging
 import os
 
 import pytest
 from plugins.shell import ShellRunner
 from plugins.systemd import Systemd
-
-logger = logging.getLogger(__name__)
 
 
 def test_gl_is_support_distro():
@@ -87,13 +84,13 @@ def test_kernel_not_tainted():
 @pytest.mark.booted(reason="Systemctl needs a booted system")
 def test_no_failed_units(systemd: Systemd, shell: ShellRunner):
     system_run_state = systemd.wait_is_system_running()
-    logger.debug(
+    print(
         f"Waiting for systemd to report the system is running took {system_run_state.elapsed_time:.2f} seconds, with state '{system_run_state.state}' and return code '{system_run_state.returncode}'."
     )
     failed_systemd_units = systemd.list_failed_units()
     for u in failed_systemd_units:
-        logger.debug(f"FAILED UNIT: {u}")
+        print(f"FAILED UNIT: {u}")
         shell(f"journalctl --unit {u.unit}")
     assert (
         not failed_systemd_units
-    ), f"{failed_systemd_units} systemd units failed to load"
+    ), f"{len(failed_systemd_units)} systemd units failed to load"


tests-ng/test_basics.py
```diff
@@ -83,13 +83,9 @@ def test_kernel_not_tainted():
 @pytest.mark.root(reason="Required for journalctl in case of errors")
 @pytest.mark.booted(reason="Systemctl needs a booted system")
 def test_no_failed_units(systemd: Systemd, shell: ShellRunner):
-    system_run_state = systemd.wait_is_system_running()
-    print(
-        f"Waiting for systemd to report the system is running took {system_run_state.elapsed_time:.2f} seconds, with state '{system_run_state.state}' and return code '{system_run_state.returncode}'."
-    )
```
yeoldegrove, line 89: I recently added a snippet on how to add debug logs: https://github.com/gardenlinux/gardenlinux/blob/85ca8d0e47df0c2c3cc68da6cf96313e64b5109d/tests-ng/DEVELOPER.md#adding-debug-logging
Do you think it is worth having this as a debug message for the line you removed?

vivus-ignis: Nice addition, will add debug messages then.

vivus-ignis: Done.

tests-ng/test_basics.py
```diff
@@ -84,12 +87,12 @@ def test_kernel_not_tainted():
 @pytest.mark.booted(reason="Systemctl needs a booted system")
 def test_no_failed_units(systemd: Systemd, shell: ShellRunner):
     system_run_state = systemd.wait_is_system_running()
-    print(
+    logger.debug(
         f"Waiting for systemd to report the system is running took {system_run_state.elapsed_time:.2f} seconds, with state '{system_run_state.state}' and return code '{system_run_state.returncode}'."
     )
     failed_systemd_units = systemd.list_failed_units()
     for u in failed_systemd_units:
-        print(f"FAILED UNIT: {u}")
```
fwilhe, line 92: this is not debug output, this is the failed case

fwilhe (changes requested):
I don't agree with the issue and I don't think this pr is helping us

tests-ng/test_basics.py
vivus-ignis: Failed units are all printed in case assertion fails.

vivus-ignis:
> I don't agree with the issue and I don't think this pr is helping us

It's not only about failed systemd units, this PR disables dmesg output as well (which was mistakenly enabled anyway).

tests-ng/test_basics.py
yeoldegrove: This is not that important as it is only outputted in case there is an error. I guess we could keep it a print.

NotTheEvilOne: I disagree here. If we hide which units fail we hide crucial information. I wonder why we do not use the assert below to print the list of failed units with Python formatting, similar to `f"{len(failed_systemd_units)} systemd units failed to load: {failed_systemd_units!r}"`.

Code changed should become better, not worse.

NotTheEvilOne (changes requested):
See comment above.

vivus-ignis:
> See comment above.

Fixed. Failed units are now logged.

tests-ng/test_basics.py
```diff
@@ -84,13 +87,13 @@ def test_kernel_not_tainted():
...
+        logger.debug(f"FAILED UNIT: {u}")
```
NotTheEvilOne: This line should still read `print` as below command prints systemd output for the unit to stdout/stderr. Otherwise it should be removed but not stay as changed to `logger.debug()`.

tests-ng/test_basics.py
```diff
@@ -84,13 +87,13 @@ def test_kernel_not_tainted():
...
         shell(f"journalctl --unit {u.unit}")
     assert (
         not failed_systemd_units
-    ), f"{len(failed_systemd_units)} systemd units failed to load"
+    ), f"{failed_systemd_units} systemd units failed to load"
```
NotTheEvilOne: Readability could be improved with: f"{len(failed_systemd_units)} systemd units failed to load: {failed_systemd_units!r}"

Above change would create output similar to `["network-online.service", "ceph.service"] systemd units failed to load` while the suggested one `2 systemd units failed to load: ["network-online.service", "ceph.service"]`.

tests-ng/test_basics.py
vivus-ignis: `!r` makes no difference when printing a list of strings. "Better" is very subjective, but getting rid of unnecessary complexity (e.g. not using noop syntax like `!r`) is approachable.

vivus-ignis: Reverted the changes to the file.

tests-ng/test_basics.py
vivus-ignis: Changes reverted.

tests-ng/test_basics.py
vivus-ignis: Changes reverted.

Issue #3878, opened 2025-11-13
FIPS: Disable `apt` workaround


Akendo:
Presently, `apt` relies on `http` to fetch its packages. Since we deployed the fix for `apt` already in #3626, there is no need for it.

PR #3879, opened 2025-11-13
Disable the apt workaround for FIPS

Akendo:
In this PR, we disable the workaround for `apt`. 

Since the code that broke during the usage of `apt` has been dealt with, we can use the mirror with HTTPS again. 

Fixes: #3878


moritzkirchnersap (approved):
LVGTM

Issue #3898, opened 2025-11-17
Advance Tests for FIPS


Akendo:
We need to test more advanced things for FIPS. 

- ~~Test that the correct HMAC is present in our GnuTLS package~~ 
- Ensure that we have HTTPS enabled on the mirror.
- Ensure that we can download the package from the mirror with HTTPS enabled. 
- Check that we can create a connection via OpenSSL to an external service.
- Check that we can create a connection via GnuTLS to an external service.
- ~~Check that libgcrypt is in FIPS mode~~
- ~~Check that GnuTLS is in FIPS mode~~
- ~~Check that OpenSSL is in FIPS mode~~
- Check that OpenSSH has the correct Cipher/KEK installed.

PR #3899, opened 2025-11-17
Add more tests for GnuTLS

Akendo:
In this PR, we add some more tests to ensure that the GnuTLS implementation we're using works as expected.

Relates: #3898

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -35,7 +35,7 @@ jobs:
     permissions:
       id-token: write # TODO: 3215 elevated permissions for included workflow
       actions: write  # TODO: 3215 elevated permissions for included workflow
-      packages: write # TODO: 3215 elevated permissions for included workflow    
+      packages: write # TODO: 3215 elevated permissions for included workflow
   test:
     needs: build
     if: ${{ success() || failure() }}
--- a/tests-ng/handlers/configure_nvme.py
+++ b/tests-ng/handlers/configure_nvme.py
@@ -115,5 +115,5 @@ def nvme_device(shell: ShellRunner, dpkg: Dpkg, kernel_module: KernelModule):
         if entry["status"] == "Loaded":
             kernel_module.unload_module(name)
             entry["status"] = None
-    if mount_package_installed == True:
+    if mount_package_installed:
         shell("DEBIAN_FRONTEND=noninteractive apt remove mount")
--- a/tests-ng/handlers/iscsi.py
+++ b/tests-ng/handlers/iscsi.py
@@ -1,7 +1,5 @@
 import os
 import re
-from doctest import FAIL_FAST
-from math import fabs
 
 import pytest
 from plugins.kernel_module import KernelModule
--- a/tests-ng/plugins/kernel_module.py
+++ b/tests-ng/plugins/kernel_module.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import re
 from dataclasses import dataclass
@@ -10,6 +11,8 @@ from .kernel_versions import KernelVersions
 from .shell import ShellRunner
 
 dependencies = re.compile("/([^/]*)\\.ko")
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass
@@ -55,7 +58,9 @@ class KernelModule:
         return result.returncode == 0
 
     def safe_load_module(self, module: str) -> bool:
-        """Load ``module`` using ``modprobe`` and save all modules that have to be unloaded to revert the change; return True on success or if the module is already loaded."""
+        """Load ``module`` using ``modprobe`` and save all modules that have to be unloaded
+        to revert the change; return True on success or if the module is already loaded.
+        """
         if not self.is_module_loaded(module):
             self._update_module_dependencies(module)
             return self.load_module(module)
@@ -63,16 +68,36 @@ class KernelModule:
 
     def unload_module(self, module: str) -> bool:
         """Unload ``module`` using ``rmmod``; return True on success."""
+        logger.debug(f"KernelModule.unload_module: about to call 'rmmod {module}'")
         result = self._shell(
-            f"rmmod {module}", capture_output=True, ignore_exit_code=True
+            f"rmmod {module}",
+            capture_output=True,
+            ignore_exit_code=True,
         )
+        logger.debug(f"rmmod stdout: <<{result.stdout}>>")
+        logger.debug(f"rmmod stderr: <<{result.stderr}>>")
+        return result.returncode == 0
+
+    def safe_unload_module(self, module: str) -> bool:
+        """Unload ``module`` using ``modprobe``; return True on success."""
+        logger.debug(
+            f"KernelModule.safe_unload_module: about to call 'modprobe -r {module}'"
+        )
+        result = self._shell(
+            f"modprobe --verbose --remove --wait 60000 {module}",
+            capture_output=True,
+            ignore_exit_code=True,
+        )
+        logger.debug(f"modprobe stdout: <<{result.stdout}>>")
+        logger.debug(f"modprobe stderr: <<{result.stderr}>>")
         return result.returncode == 0
 
     def safe_unload_modules(self) -> bool:
-        """Unload all modules and dependecies loaded by ``safe_load_module`` in the correct order using ``rmmod``; return True if all succeed"""
+        """Unload all modules and dependecies loaded by ``safe_load_module`` in the correct order
+        using ``modprobe``; return True if all succeed"""
         success = True
         for module in self._unload.static_order():
-            success &= self.unload_module(module)
+            success &= self.safe_unload_module(module)
 
         self._unload = TopologicalSorter()
         return success
--- a/tests-ng/test_sysdiff.py
+++ b/tests-ng/test_sysdiff.py
@@ -1,7 +1,9 @@
-import sys
+import logging
 
 import pytest
 from plugins.sysdiff import Sysdiff
+
+logger = logging.getLogger(__name__)
 
 before_suffix = "before-tests"
 after_suffix = "after-tests"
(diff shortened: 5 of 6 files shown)


tests-ng/test_fips.py
```diff
@@ -28,6 +35,62 @@ def test_gnutls_fips_file_is_empty():
     assert gnutls_fips_file.st_size == 0, f"The /etc/system-fips is not empty."
 
 
+@pytest.mark.feature("_fips")
+def test_gnutls_fips_dot_hmac_file_is_presented():
+    """
+    GnuTLS will perform based on the FIPS requierment a self check. This requieres to have a
+    file that contains a HMAC presented on the target system. This test ensure that the file
+    was installed.
```
```suggestion
    GnuTLS will perform a self check based on the FIPS requirements. A file that contains an HMAC needs to be present on the target system. This test ensures that the file was installed.
```

moritzkirchnersap: (sorry, the line breaks broke during the suggestion edit for whatever reason)

tests-ng/test_fips.py
```diff
@@ -28,6 +35,62 @@ def test_gnutls_fips_file_is_empty():
...
+        gnutls_fipshmac_file_state
+    ), f"The f{gnutls_fipshmac_path} file does not exist."
+
+
+@pytest.mark.feature("_fips")
+def test_gnutls_fips_dot_hmac_file_is_vaild():
+    """
+    Key problem in shipping GnuTLS packages was that the computed HMAC was not vailded for the ship
+    version of the library since the HMAC was computed for an unstripped version. This test ensures
+    that the HMAC on the system fits with the shipped one.
```
```suggestion
    One problem in shipping GnuTLS packages was that the computed HMAC was not valid for the shipped
    version of the library since the HMAC was computed for an unstripped version. This test ensures
    that the HMAC on the system fits with the shipped library version.
```

tests-ng/test_fips.py
```diff
@@ -28,6 +35,62 @@ def test_gnutls_fips_file_is_empty():
...
+    ARCH = arch()
+    gnutls_fipshmac_path = f"/usr/lib/{ARCH}-linux-gnu/.libgnutls.so.30.hmac"
+    gnutls_lib_path = f"/usr/lib/{ARCH}-linux-gnu/libgnutls.so.30"
+    # https://gitlab.com/gnutls/gnutls/-/blob/master/configure.ac?ref_type=heads#L677
+    SECRET = "orboDeJITITejsirpADONivirpUkvarP"
+
+    config = configparser.ConfigParser()
+    config.read("/usr/lib/aarch64-linux-gnu/.libgnutls.so.30.hmac")
+
+    fipshmac = hmac.new(key=SECRET.encode("UTF-8"), msg=None, digestmod=SHA256)
```
```suggestion
    fips_hmac = hmac.new(key=SECRET.encode("UTF-8"), msg=None, digestmod=SHA256)
```

tests-ng/test_fips.py
```diff
@@ -28,6 +35,62 @@ def test_gnutls_fips_file_is_empty():
...
+    # Need to read it 'b' since it's a binary file.
+    with open(gnutls_lib_path, mode="rb") as lib:
+        fipshmac.update(lib.read())
```
```suggestion
        fips_hmac.update(lib.read())
```

tests-ng/test_fips.py
```diff
@@ -28,6 +35,62 @@ def test_gnutls_fips_file_is_empty():
...
+
+    assert (
+        config["libgnutls.so.30"]["hmac"] == fipshmac.hexdigest()
```
```suggestion
        config["libgnutls.so.30"]["hmac"] == fips_hmac.hexdigest()
```

tests-ng/test_fips.py
```diff
@@ -28,6 +35,62 @@ def test_gnutls_fips_file_is_empty():
...
+    ), "Compute HMAC is incorrect!"
+
+
+@pytest.mark.feature("_fips")
+def test_gnutls_fips_dot_hmac_file_contains_vaild_hmac():
+    """
+    Test that the computed HMAC that was shipped is correct. Else the selftest will fail at the
+    secound
+    """
+    gnutls_fipshmac_path = f"/usr/lib/{arch()}-linux-gnu/.libgnutls.so.30.hmac"
```
```suggestion
    gnutls_fips_hmac_path = f"/usr/lib/{arch()}-linux-gnu/.libgnutls.so.30.hmac"
```

tests-ng/test_fips.py
```diff
@@ -28,6 +35,62 @@ def test_gnutls_fips_file_is_empty():
...
+    with open(gnutls_lib_path, mode="rb") as lib:
+        fipshmac.update(lib.read())
+
+    assert (
+        config["libgnutls.so.30"]["hmac"] == fipshmac.hexdigest()
+    ), "Compute HMAC is incorrect!"
+
+
+@pytest.mark.feature("_fips")
+def test_gnutls_fips_dot_hmac_file_contains_vaild_hmac():
```
moritzkirchnersap: Is there an assert missing in this test?

tests-ng/test_fips.py
```diff
@@ -28,6 +35,62 @@ def test_gnutls_fips_file_is_empty():
...
+    """
+    Test that the computed HMAC that was shipped is correct. Else the selftest will fail at the
+    secound
```
```suggestion
    Test that the computed HMAC that was shipped is correct. Else the selftest will fail at the
    second
```

moritzkirchnersap: Second what?

tests-ng/test_fips.py
```diff
@@ -28,6 +35,62 @@ def test_gnutls_fips_file_is_empty():
...
+    """
+    GnuTLS will perform based on the FIPS requierment a self check. This requieres to have a
+    file that contains a HMAC presented on the target system. This test ensure that the file
+    was installed.
+
+    https://www.gnutls.org/manual/html_node/FIPS140_002d2-mode.html
+    """
+    gnutls_fipshmac_path = f"/usr/lib/{arch()}-linux-gnu/.libgnutls.so.30.hmac"
+    gnutls_fipshmac_file = os.stat(gnutls_fipshmac_path)
+    gnutls_fipshmac_file_state = S_ISREG(gnutls_fipshmac_file.st_mode)
```
fwilhe: I am wondering if we should abstract this a bit. I had to google what `S_ISREG` means, and since we wanted the tests to be readable by people with less technical background than us, a helper method could be useful here.

Maybe something like `is_a_file(path)` or `is_a_directory(path)`.. not sure how much the differentiation between regular files and other non-directory files matters

tests-ng/test_fips.py
Akendo: It does, I refactored the code and left it in accidentally.

tests-ng/test_fips.py
yeoldegrove: @Akendo already suggested such a plugin recently. I just opened https://github.com/gardenlinux/gardenlinux/issues/3905 for that.

These come to my mind and could be added to a `tests-ng/plugins/file.py`:
- is_file(path)
- is_dir(path)
- has_mode(path, 0755)

tests-ng/test_fips.py
```diff
@@ -28,6 +35,54 @@ def test_gnutls_fips_file_is_empty():
...
+    version of the library since the HMAC was computed for an unstripped version. This test ensures
+    that the HMAC on the system fits with the shipped library version.
+
+    Without the HMAC the selftest will fail at second part with the follow error message:
+
+    |<1>| FIPS140-2 self testing part 2 failed
+
+    """
+    ARCH = arch()
+    gnutls_fipshmac_path = f"/usr/lib/{ARCH}-linux-gnu/.libgnutls.so.30.hmac"
```
moritzkirchnersap: I still believe fips_hmac is way more readable than fipshmac - that looks almost like a fishmac 🍔 
```suggestion
    gnutls_fips_hmac_path = f"/usr/lib/{ARCH}-linux-gnu/.libgnutls.so.30.hmac"
```

tests-ng/test_fips.py
Akendo: Thanks, going to check it out. @fwilhe `S_ISREG` is part of the [POSIX](https://man.archlinux.org/man/core/man-pages/S_ISREG.3.en#S_ISREG) standard, `s_` := stats, `ISREG` := is regular file. Was just me try to be quick, but you're correct.

tests-ng/test_fips.py
Akendo: You're correct. I've read a bit too much of the original code and kept the keyword in mind. Going to correct it. Thanks.

tests-ng/test_fips.py
Akendo: Since the changes from @yeoldegrove are still in a unmerged PR, I used the suggestion of @fwilhe

Akendo:
@fwilhe, @moritzkirchnersap didn't have the time to fix the issues until now. Feel free to review again.

tests-ng/test_fips.py
```diff
@@ -25,7 +33,50 @@ def test_gnutls_fips_file_is_empty():
...
     gnutls_fips_file = os.stat("/etc/system-fips")
 
-    assert gnutls_fips_file.st_size == 0, f"The /etc/system-fips is not empty."
+    assert gnutls_fips_file.st_size == 0, f"The /etc/system-fips is not empty!"
+
+
+@pytest.mark.feature("_fips")
+def test_gnutls_fips_dot_hmac_file_is_presented():
+    """
+    GnuTLS will perform a self check based on the FIPS requirements. A file that contains an HMAC needs to be present on the target system. This test ensures that the file was installed.
```
```suggestion
    GnuTLS will perform a self check based on the FIPS requirements. A file that
    contains an HMAC needs to be present on the target system. This test ensures
    that the file was installed.
```

tests-ng/test_fips.py
```suggestion
            kernel_cmd_file.read().strip() == 'CMDLINE_LINUX="$CMDLINE_LINUX fips=1"'
```

Only very minor but I personally wouldn't match newlines

PR #3913, opened 2025-11-20
add architecture decision for providing Garden Linux images for Lima-VM via containerized YAML generator

fwilhe:
Based on a discussion with @SujanaSubr and @yeoldegrove

### DIFF
--- a/docs/architecture/decisions/0023-lima-image-download.md
+++ b/docs/architecture/decisions/0023-lima-image-download.md
@@ -8,7 +8,7 @@ Accepted
 
 ## Context
 
-We want to provide Garden Linux images for Lima-VM in a convenient and user-friendly way. Currently, our pipeline publishes images to S3 buckets without simple or clean URLs, requiring a wrapper to retrieve the image URL. The existing `start-lima.py` script introduces additional dependencies on Python packages, Podman, and a local clone of the Garden Linux git repository. Our goal is to make this a proper feature in the pipeline, avoiding special case handling for Lima-VM where possible.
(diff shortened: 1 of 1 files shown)


docs/architecture/decisions/0023-lima-image-download.md
```diff
@@ -0,0 +1,44 @@
...
+
+Date: 2025-11-20
+
+## Status
+
+Accepted
+
+## Context
+
+We want to provide Garden Linux images for Lima-VM in a convenient and user-friendly way. Currently, our pipeline publishes images to S3 buckets without simple or clean URLs, requiring a wrapper to retrieve the image URL. The existing `start-lima.py` script introduces additional dependencies on Python packages, Podman, and a local clone of the Garden Linux git repository. Our goal is to make this a proper feature in the pipeline, avoiding special case handling for Lima-VM where possible.
```
yeoldegrove: Nitpick: I would not talk about an existing `start-lima.py` script as this is not yet part of `main` but in a PR.
Maybe we can say:
"We want a user, that just wants to run a Lima-VM, to install minimal dependencies. This does exclude a python interpreter and packages or a local clone of the Garden Linux git repository."

yeoldegrove (approved):
One optional nitpick, else good from my side.

docs/architecture/decisions/0023-lima-image-download.md
fwilhe: For this statement to make sense, I think it is needed to refer to this draft implementation.

SujanaSubr: Agree with @yeoldegrove , Also this file name will be changed after our discussion

fwilhe: I've changed it to mention the script was proposed. If you feel this is not right, please provide an alternative suggestion, not sure how to phrase it in a way that does not sound strange.

PR #3695, opened 2025-10-21
feat: add container-python-dev for multi-stage builds

Leon-hk:
**What this PR does / why we need it**:
- Adds bare-python-dev image which includes python, pip, bash, all shared libraries and a few required binaries
- The bare-python-dev can be used to develop python containers which require packages from pip
- It can also be used in a multi-stage build to copy the additional dependencies to the bare-python image, keeping it as small as possible

**Which issue(s) this PR fixes**:
Fixes #3479, #3480, #3405 

**Special notes for your reviewer**:
Reasons for choosing a multi-stage build instead of adding pip directly to bare-python:
- adding pip is quite heavy (+11.6 MB) and is not required after the build process
- if pip is required in the container one can use the bare-python-dev image
- pip is not enough, many packages additionally require shared libraries
  - choosing a perfect set of shared libraries is impossible, a too small set would miss some libraries, while a too large set ships unused files
- With good documentation, a multi-stage build is easy to use

How it works: _(Updated)_
1. `requirements.txt` is copied to the container-pythondev image
2. Install the packages with pip
3. Execute the `export_libs.py` script, which is already contained in the container-pythondev image. This will look for all ldd dependencies of the installed python packages and copies them into the `required_libs` folder
4. The second image is based on bare-python
5. Copy the `dist-packages` folder from the first image
6. Copy the `required_libs` from the first image to `/`
7. Copy the python files
8. Execute the python files with `ENTRYPOINT`

How to test: _(Updated)_
As there is no published container-pythonDev image yet, it has to be build by yourself.
1. `./build container-pythonDev`
2. `podman tag $(podman load -q --input .build/container-pythonDev-arm64-today-local.oci | cut -d ":" -f3) container-pythondev:latest`
3. Replace `ghcr.io/gardenlinux/gardenlinux/container-pythondev:1877.5` by `container-pythondev:latest` in the [Dockerfile](https://github.com/gardenlinux/gardenlinux/blob/feat/bare-python-dev/docs/01_developers/bare_container.md#adding-dependencies)

PR #3914, opened 2025-11-20
Enhance and test `exportLibs.py`

### LINKED ISSUE #3889: feat: use python-gardenlinux-lib for exporting shared libraries from container-python-dev
This is a follow up of #3695 

- Currently, the `container-python-dev` flavor is using an explicitly written `exportLibs.py` script to extract the necessary share libraries
- We already have the python-gardenlinux-lib bundling a lot of gardenlinux developer features, which can be installed to the `pythonDev` feature, as it already includes python, pip and venv packages
- PR for python-gardenlinux-lib [#253](https://github.com/gardenlinux/python-gardenlinux-lib/pull/253)
- We might also change the hardcoded `python3.13-venv` to `python-all-venv`

**Why is this needed**:
- Enhancing the structure and allow reuse of the feature

Leon-hk:
This is a follow up of #3695
Depends on merge of python-gardenlinux-lib [#253](https://github.com/gardenlinux/python-gardenlinux-lib/pull/253)

- Removes `exportLibs.py` script from `pythonDev` feature
- Installs python-gardenlinux-lib on `pythonDev` feature
- Adds feature tests to check if exporting the shared libraries for the `requests` package works as expected

Fixes #3889


docs/01_developers/bare_container.md
```diff
@@ -45,13 +45,13 @@ Their development is streamlined by the [unbase_oci](https://github.com/gardenli
 - **Description**: **bare-python** does not contain pip and has a limited set of shared libraries. To add dependencies, we recommend using a multi-stage build with the **container-pythonDev** flavor
 - **Multi-stage Dockerfile**:
   ```Dockerfile
-  FROM ghcr.io/gardenlinux/gardenlinux/container-python-dev:1877.5 as packages
+  FROM ghcr.io/gardenlinux/gardenlinux/container-python-dev:1877.8 as packages
```
fwilhe: this image is not published yet?

docs/01_developers/bare_container.md
```diff
@@ -45,13 +45,13 @@ Their development is streamlined by the [unbase_oci](https://github.com/gardenli
...
 
   COPY requirements.txt /
   RUN pip3 install -r requirements.txt --break-system-packages --no-cache-dir
-  RUN exportLibs.py
+  RUN gl-build export-python-libs
```
fwilhe: how does `gl-build` end up in that image? this is a script from the gl-python-lib, but in the pythonDev feature this library is not installed (not suggesting it should), but then an extra step would be required in this dockerfile, right?

fwilhe: nevermind, just noticed it is part of this pr

docs/01_developers/bare_container.md
Leon-hk: I think the first PR was merged after the release of 1877.7. And as this draft would also change the interface from `RUN exportLibs.py` to `RUN gl-build export-python-libs`, I've changed the version to a release which would probably contain the updated version, to have a working example

Leon-hk:
I've restored an updated version of the `exportLibs.py` script, if we decide to use the `python-gardenlinux-lib` we can rollback

features/pythonDev/exec.config
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+cd root 
+mkdir venv
+python3 -m venv venv
+
+# shellcheck source=/dev/null
+source venv/bin/activate
+pip install pyelftools
```
fwilhe, line 12: what's the rationale for this installation method? did you evaluate packaging this as a debian package? not sure if this is what we want, please get feedback from @nkraetzschmar

Leon-hk: The `exportLibs.py` script runs in a virtual environment, as it requires `pyelftools`.  This is necessary to have a completely fresh python environment of the system, as the `dist-packages` are later copied to the `bare-python` image. If we would install `pyelftools` system wide, it would be always contained in the `bare-python` image, even if it is not used.

fwilhe: that seems reasonable, thanks for explaining, but I think this is obscure enough that it should be documented (a comment in the script is fine imo)

Leon-hk: Sure, makes sense 👍 I've added a comment.

PR #3881, opened 2025-11-13
test-ng: Improve usability of parse.py plugin for structured and unstructured formats

yeoldegrove:
**What this PR does / why we need it**:

This PR refactors the parsing plugins to improve API usability, clarity, and maintainability.

Key improvements:

- **API Redesign**:

  - Renamed `FileContent` to `ParseFile` ~~and `ValidationResult` to `MatchResult` for better clarity~~.
  - ~~Method names improved (`line()` → `has_line()`, `lines()` → `has_lines()`).~~

  - **Line-based operations with `__contains__` support**:

    - `lines(format=None, comment_char=None, ignore_missing=False, ordered=False)` - Returns a `Lines` container that supports the `in` operator for Pythonic pattern checking. Automatically handles comment filtering based on format. Supports:
      - **String literals**: `assert "pattern" in parse_file.lines(path)` - Check if a line containing the pattern exists, with automatic comment filtering. Supports format-specific comment characters (e.g., INI supports both `;` and `#`).
      - **Regex patterns**: `assert re.compile(r"\d+") in parse_file.lines(path)` - Apply regex patterns to content for flexible matching.
      - **Multiple patterns (unordered)**: `assert ["pattern1", "pattern2"] in parse_file.lines(path)` - Check if multiple patterns exist anywhere in the file.
      - **Multiple patterns (ordered)**: `assert ["pattern1", "pattern2"] in parse_file.lines(path, ordered=True)` - Check if patterns appear in sequence.
    - [Example](https://github.com/gardenlinux/gardenlinux/blob/feat/parser_plugins_refactor/tests-ng/plugins/tests/test_parse_file.py#L16-L25) [Docs](https://github.com/gardenlinux/gardenlinux/blob/feat/parser_plugins_refactor/tests-ng/plugins/parse_file.py#L191-L228)

    ~~- `has_line(pattern)` - Check if a line containing the pattern exists in content/files, with automatic comment filtering. Supports format-specific comment characters (e.g., INI supports both `;` and `#`). [Example](https://github.com/gardenlinux/gardenlinux/blob/feat/parser_plugins_refactor/tests-ng/plugins/tests/test_parse.py#L15-L18) [Docs](https://github.com/gardenlinux/gardenlinux/blob/feat/parser_plugins_refactor/tests-ng/plugins/parse.py#L200-L224)~~
[...]

PR #3916, opened 2025-11-20
test-ng: add file.py for simple file tests 

yeoldegrove:

This PR adds a `file()` fixture that can check various file metadata:

##### File Type Checks

- `exists(path)` -> bool: Check if path exists (any type)
- `is_regular_file(path)` -> bool: Check if path is a regular file (not symlinks)
- `is_dir(path)` -> bool: Check if path is a directory
- `is_symlink(path, target=None, error_on_broken_symlink=True)` -> bool: Check if path is a symlink, optionally validating target

##### Permission Check

- `is_executable(path)` -> bool: Check if file is executable
- `is_readable(path)` -> bool: Check if file is readable
- `is_writable(path)` -> bool: Check if file is writable

##### Permission Metadata

- `has_mode(path, mode: str)` -> bool: Check if file has specific permissions (e.g., "0755")
- `get_mode(path)` -> str: Get file permissions as octal string (e.g., "0755")

##### Ownership Checks

- `is_owned_by_user(path, user: str)` -> bool: Check if file is owned by specific user
- `is_owned_by_group(path, group: str)` -> bool: Check if file is owned by specific group
- `is_owned_by(path, user: str, group: str)` -> bool: Check if file is owned by specific user and group

##### Ownership Metadata

- `get_owner(path)` -> tuple[str, str]: Get (user, group) ownership
- `get_user(path)` -> str: Get file owner username
- `get_group(path)` -> str: Get file owner group name

##### Other

- `get_size(path)` -> int: Get file size in bytes

##### Tests for plugins

- To discuss if needed or overkill:

Same as in #3881...
- **Plugin Testing**: Added 61 tests covering all methods, formats, edge cases, and error conditions for the `file.py`  plugin.
  - Idea: `dev_tests.yml` workflow or a developer runs `make -f tests-ng/dev.makefile test-plugins` to run integration tests of plugins. `tests-ng/plugins/tests/test_*.py` can contain tests that are run automatically after each push or locally.

Fixes #3904

### DIFF
--- a/tests-ng/test_basics.py
+++ b/tests-ng/test_basics.py
@@ -1,6 +1,7 @@
 import os
 
 import pytest
+from plugins.file import File
 from plugins.shell import ShellRunner
 from plugins.systemd import Systemd
 
@@ -19,32 +20,54 @@ def test_no_man(shell: ShellRunner):
     ), "man ls, did not fail with 'not found' as expected"
 
 
-def test_fhs(shell: ShellRunner):
-    expected_dirs = {
-        "bin",
-        "boot",
-        "dev",
-        "etc",
-        "home",
-        "lib",
-        "mnt",
-        "opt",
-        "proc",
-        "root",
-        "run",
-        "sbin",
-        "srv",
-        "sys",
-        "tmp",
-        "usr",
-        "var",
-    }
+@pytest.mark.parametrize(
+    "dir",
+    [
+        "/boot",
+        "/dev",
+        "/etc",
+        "/home",
+        "/mnt",
+        "/opt",
+        "/proc",
+        "/root",
+        "/run",
+        "/srv",
+        "/sys",
+        "/tmp",
+        "/usr",
+        "/var",
+    ],
+)
+def test_fhs_directories(file: File, dir: str):
+    assert file.is_dir(f"{dir}"), f"expected FHS directory {dir} does not exist"
 
-    if os.uname().machine == "x86_64":
-        expected_dirs.add("lib64")
 
-    for dir in sorted(expected_dirs):
-        assert os.path.isdir(f"/{dir}"), f"expected FHS directory /{dir} does not exist"
+@pytest.mark.parametrize(
+    "link,target",
+    [
+        ("/bin", "/usr/bin"),
+        ("/lib", "/usr/lib"),
+        ("/sbin", "/usr/sbin"),
+    ],
+)
+def test_fhs_symlinks(file: File, link: str, target: str):
+    assert file.is_symlink(
+        link, target=target
+    ), f"expected FHS symlink {link} does not exist"
+
+
+@pytest.mark.arch("amd64", reason="links only present on amd64")
+@pytest.mark.parametrize(
+    "link,target",
+    [
+        ("/lib64", "/usr/lib64"),
+    ],
+)
+def test_fhs_symlinks_amd64(file: File, link: str, target: str):
+    assert file.is_symlink(
+        link, target=target
+    ), f"expected FHS symlink {link} does not exist"
 
 
 @pytest.mark.booted(
--- a/tests-ng/test_fips.py
+++ b/tests-ng/test_fips.py
@@ -9,31 +9,28 @@ from platform import machine as arch
 from typing import List
 
 import pytest
+from plugins.file import File
 from plugins.kernel_cmdline import kernel_cmdline
 
 
 @pytest.mark.feature("_fips")
-def test_gnutls_fips_file_was_created():
+def test_gnutls_fips_file_was_created(file: File):
     """
     GnuTLS requires to have the /etc/system-fips to be present as prerequisite
     to enable the FIPS mode.
 
     https://www.gnutls.org/manual/html_node/FIPS140_002d2-mode.html
     """
-    gnutls_fips_file = "/etc/system-fips"
-    assert os.path.isfile(
-        gnutls_fips_file
-    ), f"The {gnutls_fips_file} file does not exist."
+
+    assert file.is_regular_file("/etc/system-fips")
 
 
 @pytest.mark.feature("_fips")
-def test_gnutls_fips_file_is_empty():
+def test_gnutls_fips_file_is_empty(file: File):
     """
     The /etc/system-fips should be without any content.
     """
-    gnutls_fips_file = os.stat("/etc/system-fips")
-
-    assert gnutls_fips_file.st_size == 0, f"The /etc/system-fips is not empty!"
+    assert file.get_size("/etc/system-fips") == 0, f"The /etc/system-fips is not empty."
 
 
 @pytest.mark.feature("_fips")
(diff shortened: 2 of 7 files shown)


tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+
+        Args:
+            path: File path.
+
+        Returns:
+            bool: True if the path exists, False otherwise.
+        """
+        return Path(path).exists()
+
+    def is_file(self, path: str | Path) -> bool:
```
NotTheEvilOne: The method name is misleading. Looking at `pathlib.Path.is_file()` as a similar one would imply that this function returns true as long as a path exists and is not a directory. As the description and code indicates this is true for regular files instead only. Maybe you find a better name for the method?

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+
+        Args:
+            path: File path.
+
+        Returns:
+            bool: True if the path is a directory, False otherwise.
+        """
+        p = Path(path)
+        try:
+            return stat.S_ISDIR(p.lstat().st_mode)
```
NotTheEvilOne: `Path.is_dir(follow_symlinks=False)` should return the same result. Please note that `stat()` is usually more expensive (time consuming / memory wise).

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+
+        Raises:
+            FileNotFoundError: If error_on_broken_symlink is True and the symlink
+                target does not exist.
+        """
+        p = Path(path)
+        if not p.is_symlink():
+            return False
+
+        if target is not None:
```
NotTheEvilOne:
```
if target is None:
  return True
```

Would make the code more readable here :)

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+            symlink_target = p.readlink()
+            target_path = Path(target)
```
NotTheEvilOne: I think `pathlib.Path.samefile()` should give you the same logic here. Would be cool if one line could replace the rest of the method :)

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+        """
+        stat_info = Path(
+            path
+        ).stat()  # Raises FileNotFoundError or PermissionError naturally
+        mode = stat_info.st_mode
+        current_uid = os.getuid()
+        current_gids = {os.getgid()} | set(os.getgroups())
+
+        # Map permission types to (owner, group, other) bit masks
+        permission_bits = {
```
NotTheEvilOne: This could be constants (e.g. top of file) removing the need to map memory for each call to the method.

Not sure if something like a contant like PERMISSIONS_READ and a generic approach in this method to what is actually checked would be more suitable than a string like "read", "write" or "execute". Could be octets given as inputs like 0x700 etc.

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+            path: File path.
+
+        Returns:
+            bool: True if the file is executable, False otherwise.
+
+        Raises:
+            FileNotFoundError: If the path does not exist.
+            PermissionError: If permission to access the path is denied.
+        """
+        return self._check_permission(path, "execute")
```
NotTheEvilOne: `os.access` with `os.X_OK` should be used here as for portability and other reasons I guess more than just file permissions are relevant to return a correct result if a path is executable or not.

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+            path: File path.
+
+        Returns:
+            bool: True if the file is writable, False otherwise.
+
+        Raises:
+            FileNotFoundError: If the path does not exist.
+            PermissionError: If permission to access the path is denied.
+        """
+        return self._check_permission(path, "write")
```
NotTheEvilOne: Same as for `is_executable()`.

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+            path: File path.
+
+        Returns:
+            bool: True if the file is readable, False otherwise.
+
+        Raises:
+            FileNotFoundError: If the path does not exist.
+            PermissionError: If permission to access the path is denied.
+        """
+        return self._check_permission(path, "read")
```
NotTheEvilOne: Same as for `is_executable()`.

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+
+        Returns:
+            Tuple[str, str]: A tuple of (username, groupname).
+
+        Raises:
+            FileNotFoundError: If the path does not exist.
+            PermissionError: If permission to access the path is denied.
+        """
+        stat_info = Path(path).stat()
+        user = pwd.getpwuid(stat_info.st_uid).pw_name
```
NotTheEvilOne: Should call `self.get_user()` / `self.get_group()` for easier maintainability. This way only these methods need to be changed if needed.

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+            str: Group name of the file owner.
+
+        Raises:
+            FileNotFoundError: If the path does not exist.
+            PermissionError: If permission to access the path is denied.
+        """
+        stat_info = Path(path).stat()
+        return grp.getgrgid(stat_info.st_gid).gr_name
+
+    def has_user(self, path: str | Path, user: str) -> bool:
```
NotTheEvilOne: This should maybe be called `is_owned_by_user()`.

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+        Returns:
+            bool: True if the file has the specified user, False otherwise.
+
+        Raises:
+            FileNotFoundError: If the path does not exist.
+            PermissionError: If permission to access the path is denied.
+        """
+        return self.get_user(path) == user
+
+    def has_group(self, path: str | Path, group: str) -> bool:
```
NotTheEvilOne: This should maybe be called `is_owned_by_group()`.

tests-ng/plugins/file.py
```diff
@@ -0,0 +1,368 @@
...
+        Returns:
+            bool: True if the file has the specified group, False otherwise.
+
+        Raises:
+            FileNotFoundError: If the path does not exist.
+            PermissionError: If permission to access the path is denied.
+        """
+        return self.get_group(path) == group
+
+    def has_owner(self, path: str | Path, user: str, group: str) -> bool:
```
NotTheEvilOne: This should maybe be called `is_owned_by()` accepting user and group given.

tests-ng/plugins/file.py
yeoldegrove: Thanks for noticing this inconsistency. `is_regular_file()` comes to mind. Will change it.

tests-ng/plugins/file.py
yeoldegrove: Nice suggestion, will change it.

tests-ng/plugins/file.py
yeoldegrove: Will move the explicit "is None" a bit up.

yeoldegrove: Reworked 'is_symlink()' for better readability instead. Even `if target is None` we need to check for broken symlinks.

tests-ng/plugins/file.py
yeoldegrove: We need some logic to find broken symlinks and non-matching targets and `pathlib.Path.samefile()` won't cover this automatically.
Nevertheless I reworked `is_symlink()` for better readability.

tests-ng/plugins/file.py
yeoldegrove: Removed `_check_permissions()` completely.

tests-ng/plugins/file.py
yeoldegrove: Great hint. This made the whole `_check_permissions()` function obsolete.

tests-ng/plugins/file.py
yeoldegrove: :+1:

tests-ng/plugins/file.py
yeoldegrove: :+1:

tests-ng/plugins/file.py
yeoldegrove: :+1:

tests-ng/plugins/file.py
yeoldegrove: Yes, the renaming helps understanding what the function does. It is a bit long but this can be sacrificed for better clarity, I guess.

tests-ng/plugins/file.py
yeoldegrove: Yes, the renaming helps understanding what the function does. It is a bit long but this can be sacrificed for better clarity, I guess.

tests-ng/plugins/file.py
yeoldegrove: Yes, the renaming helps understanding what the function does. It is a bit long but this can be sacrificed for better clarity, I guess.

yeoldegrove:
@NotTheEvilOne Adressed all your comments, please re-review.

.github/workflows/dev_test-ng.yml
NotTheEvilOne, line 1: Would it make sense to rename this workflow to `test_*.yml` as reuse in other workflows would have consistent naming this way.

yeoldegrove: @NotTheEvilOne I wanted to destuinguish it from `test_*.yml` as this is not about running tests on any of our artifacts or reusing it in other workflows but suits a development purpose.

Maybe re-naming it `dev_tests.yml` and combining it with `lint_tests-ng.yml` would be less ugly. Those two serve a similar purpose.

PR #3929, opened 2025-11-24
Check adr numbering

fwilhe:
Introduce a new workflow to ensure ordering of ADRs is correct (no id is reused, no gaps between ids).

Discovered an issue with 'ADR 0013: System State Diffing in tests-ng', fixing this by putting it to the back to avoid having to rename all ADRs that were created since then.

From now on this should not happen with this workflow.

### DIFF
--- a/.github/workflows/check_adr_numbering.yml
+++ b/.github/workflows/check_adr_numbering.yml
@@ -19,7 +19,6 @@ jobs:
       - name: Verify ADRs are numbered in sequence
         run: |
           #!/usr/bin/env python3
-          from __future__ import annotations
 
           import re
           import sys


.github/workflows/check_adr_numbering.yml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ubuntu-slim
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.14"
+      - name: Verify ADRs are numbered in sequence
+        run: |
+          #!/usr/bin/env python3
+          from __future__ import annotations
```
moritzkirchnersap:
```suggestion
```

I believe this is unnecessary. Apart from that: great :)

PR #3932, opened 2025-11-24
Refactor generate-lima-yaml.py to improve error handling and streamline image retrieval logic

fwilhe:


bin/generate-lima-yaml.py
```diff
@@ -32,68 +33,73 @@ def construct_command(version):
...
-        default='nightly',
-        help="By default starts latest nightly build. \
-            Provide specific version for older nightly builds. Default:nightly"
-    )
-
-    parser.add_argument(
-        "--arch",
-        default=platform.machine(),
-        help="Optional: Provide architecture if specific architecture is needed. Default: Platform architecture"
+        default='latest',
```
SujanaSubr, line 98: Why we make latest as default? Currently we do not support lima VM in releases, hence i would suggest to keep nightly as default

bin/generate-lima-yaml.py
```diff
@@ -32,68 +33,73 @@ def construct_command(version):
...
-            check_whether_image_exists = output["releases"][0]["flavors"].get(image_name)
-            if not check_whether_image_exists:
-                print(f"Error: No Lima image found for the specified version.")
+            try:
+                lima_image_for_arch = output["releases"][0]["flavors"][image_name]
+            except KeyError:
+                print(f"The specified version {version} has no lima image available. Try a newer version.", file=sys.stderr)
+                exit(1)
+
+            if not lima_image_for_arch:
```
SujanaSubr, line 67: This is not required. This case will never hit since it will fail at "try"

PR #3945, opened 2025-11-26
Add container image for lima manifest generation

SujanaSubr:
Build containerfile for creating lima manifests

https://github.com/gardenlinux/gardenlinux/blob/main/docs/architecture/decisions/0023-lima-image-download.md

### DIFF
--- a/.github/workflows/build_lima_yaml_container.yml
+++ b/.github/workflows/build_lima_yaml_container.yml
@@ -1,4 +1,4 @@
-name: build_lima_manifest_generation_container
+name: Build Lima Container
 on:
   workflow_dispatch:
   schedule:
--- a/Containerfile.lima-manifest
+++ b/Containerfile.lima-manifest
@@ -5,6 +5,6 @@ LABEL org.opencontainers.image.description="Garden Linux Lima Manifest Generatio
           
 WORKDIR /workspace
 COPY ./bin/generate-lima-yaml.py  /workspace
-RUN apt update && apt install -y python3-pip
+RUN apt-get update && apt-get install -y python3-pip
 RUN pip install pyyaml
-ENTRYPOINT ["./generate-lima-yaml.py"]
+ENTRYPOINT ["/workspace/generate-lima-yaml.py"]


.github/workflows/build_lima_yaml_container.yml
```diff
@@ -0,0 +1,39 @@
...
+          FROM ghcr.io/gardenlinux/glrd:latest
+          WORKDIR /workspace
+          COPY ./bin/generate-lima-yaml.py  /workspace
+          RUN apt update && apt install -y python3-pip
+          RUN pip install pyyaml
+          ENTRYPOINT ["./generate-lima-yaml.py"]
+          EOF
+      - name: Build Container Image
+        run: |
+          podman build -t ghcr.io/${{ github.repository }}/generate_lima_manifest:latest .
```
fwilhe: this container image should be built both for amd64 and for arm64

also, it should be published to ghcr.io

you can find an example for how I'm doing this here https://github.com/gardenlinux/glvd-postgres/blob/main/.github/workflows/build-postgres-container.yml#L63

.github/workflows/build_lima_yaml_container.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          apt update
+          apt install -y podman
+                
+      - name: Create Dockerfile
+        run: |
+          cat << 'EOF' > Dockerfile
```
fwilhe: creating this file here inside the pipeline feels a bit hacky.. I'm okay with that for now, but it is certainly not idea.

we could also create it as a file in the root of the project (which we need to be able to copy the python script from bin) and give it a disticntive and descriptive name such as `Containerfile.lima-manifest` or similar

.github/workflows/build_lima_yaml_container.yml
```diff
@@ -0,0 +1,44 @@
+name: build_lima_manifest_generation_container
```
fwilhe:
```suggestion
name: Build Lima Container
```

Containerfile.lima-manifest
```diff
@@ -0,0 +1,10 @@
+FROM ghcr.io/gardenlinux/glrd:latest
+
+LABEL org.opencontainers.image.source=https://github.com/gardenlinux/gardenlinux
+LABEL org.opencontainers.image.description="Garden Linux Lima Manifest Generation Container"
+          
+WORKDIR /workspace
+COPY ./bin/generate-lima-yaml.py  /workspace
+RUN apt update && apt install -y python3-pip
```
fwilhe: minor: in containerfiles, using `apt-get` is preferred as apt (still?) has no 'stable interface'

Containerfile.lima-manifest
```diff
@@ -0,0 +1,10 @@
...
+RUN pip install pyyaml
+ENTRYPOINT ["./generate-lima-yaml.py"]
```
fwilhe: unsure if using a relative path here is good. it seems to work but I would always have an absolute path in the entrypoint.

PR #3952, opened 2025-11-28
Add architectural decision document for organizing Python code in Garden Linux

fwilhe:
Decision **draft** document based on the discussions we had.

This is my proposal. I know this is not yet agreed on, but is intended to make the discussion more productive.

Please comment, and if you disagree completely with this proposal, please write your own alternative. This will make it easier to make a decision.

### DIFF
--- a/docs/architecture/decisions/00xx-python-gardenlinux-lib.md
+++ b/docs/architecture/decisions/00xx-python-gardenlinux-lib.md
@@ -1,0 +1,38 @@
+# xx. Organizing Python Code for Garden Linux
+
+Date: 2025-11-28
+
+## Status
+
+Proposed
+
+## Context
+
+Historically, Python code for Garden Linux resided in the main repository, often without proper testing or organization. Some scripts were embedded directly in GitHub Actions workflows, while others were placed in the `hack/` directory. Dependency management for this code was inconsistent and lacked a systematic approach.
+
+Recently, a separate repository, `python-gardenlinux-lib`, was introduced to host Python code for various usages. The work done to implement this separation brought benefits such as dedicated tests and linters, but these improvements are not inherently tied to having a separate repository.
+
+Currently, the lib repo contains code used in GitHub Actions workflows and feature-related logic. This introduces an indirection, making debugging and modifying pipeline code more complex, as changes require updates in the lib repo, releasing a new version, and then updating the dependency in the main repo.
+
+Examples for scope creep in the lib:
+
+- [exportLibs.py](https://github.com/gardenlinux/python-gardenlinux-lib/pull/253)
+- [release-notes](https://github.com/gardenlinux/python-gardenlinux-lib/tree/main/src/gardenlinux/github/release_notes)
+
+## Decision
+
+- Retain the `python-gardenlinux-lib` repository, but limit its scope to true library code, such as interfacing with OCI registries, parsing features, and generating cnames.
+  - Exact decisions on individual code files and directories need to be made on a case-by-case basis, considering the following:
+    - Does the code have more than one (potential) consumer?
+    - Does leaving the code in the lib force the pipelines to install the lib?
+- Move pipeline and feature-related Python code back into the main Garden Linux repository.
+- Continue to maintain tests and linters for all Python code, regardless of its location.
+- Introduce rules for where to place python code in the main repo, how to track dependencies and implement tests.
+- Introduce a PR template for the lib to remind of above criteria to critically reflect if the code should go into the library.
+
+## Consequences
+
+- The library repo will be focused on reusable, generic components.
+- Pipeline and feature logic will reside in the main repo, reducing indirection and simplifying debugging and development.
+- Changes to pipeline code can be made directly in the main repo without requiring updates and releases of the lib repo.
+- Testing and code quality standards will be preserved across both repositories.


docs/architecture/decisions/00xx-python-gardenlinux-lib.md
```diff
@@ -0,0 +1,38 @@
...
+- [exportLibs.py](https://github.com/gardenlinux/python-gardenlinux-lib/pull/253)
+- [release-notes](https://github.com/gardenlinux/python-gardenlinux-lib/tree/main/src/gardenlinux/github/release_notes)
+
+## Decision
+
+- Retain the `python-gardenlinux-lib` repository, but limit its scope to true library code, such as interfacing with OCI registries, parsing features, and generating cnames.
+  - Exact decisions on individual code files and directories need to be made on a case-by-case basis, considering the following:
+    - Does the code have more than one (potential) consumer?
+    - Does leaving the code in the lib force the pipelines to install the lib?
+- Move pipeline and feature-related Python code back into the main Garden Linux repository.
```
NotTheEvilOne, line 28: - I would like to suggest to either rename `python-gardenlinux-lib` to `python-gardenlinux` and leave / use the tools introduced there as they are today or
- Introduce an `python-gardenlinux` or `python-gardenlinux-cli` repository to contain these tools.

Rationale: While maintaining code and introducing features related to the `gardenlinux` repository in it as well seems to be appealing at first, it does not really remove the dependency of library code and in this regard versioning and updating dependencies.

Furthermore improving and maintaining code quality is not really possible in the `gardenlinux` repository as specific workflows and tools and configurations are needed and either depend on a second repository (source like `python-gardenlinux-lib`) or need to be copied over. Later would derive the configuration at a later point in time most likely.

While restructuring the GitHub workflows it showed to be improving overall maintenance of GitHub workflow actions as we have been able to access versioned libraries with a known and specific set of functionality needed at that point in time.

The `python-gardenlinux-lib` introduces code and tools used in GitHub workflows but which can often also used locally for specific use cases. This might be for disaster recovery or if certain issues occur which need manual intervention.

Moving the code to other locations (especially "consuming" repositories) would hinder or make this use case more difficult. First one fits all approach would most likely not work in the long run as not everything is cosumed in the `gardenlinux` repository. Second I think maintenance and code quality will suffer from a splitted code base. Last but not least I would see an approach with splitting library and tools code as the most promising compromise.

docs/architecture/decisions/00xx-python-gardenlinux-lib.md
```diff
@@ -0,0 +1,38 @@
...
+Historically, Python code for Garden Linux resided in the main repository, often without proper testing or organization. Some scripts were embedded directly in GitHub Actions workflows, while others were placed in the `hack/` directory. Dependency management for this code was inconsistent and lacked a systematic approach.
+
+Recently, a separate repository, `python-gardenlinux-lib`, was introduced to host Python code for various usages. The work done to implement this separation brought benefits such as dedicated tests and linters, but these improvements are not inherently tied to having a separate repository.
+
+Currently, the lib repo contains code used in GitHub Actions workflows and feature-related logic. This introduces an indirection, making debugging and modifying pipeline code more complex, as changes require updates in the lib repo, releasing a new version, and then updating the dependency in the main repo.
+
+Examples for scope creep in the lib:
+
+- [exportLibs.py](https://github.com/gardenlinux/python-gardenlinux-lib/pull/253)
+- [release-notes](https://github.com/gardenlinux/python-gardenlinux-lib/tree/main/src/gardenlinux/github/release_notes)
```
NotTheEvilOne, line 20: The `release-notes` tool has never been in a stage that would have allowed it to be merged to `python-gardenlinux-lib` as it suffered from poor code quality and bad integration from the day it was introduced. It resulted in various issues during release cycles as far as I know as it has been activated and used before being cleaned up properly.

Question about if it is suitable at all for the library has already been asked in the [original PR](https://github.com/gardenlinux/python-gardenlinux-lib/pull/179#discussion_r2317896850).

docs/architecture/decisions/00xx-python-gardenlinux-lib.md
```diff
@@ -0,0 +1,38 @@
...
+
+Historically, Python code for Garden Linux resided in the main repository, often without proper testing or organization. Some scripts were embedded directly in GitHub Actions workflows, while others were placed in the `hack/` directory. Dependency management for this code was inconsistent and lacked a systematic approach.
+
+Recently, a separate repository, `python-gardenlinux-lib`, was introduced to host Python code for various usages. The work done to implement this separation brought benefits such as dedicated tests and linters, but these improvements are not inherently tied to having a separate repository.
+
+Currently, the lib repo contains code used in GitHub Actions workflows and feature-related logic. This introduces an indirection, making debugging and modifying pipeline code more complex, as changes require updates in the lib repo, releasing a new version, and then updating the dependency in the main repo.
+
+Examples for scope creep in the lib:
+
+- [exportLibs.py](https://github.com/gardenlinux/python-gardenlinux-lib/pull/253)
```
NotTheEvilOne, line 19: Main reason for the `exportLibs.py` PR have been to benefit from the `python-gardenlinux-lib` repository structure and tooling. It allowed to reuse third-party packages instead of reinventing the wheel for a specific use case.

I see this example as a first step towards supporting tools to build and maintain the GardenLinux repository and code as an `gl-build` tool (for example) that allows advanced logic to be implemented in a structured way with code quality and version maintenance in mind.

Of course this and other tools could be splitted from pure library code and live in it's own repository reusing existing code.

docs/architecture/decisions/00xx-python-gardenlinux-lib.md
fwilhe: > While maintaining code and introducing features related to the gardenlinux repository in it as well seems to be appealing at first, it does not really remove the dependency of library code and in this regard versioning and updating dependencies.

Why? I don't understand.

The code was in gardenlinux before without dependency on this lib. Can you elaborate on why this is not feasible?

NotTheEvilOne: Sure. The code used in the GitHub pipelines have enhanced significantly since the days it only parsed a single input in a regular expression. The `python-gardenlinux-lib` supports release files to separate actions from the originating `features` directory these days, improves OCI and S3 upload handling, provides classes to handle GardenLinux canonical names in a unified way. It support APT and Git repositories, `flavors.yaml` parsing and conversation to JSON and GitHub API calls.

Everything with unit tests (at least intended) and in a code quality checked way for better maintainability.

docs/architecture/decisions/00xx-python-gardenlinux-lib.md
```diff
@@ -0,0 +1,38 @@
...
+
+Proposed
+
+## Context
+
+Historically, Python code for Garden Linux resided in the main repository, often without proper testing or organization. Some scripts were embedded directly in GitHub Actions workflows, while others were placed in the `hack/` directory. Dependency management for this code was inconsistent and lacked a systematic approach.
+
+Recently, a separate repository, `python-gardenlinux-lib`, was introduced to host Python code for various usages. The work done to implement this separation brought benefits such as dedicated tests and linters, but these improvements are not inherently tied to having a separate repository.
+
+Currently, the lib repo contains code used in GitHub Actions workflows and feature-related logic. This introduces an indirection, making debugging and modifying pipeline code more complex, as changes require updates in the lib repo, releasing a new version, and then updating the dependency in the main repo.
```
moritzkirchnersap, line 15: > making debugging and modifying pipeline code more complex, as changes require updates in the lib repo, releasing a new version, and then updating the dependency in the main repo.

Let's say I want to debug a pipeline that fails due to python-gardenlinux-lib code: how would I debug that in the pipeline context? I'm asking out of curiosity not to critique.

fwilhe: not sure if I understand the question correctly 

I'm not talking about attaching a debugger, but I am talking about making changes with the intention of isolating an issue.

Current procedure would probably be:
- create branch on the main repo
- change the version ref of the lib to some branch
- work on that branch of the lib (add prints, whatever is needed, change code)
- once that seems to work, open a PR on the lib, get it merged, create a new release, update the dependency in the main repo

Procedure if this ADR is adopted:
- create branch on the main repo
- work on that branch of the main repo (add prints, whatever is needed, change code)
- once that seems to work, open a PR on the main repo, get it merged

less indirection is better, imo

yeoldegrove: For functionality that is consumed by `gardenlinux`  from `python-gardenlinux-lib` either by workflow or by script, the used functionality should have tests (in `python-gardenlinux-lib`) so that e.g. malformed input is detected before a release is made. If something still slips through, `python-gardenlinux-lib` already exposes script endpoints to test the functionality (if testing is even possible due to the complex nature of functionality) and may contain `--debug` or `--dry-run` flags for easier testing.

yeoldegrove:
While I like @Gchbg’s idea of a clean two-repo world (`gardenlinux` + `gardenlinux-tooling`), I don’t think it fully reflects the complexity of our current build and release setup. I’m also not sure how much @nkraetzschmar’s new package build system will simplify this part in the short term.

Right now, the picture roughly looks like this:

* Package builds in ~120 `(bp-)package` repos
  * This *might* change with the new package build system.
* `builder` repo with tooling to build **both** packages and images
  * Unclear yet how this will evolve with the new build system.
* `python-gardenlinux-lib` with tooling around:
  * parsing features and flavors
  * parsing APT repositories
  * releasing to S3 buckets
  * releasing to OCI registries
  * releasing on GitHub
* `gardenlinux` repo which (originally) was meant to be mostly platforms, features and flavors, but also contains:
  * automation to build, test and release images
    * using `builder` and `python-gardenlinux-lib`
    * plus a lot of bash/python scripts maintained directly in this repo
  * and in practice much more; introducing `python-gardenlinux-lib` already removed many “unstable” parts that differed between release branches and were hard to track (technical debt, removed by @NotTheEvilOne ). With the discontinuation of 1592 this should improve further.

So there is no truly “ideal” layout; we’re iterating towards something more maintainable rather than doing a clean greenfield design.

I’m not fundamentally against a rename or a split into a `*-cli` repo for the tooling. Keeping CLI and lib code close can also be an advantage, because it keeps shared logic and user-facing tools in sync.

Given all of the above, a few possible conclusions / next steps I could imagine are:
  * Create a new ADR to clearly define what *must* live in which repositories, also add developer and user documentation on that.
  * Consider renaming `python-gardenlinux-lib` to something like `python-gardenlinux` and optionally split out a `gardenlinux-cli` if the CLI surface becomes large enough to justify its own package/repo.
  * Keep the current repo setup until the new package build system PoC has proven itself.

fwilhe:
There is no agreement on this, closing the PR, may be revisited later if there is interest

PR #3970, opened 2025-12-01
Feature/3165 dependabot for opentofu providers

### LINKED ISSUE #3165: dependabot for OpenTofu provider versions

It would make sense to evaluate if dependabot can also manage OpenTofu provider versions.

**Why is this needed**:

OpenTofu provider versions may contain security fixes.

commel:
Enables opentofu support für dependabot.

Fixes #3165

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,7 +23,6 @@ updates:
     directory: "/tests-ng/util/tf/"
     schedule:
       interval: "weekly"
-    versioning-strategy: "increase-if-necessary"
     ignore:
       - dependency-name: "hashicorp/azurerm"
 
@@ -31,6 +30,5 @@ updates:
     directory: "/tests/platformSetup/tofu/"
     schedule:
       interval: "weekly"
-    versioning-strategy: "increase-if-necessary"
     ignore:
       - dependency-name: "hashicorp/azurerm"
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,8 +1,0 @@
-# terraform/opentofu configuration
-tests/platformSetup/tofu/*.tf @gardenlinux/maintainers
-tests-ng/util/tf/*.tf @gardenlinux/maintainers
-
-# python dependencies
-tests/Pipfile @gardenlinux/maintainers
-tests-ng/util/requirements.txt @gardenlinux/maintainers
-tests-ng/util/requirements-dev.txt @gardenlinux/maintainers
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,3 +3,12 @@
 
 # Gardener Responsibility for the Gardener Feature
 /features/gardener/* @gardenlinux/gardener-skipper-operating-system
+
+# terraform/opentofu configuration
+tests/platformSetup/tofu/*.tf @gardenlinux/maintainers
+tests-ng/util/tf/*.tf @gardenlinux/maintainers
+
+# python dependencies
+tests/Pipfile @gardenlinux/maintainers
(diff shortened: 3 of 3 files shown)


.github/dependabot.yml
```diff
@@ -6,21 +6,32 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-    reviewers:
-      - "gardenlinux/garden-linux-maintainers"
-
+    assignees:
```
yeoldegrove: Why is this changed from `reviewers` to assignee?

.github/dependabot.yml
```diff
@@ -6,21 +6,32 @@ updates:
...
+ 
   - package-ecosystem: "pip"
     directory: "/tests"
     schedule:
       interval: "weekly"
     versioning-strategy: "increase-if-necessary"
-    reviewers:
-      - "gardenlinux/garden-linux-maintainers"
-
+    assignees:
```
yeoldegrove: Why is this changed from `reviewers` to assignee?

.github/dependabot.yml
```diff
@@ -6,21 +6,32 @@ updates:
...
+      - gardenlinux/garden-linux-maintainers
+    
   - package-ecosystem: "pip"
     directory: "/tests-ng/util"
     schedule:
       interval: "weekly"
     versioning-strategy: "increase-if-necessary"
-    reviewers:
-      - "gardenlinux/garden-linux-maintainers"
+    assignees:
```
yeoldegrove: Why is this changed from `reviewers` to assignee?

.github/dependabot.yml
```diff
@@ -6,21 +6,32 @@ updates:
...
+      - gardenlinux/garden-linux-maintainers
+
+  - package-ecosystem: "terraform"
+    directory: "/tests-ng/util/tf/"
```
yeoldegrove, line 23: The old test framework should also be tested against (if this is easily possible).

.github/dependabot.yml
```diff
@@ -6,21 +6,32 @@ updates:
...
+    schedule:
+      interval: "weekly"
+    versioning-strategy: "increase-if-necessary"
+    ignore:
+      - dependency-name: "hashicorp/azurerm"
+    assignees:
```
yeoldegrove: Why is this changed from `reviewers` to assignee?

.github/dependabot.yml
commel: In the configuration documentation reviewers is not mentioned anymore: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference

But the customization documentation still refers to this: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependabot-prs#automatically-adding-reviewers

Maybe we just have to try it out.

yeoldegrove: I would read https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependabot-prs#automatically-adding-reviewers as "use `CODEOWNERS.md` to assign reviewers". I guess we do not want `assignees` at all, so we should just drop both.

commel: Good point, done. The file is called `CODEOWNERS` I think.

yeoldegrove: `CODEOWNERS` already exists in the root of this repo and we don't need to overwrite it, yet.

This is still thrown by the CI.

```
The property '#/updates/3/' contains additional properties ["versioning-strategy"] outside of the schema when none are allowed
The property '#/updates/4/' contains additional properties ["versioning-strategy"] outside of the schema when none are allowed
```

PR #3996, opened 2025-12-02
feat: output junit xml on containerized run of test-ng

nkraetzschmar:


tests-ng/util/container/Containerfile
```diff
@@ -15,4 +15,4 @@ COPY --from=compile /work/enter_host_ns /enter_host_ns
 COPY --from=extract /work/tests-ng /tests-ng
 WORKDIR /tests-ng
 ENTRYPOINT [ "/enter_host_ns" ]
-CMD [ "./run_tests", "--system-booted" ]
+CMD [ "./run_tests", "--junit-xml", "output/test-ng.xml", "--system-booted" ]
```
yeoldegrove, line 18: Could we, out of consistency, use the same directory and log file name as in the other test environments?

https://github.com/gardenlinux/gardenlinux/blob/3e0851e2d9eac4dec5ba1f4fd7663d7203a1ebfa/tests-ng/util/run_chroot.sh#L38-L40
https://github.com/gardenlinux/gardenlinux/blob/3e0851e2d9eac4dec5ba1f4fd7663d7203a1ebfa/tests-ng/util/run_qemu.sh#L54-L56
https://github.com/gardenlinux/gardenlinux/blob/3e0851e2d9eac4dec5ba1f4fd7663d7203a1ebfa/tests-ng/util/run_cloud.sh#L79-L81

nkraetzschmar: Those are the names _outside_ each test environment; as in how the file is called outside the chroot, vm, etc, right?

While this container is _primarily_ intended to run tests on gardener / kubernetes nodes, it could be used in all sorts of environments. So I'd say the file inside the container should keep the generic `test-ng.xml` name and can then be bind mounted to e.g. `gardener.test-ng.xml` outside the container depending on usage.

gehoern (approved):

Issue #3989, opened 2025-12-02
test-ng: avoid need to re-create the bundle when re-running tests


fwilhe:
Currently, small changes in the tests require a lot of file operations for re-creating bundle files. It would be good if we can make re-running tests with small changes more efficiently 

Issue based on a discussion with @vivus-ignis , @yeoldegrove , @moritzkirchnersap and @fwilhe

Issue #3860, opened 2025-11-12
test-ng: enhance experience of `dev` feature and debugging options


yeoldegrove:
**What would you like to be added**:

The `dev` feature currently provides you with a root shell on the tty if you use it to build a feature.
This root shell is very rudimentary as input is not shown and output is not parsed properly.
https://github.com/gardenlinux/gardenlinux/blob/f9203ab616f9318d629846c11cab0a61d9060eda/tests-ng/util/run_qemu.sh#L407-L411

This could be enhanced.

In addition to that, linux systems can pass `--debug` to `./test-ng` to get a Qemu console showing a graphical tty. This does not work on mac currently.

**Why is this needed**:

Enhance the debugging possibilities in case images do not boot.

PR #4047, opened 2025-12-11
3989 test ng avoid need to re create the bundle

vivus-ignis:

Fixes https://github.com/gardenlinux/gardenlinux/issues/3989, https://github.com/gardenlinux/gardenlinux/issues/3860 and replaces https://github.com/gardenlinux/gardenlinux/pull/3948

**Definition of Done:**
- [x] Solution works both on x86 linux and macos
- [x] Solution works on arm macos
- [x] No modifications of GL VM images needed
- [x] No VM runners like vagrant or lima used, just plain qemu
- [x] No runtime rebuilds should be required if anything changes in tests-ng directory
- [x] Tests runner stays in background and reacts on FS events to re-run tests on changes

**How to test:**

Run test-ng with '--dev' option. Edit any test-ng file and watch the test suite being rerun automatically. Try to update several files at the same time and see they are being synced, after which test runner starts.

### DIFF
--- a/features/gcp/exec.config
+++ b/features/gcp/exec.config
@@ -7,7 +7,7 @@ ln -sf /usr/share/zoneinfo/UTC /etc/localtime
 # cannot execute startup script in /tmp, usr /var/tmp instead
 sed -i 's/^run_dir =\s*/run_dir = \/var\/tmp/g' /etc/default/instance_configs.cfg
 
-cat >>/etc/ssh/sshd_config <<EOF
+cat >>/etc/ssh/sshd_conf <<EOF
 
 # Needed for google oslogin
 AuthorizedKeysCommand /usr/libexec/google_authorized_keys


tests-ng/util/dev.sh
```diff
@@ -0,0 +1,340 @@
...
+		go install github.com/AgentCosmic/xnotify@v${XNOTIFY_VERSION}
+		cp -v "$(go env GOPATH)/bin/xnotify" "${XNOTIFY_CLIENT_BIN_FILE}"
+		printf "Done.\n"
+	)
+}
+
+build_unfsd_podman() {
+	printf "==>\t\tbuilding unfsd...\n"
+	cat <<EOF >"${BUILD_DIR}/Containerfile.unfsd"
+FROM ubuntu:16.04 AS build
```
fwilhe: ubuntu 16? why do we need this old version?

vivus-ignis: For better portability. It has the oldest glibc I could find in a working distro.

yeoldegrove (changes requested):
Overall I think the PR is tackling a real pain point and the improved dev-loop experience is valuable - thanks for pushing this forward.

My main concern is that the current design feels quite heavy in footprint relative to the problem it’s solving, especially since this is "only" a developer-convenience feature.
Pulling in additional components like `websitino` and `daemonize` introduces dependencies that we need to compile, maintain and patch and might not be easily integratable in dependabot (we had the same issue with the python dist but solved it with custom tooling).
I'm not fully convinced we need a standalone webserver and "daemon manager" in this stack as we can easily serve files and fork processes in python.

If implementing that purely in Python feels too complex or out of scope, even a lightweight approach based on `scp` and `inotify` would be sufficient I guess.
I quickly built [this](https://github.com/gardenlinux/gardenlinux/commit/2a2abcb5726942aec386828d8b4156cef1e1e326) which works out quite fine and serves a similar purpose.

None of this stands against the value of the feature, but it might be worth exploring whether a slimmer design could achieve most of the benefits with less dependency surface. Reducing the footprint would make the tooling easier to adopt, maintain, and secure over time.

PR #4059, opened 2025-12-15
Add tests for gnutls to check if it's in FIPS-mode.

Akendo:

In this PR, we add a Python test to invoke the C function [`gnutls_fips140_mode_enabled`](https://manpages.debian.org/testing/gnutls-doc/gnutls_fips140_mode_enabled.3.en.html) to ensure that 

While there are some Python implementations that allow the usage of GnuTLS, all implementations are not working or too old to be used in our testing frame. With this approach, we are using ctypes to invoke the function directly.  

Relates: [#3898](https://github.com/gardenlinux/gardenlinux/issues/3898)

### DIFF
--- a/tests-ng/test_fips.py
+++ b/tests-ng/test_fips.py
@@ -49,8 +49,7 @@ def test_gnutls_is_in_fips_mode():
     """
     shared_lib_name = find_library("gnutls")
     gnutls = CDLL(shared_lib_name)
-    rc = gnutls.gnutls_fips140_mode_enabled()
-    assert bool(rc), "Error GnuTLS can't be started in FIPS mode."
+    assert gnutls.gnutls_fips140_mode_enabled(), "Error GnuTLS can't be started in FIPS mode."
 
 
 @pytest.mark.feature("_fips")


tests-ng/test_fips.py
```diff
@@ -34,6 +36,23 @@ def test_gnutls_fips_file_is_empty(file: File):
...
+    It will return a C-type true.
+
+    https://www.gnutls.org/manual/html_node/FIPS140_002d2-mode.html
+    https://manpages.debian.org/testing/gnutls-doc/gnutls_fips140_mode_enabled.3.en.html
+
+    """
+    shared_lib_name = find_library("gnutls")
+    gnutls = CDLL(shared_lib_name)
+    rc = gnutls.gnutls_fips140_mode_enabled()
+    assert bool(rc), "Error GnuTLS can't be started in FIPS mode."
```
```suggestion
    gnutls = CDLL(shared_lib_name)
    assert gnutls.gnutls_fips140_mode_enabled(), "Error GnuTLS can't be started in FIPS mode."
```

moritzkirchnersap (changes requested):
See comment, that works without using a return code variable and it looks a lot cleaner IMHO

moritzkirchnersap (approved):

tests-ng/test_fips.py
```diff
@@ -34,6 +36,22 @@ def test_gnutls_fips_file_is_empty(file: File):
...
+    """
+    This code will call up the GnuTLS library directly with ctypes.
+    It invokes the gnutls_fips140_mode_enabled to return true when the library is in FIPS mode;
+    It will return a C-type true.
+
+    https://www.gnutls.org/manual/html_node/FIPS140_002d2-mode.html
+    https://manpages.debian.org/testing/gnutls-doc/gnutls_fips140_mode_enabled.3.en.html
+
+    """
+    shared_lib_name = find_library("gnutls")
```
nkraetzschmar, line 50: I'm not sure using `find_library` is entirely stable here 🤔

Does ctypes simply re-use the same search paths that regular dynamic lib loading/linking would use if requested directly by the python interpreter binary itself? If so then this *could* result in unexpected behaviour, os the python binary bundled with the test framework is linked with

```
0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/../lib]
```

in order to make the whole framework re-locatable. When used in such a way the rpath might interfere with the intended behaviour though.

nkraetzschmar (changes requested):

pnpavlov (approved):
LGTM

tests-ng/test_fips.py
pnpavlov: @Akendo , please still consider the feedback of @nkraetzschmar

Akendo: @nkraetzschmar good question, but the code does plenty of strategies to detect where a shared library has been placed. It invokes tools to get this, and do not rely on rpath. It effectively uses `ldconfig -p`.

Akendo: https://github.com/python/cpython/blob/059316ade6d971cf11e8fbaa73dbedb5babebf02/Lib/ctypes/util.py#L395

Akendo: It also tries to use `gcc` and so, but other things too:

>The purpose of the [find_library()](https://docs.python.org/3/library/ctypes.html#ctypes.util.find_library) function is to locate a library in a way similar to what the compiler or runtime loader does (on platforms with several versions of a shared library the most recent should be loaded), while the ctypes library loaders act like when a program is run, and call the runtime loader directly.

https://docs.python.org/3/library/ctypes.html#ctypes.util.find_library

PR #4060, opened 2025-12-15
Add test for libgcrypt

Akendo:

In this PR, we add a Python test to invoke the C function [`gcry_control`](https://www.gnupg.org/documentation/manuals/gcrypt/Controlling-the-library.html#index-gcry_005ffips_005fmode_005factive) to ensure that libgcrypt is in FIPS mode. 

Relates: [#3898](https://github.com/gardenlinux/gardenlinux/issues/3898)

### DIFF
--- a/tests-ng/test_fips.py
+++ b/tests-ng/test_fips.py
@@ -37,6 +37,22 @@ def test_gnutls_fips_file_is_empty(file: File):
 
 
 @pytest.mark.feature("_fips")
+def test_gnutls_is_in_fips_mode():
+    """
+    This code will call up the GnuTLS library directly with ctypes.
+    It invokes the gnutls_fips140_mode_enabled to return true when the library is in FIPS mode;
+    It will return a C-type true.
+
+    https://www.gnutls.org/manual/html_node/FIPS140_002d2-mode.html
+    https://manpages.debian.org/testing/gnutls-doc/gnutls_fips140_mode_enabled.3.en.html
+
+    """
+    shared_lib_name = find_library("gnutls")
+    gnutls = CDLL(shared_lib_name)
+    assert gnutls.gnutls_fips140_mode_enabled(), "Error GnuTLS can't be started in FIPS mode."
+
+
+@pytest.mark.feature("_fips")
 def test_gnutls_fips_dot_hmac_file_is_presented():
     """
     GnuTLS will perform a self check based on the FIPS requirements. A file that


tests-ng/test_fips.py
```diff
@@ -100,6 +102,38 @@ def test_libgcrypt_fips_file_is_empty(file: File):
...
+
+    https://www.gnupg.org/documentation/manuals/gcrypt/Enabling-FIPS-mode.html#Enabling-FIPS-mode
+    https://www.gnupg.org/documentation/manuals/gcrypt/Controlling-the-library.html#index-gcry_005ffips_005fmode_005factive
+    """
+    shared_lib_name = find_library("gcrypt")
+    libgcrypt = CDLL(shared_lib_name)
+
+    # See the gcrypt.h
+    GCRYCTL_FIPS_MODE_P = 55
+    rc = libgcrypt.gcry_control(c_int(GCRYCTL_FIPS_MODE_P), c_int(1))
```
moritzkirchnersap: Are we positive this returns what we expect? On my local tests this returns:

`>>> libgcrypt.gcry_control (c_int(GCRYCTL_FIPS_MODE_P), c_int(1))`
`16777217`

Which then as a bool:

`>>> bool(16777217)`
`True`

Which kinda makes sense in python but I'm not sure if it is what we want to check? And with kinda I mean the way that python handles values when casting to bool might be a bit different from what one expects...

Akendo: C does not have a strong assumption here; the only norm is a number different from zero for true. Sometimes a negative number can also be returned as a negative number to indicate false. But I guess for Python this matters little.

Akendo: Only starting with C99 will you have some more normed way of C to deal with boolean. And I think with C23 you have more strictly defined.

Akendo: Crypto libraries also sometimes have some magic optimization to use this as a source of randomness.

tests-ng/test_fips.py
```diff
@@ -100,6 +102,37 @@ def test_libgcrypt_fips_file_is_empty(file: File):
...
+    headers.
+    # https://git.gnupg.org/cgi-bin/gitweb.cgi?p=libgcrypt.git;a=blob;f=src/gcrypt.h.in;h=712a8dd7931e76c0a83aee993bf22f34e420bfc2;hb=737cc63600146f196738a6768679eb016cf866e9#l316
+    The package libgcrypt20 needs to be presented.
+
+    The gcry_control requires c type integer as inputs.
+
+    https://www.gnupg.org/documentation/manuals/gcrypt/Enabling-FIPS-mode.html#Enabling-FIPS-mode
+    https://www.gnupg.org/documentation/manuals/gcrypt/Controlling-the-library.html#index-gcry_005ffips_005fmode_005factive
+    """
+    shared_lib_name = find_library("gcrypt")
```
nkraetzschmar, line 144: see https://github.com/gardenlinux/gardenlinux/pull/4059/changes#r2620385022

PR #4061, opened 2025-12-15
Add test for OpenSSL FIPS mode

Akendo:

In this PR we add two tests to check for the `fips.so` and that we can run openssl with FIPS mode.

### DIFF
--- a/tests-ng/test_fips.py
+++ b/tests-ng/test_fips.py
@@ -49,7 +49,9 @@ def test_gnutls_is_in_fips_mode():
     """
     shared_lib_name = find_library("gnutls")
     gnutls = CDLL(shared_lib_name)
-    assert gnutls.gnutls_fips140_mode_enabled(), "Error GnuTLS can't be started in FIPS mode."
+    assert (
+        gnutls.gnutls_fips140_mode_enabled()
+    ), "Error GnuTLS can't be started in FIPS mode."
 
 
 @pytest.mark.feature("_fips")
@@ -119,7 +121,6 @@ def test_libgcrypt_fips_file_is_empty(file: File):
 
 
 @pytest.mark.feature("_fips")
-
 def test_that_openssl_has_fips_provider_is_presented(file: File):
     """
     We have to ensure that the fips.so is presented on the system.
@@ -152,9 +153,11 @@ def test_libssl_is_in_fips_mode():
     libssl.EVP_default_properties_is_fips_enabled.argtypes = [c_void_p]
 
     ctx = libssl.OSSL_LIB_CTX_get0_global_default()
-    assert libssl.EVP_default_properties_is_fips_enabled(ctx), "Error openssl can't be started in FIPS mode."
+    assert libssl.EVP_default_properties_is_fips_enabled(
+        ctx
+    ), "Error openssl can't be started in FIPS mode."
 
-    
+
 def test_libgcrypt_is_in_fips_mode():
     """
      This will check if libgcrypt is in FIPS mode. There is no other way to call libgcrypt from
@@ -182,7 +185,9 @@ def test_libgcrypt_is_in_fips_mode():
 
     # See the gcrypt.h
     GCRYCTL_FIPS_MODE_P = 55
-    assert libgcrypt.gcry_control(c_int(GCRYCTL_FIPS_MODE_P), c_int(1)), "Error libgcrypt can't be started in FIPS mode."
+    assert libgcrypt.gcry_control(
+        c_int(GCRYCTL_FIPS_MODE_P), c_int(1)
+    ), "Error libgcrypt can't be started in FIPS mode."
 
 
 @pytest.mark.feature("_fips")


tests-ng/test_fips.py
```diff
@@ -100,6 +102,43 @@ def test_libgcrypt_fips_file_is_empty(file: File):
...
+
+    Frist we have to setup a OSSL_LIB_CTX, for this we use OSSL_LIB_CTX_get0_global_default
+    that returns a c pointer.
+
+    Once we have the global context, we can use this to get the default properties of the
+    high level algorithm has enabled. Similar to OSSL_LIB_CTX_get0_global_default, we need
+    to define the expect return value. Since this is a C boolean we have to set it to int,
+    but it we also have to define the arguments type to be a arry with a c pointer, since it
+    expect the pointer to the context  obj.
+    """
```
fwilhe, line 145:
```suggestion
    """
    We get OSSL_LIB_CTX_get0_global_default from libssl. For this we need
    to setup the correct return values

    First we have to setup a OSSL_LIB_CTX, for this we use OSSL_LIB_CTX_get0_global_default
    that returns a c pointer.

    Once we have the global context, we can use this to get the default properties of the
    high level algorithm has enabled. Similar to OSSL_LIB_CTX_get0_global_default, we need
    to define the expect return value. Since this is a C boolean we have to set it to int,
    but it we also have to define the arguments type to be an array with a c pointer, since it
    expect the pointer to the context  obj.
    """
```

tests-ng/test_fips.py
```diff
@@ -100,6 +102,43 @@ def test_libgcrypt_fips_file_is_empty(file: File):
...
+
+    libssl.OSSL_LIB_CTX_get0_global_default.restype = c_void_p
+    libssl.OSSL_LIB_CTX_get0_global_default.argtypes = []
+    libssl.EVP_default_properties_is_fips_enabled.restype = c_int
+    libssl.EVP_default_properties_is_fips_enabled.argtypes = [c_void_p]
+
+    ctx = libssl.OSSL_LIB_CTX_get0_global_default()
+    rc = libssl.EVP_default_properties_is_fips_enabled(ctx)
+    assert bool(rc), "Error openssl can't be started in FIPS mode."
+
```
moritzkirchnersap, line 191:
```suggestion
    assert libssl.EVP_default_properties_is_fips_enabled(ctx), "Error openssl can't be started in FIPS mode."
```

RC Code should be zero or one, right?

moritzkirchnersap (changes requested):

tests-ng/test_fips.py
Akendo: Missed committing the spellchecked version. Thanks for spotting this!

tests-ng/test_fips.py
Akendo: C does not have a strong assumption here; the only norm is a number different from zero for true. Sometimes a negative number can also be returned as a negative number. But I guess for Python this matters little.

tests-ng/test_fips.py
```diff
@@ -100,6 +102,43 @@ def test_libgcrypt_fips_file_is_empty(file: File):
...
+    libssl = CDLL(shared_lib_name)
+
+    libssl.OSSL_LIB_CTX_get0_global_default.restype = c_void_p
+    libssl.OSSL_LIB_CTX_get0_global_default.argtypes = []
+    libssl.EVP_default_properties_is_fips_enabled.restype = c_int
+    libssl.EVP_default_properties_is_fips_enabled.argtypes = [c_void_p]
+
+    ctx = libssl.OSSL_LIB_CTX_get0_global_default()
+    rc = libssl.EVP_default_properties_is_fips_enabled(ctx)
+    assert bool(rc), "Error openssl can't be started in FIPS mode."
```
```suggestion
```

Those two lines are not needed any longer, right?

moritzkirchnersap (changes requested):

tests-ng/test_fips.py
moritzkirchnersap: Python follows the same assumptions. I didn't express myself properly. What I meant was: the return code can be handed to assert directly.

And my suggestion only went over one line when it should have covered three.

tests-ng/test_fips.py
```diff
@@ -100,6 +102,43 @@ def test_libgcrypt_fips_file_is_empty(file: File):
...
+returns a C pointer.
+
+    Once we have the global context, we can use this to get the default properties of the high-level
+    algorithm that has been enabled. Similar to OSSL_LIB_CTX_get0_global_default, we need to define
+    the expected return value. Since this is a C boolean, we have to set it to int, but we also have
+    to define the argument type to be an array with a C pointer, since it is the pointer to the
+    context object.
+    """
+
+    shared_lib_name = find_library("ssl")
```
nkraetzschmar, line 147: see https://github.com/gardenlinux/gardenlinux/pull/4059/changes#r2620385022

pnpavlov (approved):
LGTM

PR #4067, opened 2025-12-16
GnuTLS FIPS test needs a booted VM.

Akendo:
Fixes the GnuTLS running in chroot. It needs to have a booted system with fips mode enabled to pass.

### DIFF
--- a/tests-ng/plugins/efivars.py
+++ b/tests-ng/plugins/efivars.py
@@ -1,0 +1,38 @@
+import os
+from typing import Optional
+
+import pytest
+
+
+class EfiVarsUUID:
+    def __init__(self, uuid: str, efivars_dir: str):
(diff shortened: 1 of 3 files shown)


tests-ng/test_fips.py
```diff
@@ -37,6 +37,7 @@ def test_gnutls_fips_file_is_empty(file: File):
 
 
 @pytest.mark.feature("_fips")
+@pytest.mark.booted("GnuTLS needs to have a VM booted with boot FIPS mode.")
```
fwilhe: i think you need to pass the parameter as described here https://github.com/gardenlinux/gardenlinux/blob/dd61237dda55535b7e66cbac5cb1a088be08957e/tests-ng/DEVELOPER.md?plain=1#L408

PR #4071, opened 2025-12-16
feat: test-ng: migrate umask tests

### LINKED ISSUE #4036: Migrate `base/test/test_mask.py` to new testing framework

nkraetzschmar:
Resolves #4036

> [!NOTE]
> https://github.com/gardenlinux/gardenlinux/pull/4071/changes/99d468e0e76ada82cc8efa72a482ff00037b2f2f implies that all changes to mount points are ignored by sysdiff. For now that is needed since otherwise `su --login` starting the `run-user-0.mount` will always fail the sysdiff checks. Eventually we should add an explicit mount point check that only ignores those mount points and not others, by directly comparing the mount points and not just the `/proc/mounts` and `/etc/mtab` file contents.

### DIFF
--- a/tests-ng/plugins/sysdiff.py
+++ b/tests-ng/plugins/sysdiff.py
@@ -388,7 +388,9 @@ class SnapshotManager:
             LoadedKernelModule(name=module) for module in kernel_modules_list
         ]
 
-        ignore_patterns = DEFAULT_IGNORE_PATTERNS + file_collector.load_ignore_patterns(ignore_file)
+        ignore_patterns = DEFAULT_IGNORE_PATTERNS + file_collector.load_ignore_patterns(
+            ignore_file
+        )
         normalized_paths = file_collector.normalize_paths(paths)
 
         files_dict = file_collector.collect_file_hashes(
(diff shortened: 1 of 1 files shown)


tests-ng/test_umask.py
```diff
@@ -0,0 +1,15 @@
...
+from plugins.parse_file import ParseFile
+from plugins.shell import ShellRunner
+
+
+def test_umask_file(parse_file: ParseFile):
+    config = parse_file.parse("/etc/login.defs", format="spacedelim")
+    assert config["UMASK"] == "027"
+
+
+@pytest.mark.root
```
fwilhe, line 11: why does this need root?

tests-ng/test_umask.py
```diff
@@ -0,0 +1,15 @@
...
+def test_umask_cmd(shell: ShellRunner):
+    result = shell("su -l -c 'umask'", capture_output=True)
```
fwilhe:
```suggestion
    result = shell("su --login --command 'umask'", capture_output=True)
```

please use long opts for readability

tests-ng/test_umask.py
nkraetzschmar: we don't really know / can know what the umask for some other user might be, but for root we can be certain that it _should_ be whatever the default is and would be a valid test failure if this is changed. but for other users a modified umask in e.g. `.bashrc` is totally valid.

PR #4072, opened 2025-12-16
Add labels to container images

### LINKED ISSUE #4057: Add labels to our container images to indicate product, version, architecture
Based on a discussion with @skateball

It would be nice if we had [labels](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#labelling-container-images) on our OCI images to indicate what this image is.

This is useful for example in cases where a tag is re-used, like the SapMachine 25 image based on Garden Linux 1592 will be updated when we have a new minor release of 1592, but will use the same tag that does not contain the minor version. The image will still be available by its SHA, but it is not easy to tell what this image is.

A label containing the information that this is `Garden Linux 1592.8 arm64` would be very helpful in this case.

saubhikdattagithub:
Adds labels to container images, which allows to identify an image even if it has no tag.

Fixes #4057

### DIFF
--- a/features/container/image.oci
+++ b/features/container/image.oci
@@ -29,8 +29,8 @@ cat > config << EOF
       "org.opencontainers.image.architecture": "$BUILDER_ARCH",
       "org.opencontainers.image.source": "https://github.com/gardenlinux/gardenlinux",
       "org.opencontainers.image.revision": "$BUILDER_COMMIT",
-      "org.opencontainers.image.description": "$BUILDER_FEATURES"
-    }
+      "org.opencontainers.image.description": "Garden Linux $BUILDER_VERSION $BUILDER_ARCH $BUILDER_FEATURES"
+}
   },
   "rootfs": {
     "type": "layers",


saubhikdattagithub:
Editing `features/container/image.oci` can result in Labels as below from local image created (oci)

```bash
cat > config << EOF
{
  "architecture": "$BUILDER_ARCH",
  "os": "linux",
  "config": {
    "Cmd": [ "/bin/bash" ],
    "Labels": {
      "org.opencontainers.image.title": "Garden Linux",
      "org.opencontainers.image.version": "$BUILDER_VERSION",
      "org.opencontainers.image.architecture": "$BUILDER_ARCH"
    }
  },
  "rootfs": {
    "type": "layers",
    "diff_ids": [
      "sha256:$rootfs_sha256"
    ]
  },
```
JSON parameters including Labels (`org.opencontainers.image.ABC`):

```bash
[
     {
          "Id": "41f0c8aa6d2411364926d216fd56b591f90601077bcc009e61dd16b9a1bb703b",
          "Digest": "sha256:a9d3f2044703234f3b22b26710958797c6d5ccc4874041d7759696a383434ade",
          "RepoTags": [],
          "RepoDigests": [],
          "Parent": "",
          "Comment": "",
          "Created": null,
          "Config": {
               "Cmd": [
                    "/bin/bash"
               ],
               "Labels": {
                    "org.opencontainers.image.architecture": "arm64",
                    "org.opencontainers.image.title": "Garden Linux",
                    "org.opencontainers.image.version": "today"
               }
          },
          "Version": "",
          "Author": "",
          "Architecture": "arm64",
          "Os": "linux",
          "Size": 569580928,
          "VirtualSize": 569580928,
          "GraphDriver": {
               "Name": "overlay",
               "Data": {
                    "UpperDir": "/var/home/core/.local/share/containers/storage/overlay/55d9ec1de80caa1d0aac9445f6765cc41ea60f9e7831d730dabbecfa662ceaeb/diff",
                    "WorkDir": "/var/home/core/.local/share/containers/storage/overlay/55d9ec1de80caa1d0aac9445f6765cc41ea60f9e7831d730dabbecfa662ceaeb/work"
               }
          },
          "RootFS": {
               "Type": "layers",
               "Layers": [
                    "sha256:55d9ec1de80caa1d0aac9445f6765cc41ea60f9e7831d730dabbecfa662ceaeb"
               ]
          },
          "Labels": {
               "org.opencontainers.image.architecture": "arm64",
               "org.opencontainers.image.title": "Garden Linux",
               "org.opencontainers.image.version": "today"
          },
          "Annotations": {},
          "ManifestType": "application/vnd.oci.image.manifest.v1+json",
          "User": "",
          "History": [
               {}
          ],
          "NamesHistory": null
     }
]
```

**_Query:_** Is the VERSION variable (i.e $BUILDER_VERSION) to get the RELEASE VERSION: 1592.XY, correct?

features/container/image.oci
```diff
@@ -26,7 +27,13 @@ cat > config << EOF
     "Labels": {
       "org.opencontainers.image.title": "Garden Linux",
       "org.opencontainers.image.version": "$BUILDER_VERSION",
-      "org.opencontainers.image.architecture": "$BUILDER_ARCH"
+      "org.opencontainers.image.architecture": "$BUILDER_ARCH",
+      "org.opencontainers.image.source": "https://github.com/gardenlinux/gardenlinux",
+      "org.opencontainers.image.revision": "$BUILDER_COMMIT",
+      "org.opencontainers.image.created": "$NOW",
```
fwilhe: nice idea, but we should remove this as this interferes with reproducible builds. running the same build twice will produce two different artifacts because the time is different, which is not a property we want

features/container/image.oci
```diff
@@ -26,7 +27,13 @@ cat > config << EOF
...
+      "org.opencontainers.image.description": "Garden Linux $BUILDER_VERSION $BUILDER_ARCH",
+      "org.opencontainers.image.ref.name": "$BUILDER_FEATURES",
+      "org.opencontainers.image.base.name": "$BUILDER_CNAME"
```
fwilhe: both of those properties don't match the description from the oci spec. please remove them.

features/container/image.oci
```diff
@@ -26,7 +27,13 @@ cat > config << EOF
     "Labels": {
       "org.opencontainers.image.title": "Garden Linux",
       "org.opencontainers.image.version": "$BUILDER_VERSION",
-      "org.opencontainers.image.architecture": "$BUILDER_ARCH"
+      "org.opencontainers.image.architecture": "$BUILDER_ARCH",
+      "org.opencontainers.image.source": "https://github.com/gardenlinux/gardenlinux",
+      "org.opencontainers.image.revision": "$BUILDER_COMMIT",
+      "org.opencontainers.image.created": "$NOW",
+      "org.opencontainers.image.description": "Garden Linux $BUILDER_VERSION $BUILDER_ARCH",
```
fwilhe: please include the builder features here

saubhikdattagithub: Okay, but wont we need a single Label containing `Garden Linux 1592.8 arm64` ?

> A label containing the information that this is Garden Linux 1592.8 arm64 would be very helpful in this case.

fwilhe: no, the above was my first idea, but I do think that including the features is useful. pease add them.

fwilhe: No, that's not what I meant. Adding the features additionally. let me quickly push a fix.

fwilhe: this is what I meant

```
      "org.opencontainers.image.description": "Garden Linux $BUILDER_VERSION $BUILDER_ARCH $BUILDER_FEATURES"
```

this will give us output like
`Garden Linux today arm64 _curl,sapmachine,_slim,base,container`, which I think is useful

saubhikdattagithub: Ah, this is NICE :-)

fwilhe: the important thing imo is that this allows a pretty good trace of what the image is, even if it has no tag. that's what the original intention of the issue was.

PR #4083, opened 2025-12-19
Skip 2 tests for CIS compliance

saubhikdattagithub:
Request to skip 2 test-ng (sysctl: rp_filter=1 and umask=0027) for CIS compliance

Fixes #3829 

image
image

### DIFF
--- a/tests-ng/test_cis.py
+++ b/tests-ng/test_cis.py
@@ -1,6 +1,7 @@
 import pytest
 from plugins.shell import ShellRunner
 
+
 @pytest.mark.feature("cis")
 @pytest.mark.root(reason="CIS audit requires root privileges")
 @pytest.mark.booted(reason="CIS tests within a booted system")


tests-ng/test_umask.py
```diff
@@ -9,6 +9,7 @@ def test_umask_file(parse_file: ParseFile):
 
 
 @pytest.mark.root
+@pytest.mark.feature("not cis", reason="CIS needs umask 0027")
```
fwilhe: this test asserts that the umask is 0027, I don't understand the provided reason. is there a typo?

tests-ng/test_sysctls.py
```diff
@@ -2,6 +2,7 @@
 
 
 @pytest.mark.booted
+@pytest.mark.feature("not cis", reason="CIS needs rp_filter to be strict")
```
fwilhe, line 5: is there a cis version of this test, to ensure the cis feature that this set correctly?

saubhikdattagithub: No, all sysctl parameters, donot have individual tests to be verified.

tests-ng/test_umask.py
saubhikdattagithub: > this test asserts that the umask is 0027, I don't understand the provided reason. is there a typo?

yes, my bad. CIS has `echo "umask 077" >> /etc/profile` 
the reason mentioned here has a typo, in CIS a stricter permission is enforced with UMASK (077) (file:600/dir:700) against UMASK (027) (files:640/dir:750)

tests-ng/test_sysctls.py
nkraetzschmar: we might need to discuss this with gardener. cause while I personally would prefer always setting `rp_filter=1` this may cause issues with CNI. probably fine to skip for CIS for now, but this will mean CIS feature will have limited compatibility with gardener.

tests-ng/test_umask.py
```diff
@@ -9,6 +9,7 @@ def test_umask_file(parse_file: ParseFile):
 
 
 @pytest.mark.root
+@pytest.mark.feature("not cis", reason="CIS has 077")
```
nkraetzschmar, line 12: please add a CIS specific test that checks for `0077`

tests-ng/test_sysctls.py
saubhikdattagithub: In that case, alternately, we could agree and mark this as Passed within features/cis with relevant documentation

saubhikdattagithub: net.ipv4.rp_filter=1 check disabled in CIS compliance for Gardener requirement

Issue #3774, opened 2025-11-03
Add support for baremetal instance types


pnpavlov:
[Automatic In-Place Updates](https://gardener.cloud/docs/gardener/shoot-operations/shoot_updates/#automatic-in-place-updates) works only with Garden Linux images that have -inplace in the name, for example 1877.2.0-inplace. I would like to use this feature with bare metal nodes, but bare metal nodes require images with -baremetal and these are not supported for in-place updates:
machine image version 'gardenlinux:1877.1.0-baremetal' does not support in-place updates, supported machine image versions are: [gardenlinux:1877.4.0-inplace]

Would it be possible to release -inplace-baremetal versions of images? In-place updates would be incredibly useful for baremetal nodes for a whole bunch of reasons.

Acceptance criteria:
- In-Place Update has been successfully implemented and tested on public cloud bare metal nodes
- Openstack isn't in scope. Will be handle in a separate issue

pnpavlov:
Scope of this issue: Please focus on the public cloud providers AWS, Azure, Google. 
Out of scope: Private cloud destinations.

PR #4086, opened 2025-12-22
usi feature for openstackbaremetal artifacts

vivus-ignis:
There is no openstackbaremetal images which support inplace updates and this is requested by the gardener team.

Fixes https://github.com/gardenlinux/gardenlinux/issues/3774

### DIFF
--- a/flavors.yaml
+++ b/flavors.yaml
@@ -1,4 +1,3 @@
----
 targets:
   - name: ali
     category: public-cloud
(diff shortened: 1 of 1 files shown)


yeoldegrove (changes requested):
Can we please double check if this really fixes #3774 as it does not mention OpenStack.
Also, is this only supposed to be released on `arm64`?

PR #4102, opened 2026-01-02
fix: use 4 cores for the qemu tests on arm

5kt:


fwilhe:
@5kt @nkraetzschmar any reason to keep this in draft state? Merge or close?

PR #4129, opened 2026-01-07
cleanup: drop old tests

nkraetzschmar:
Since we have now migrated all tests to the new test framework we can finally drop the old test framework.


gehoern (approved):
lots of files to be remove, but the changed code looks good to me!

PR #4142, opened 2026-01-12
Update README.md to reflect deletion of tests and update to tests-ng

moritzkirchnersap:


README.md
```diff
@@ -84,7 +84,7 @@ Garden Linux images can be built with the *trustedboot* feature—an extension t
 # Test
 
 To run unit tests for a specific target, use the command `./test ${target}`.
```
fwilhe: this is outdated aswell, test-ng script is called with a path to the image under test

PR #4147, opened 2026-01-13
Ensure GitHub workflow cloud tests are only executed if prerequisites are met

NotTheEvilOne:
This PR differentiates between GitHub workflow tests executed on GitHub and on cloud resources. As cloud resource set-up and handling is usually more expensive (e.g. time wise) the tests are only executed if all other tests are successfully. To not execute `chroot` and `oci` tests on each run as they indicate deeper issues that needs to be resolved first we might allow `fail-fast` for these tests to faster stop testing in these cases.

`qemu` tests are prerequisites for `cloud` tests as they most specifically verify availability of specific services and packages required for certain cloud providers.

As there was a glitch with GitHub workflow matrix results `cloud` tests have been executed unconditionally even if `qemu` tests failed before.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -83,7 +83,7 @@ jobs:
             const gitHubLib = await import("${{ github.workspace }}/.github/workflows/github.mjs");
 
             const gitHubRef = "${{ github.head_ref == '' && github.ref_name || github.head_ref }}";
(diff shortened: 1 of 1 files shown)


yeoldegrove:
@NotTheEvilOne @5kt What do you think about settings `dispatchRetryWorkflow` to only run 3 times in the nightly for both build and tests jobs?

Issue #4197, opened 2026-01-19
Error GnuTLS can't be started in FIPS mode


yeoldegrove:
**What happened:**

Our nightly build's CI fails with "Error GnuTLS can't be started in FIPS mode".

**Debug Logs:**

https://github.com/gardenlinux/gardenlinux/actions/runs/21127073590/job/60751399427

```
 _____________________________________________ test_gnutls_is_in_fips_mode ______________________________________________
test_fips.py:53: in test_gnutls_is_in_fips_mode
    assert (
E   AssertionError: Error GnuTLS can't be started in FIPS mode.
E   assert 0
E    +  where 0 = <_FuncPtr object at 0xffffab433a70>()
E    +    where <_FuncPtr object at 0xffffab433a70> = <CDLL 'libgnutls.so.30', handle aaab1ae6ab80 at 0xffffab05a990>.gnutls_fips140_mode_enabled
```

PR #4201, opened 2026-01-19
Enhance the HMAC test validate all hashes.

Akendo:
Based on the problem with nightly and FIPS, this will improve the tests to not only test for the HMAC file for GnuTLS, but also for GMP and Nettle. 

Relates: #4197


yeoldegrove (approved):
LGTM

tests-ng/test_fips.py
```diff
@@ -82,7 +63,12 @@ def test_gnutls_fips_dot_hmac_file_is_vaild():
     |<1>| FIPS140-2 self testing part 2 failed
 
     """
+    lib_paths = {}
     gnutls_lib_path = f"/usr/lib/{arch()}-linux-gnu/libgnutls.so.30"
+    lib_paths["libhogweed"] = f"/usr/lib/{arch()}-linux-gnu/libhogweed.so.6"
```
moritzkirchnersap, line 68: Why does gnutls_lib_path get an extra treatment? Would it not be cleaner to put it into lip_paths and the assert for loop and to to treat every lib equal? Or am I missing something?

Akendo: Already fixed. I had issues with the symlinks those libraries sometimes have. This was my code I knew would work.

moritzkirchnersap (changes requested):
Small comment about the file layout more than the logic

tests-ng/test_fips.py
moritzkirchnersap: And what do we think about parametrizing instead of a for looping?

tests-ng/test_fips.py
```diff
@@ -82,22 +63,47 @@ def test_gnutls_fips_dot_hmac_file_is_vaild():
...
+    for key in lib_paths.keys():
+        fips_hmac = hmac.new(key=SECRET.encode("UTF-8"), msg=None, digestmod=SHA256)
+        # Need to read it 'b' since it's a binary file.
+        with open(lib_paths[key], mode="rb") as lib:
+            fips_hmac.update(lib.read())
+
+        assert (
+            config[os.path.basename(lib_paths[key])]["hmac"] == fips_hmac.hexdigest()
+        ), f"Computed HMAC is incorrect for {lib_paths[key]}!"
+
```
moritzkirchnersap, line 88:
```suggestion
    lib_paths = [
        f"/usr/lib/{arch()}-linux-gnu/libgnutls.so.30",
        f"/usr/lib/{arch()}-linux-gnu/libhogweed.so.6",
        f"/usr/lib/{arch()}-linux-gnu/libnettle.so.8",
        f"/usr/lib/{arch()}-linux-gnu/libgmp.so.10",
    ]

gnutls_fips_hmac_file = f"/usr/lib/{arch()}-linux-gnu/.libgnutls.so.30.hmac"
# https://gitlab.com/gnutls/gnutls/-/blob/master/configure.ac?ref_type=heads#L677
SECRET = "orboDeJITITejsirpADONivirpUkvarP"

config = configparser.ConfigParser()
config.read(gnutls_fips_hmac_file)

    for lib_file in lib_paths:
        fips_hmac = hmac.new(key=SECRET.encode("UTF-8"), msg=None, digestmod=SHA256)
        # Need to read it 'b' since it's a binary file.
        with open(lib_file, mode="rb") as lib:
            fips_hmac.update(lib.read())

        assert (
            config[os.path.basename(lib_file)]["hmac"] == fips_hmac.hexdigest()
        ), f"Computed HMAC is incorrect for {lib_file}!"

```

PR #4207, opened 2026-01-20
Draft for building the linux kernel with package-build

Akendo:


docs/01_developers/tutorials/How to build a package.md
```diff
@@ -0,0 +1,129 @@
...
+
+Now we need to place a target package we want to build.
+
+## Building a package
+
+We recommend using a separate folder next to the package-build directory because checking out a
+separate repository within the package-build folder will confuse the git control structure. It’s
+possible to download the zip file only, which this tutorial does not cover.
+
+To make the target package available within the package-build folder, we're going to use a softlink.
```
fwilhe, line 56: I would not recommend creating a link, this link might get committed accidentally to the package-build repo easily

Instead, the recommended procedure should be to call `build` from the parent directory, i.e. if `package-build` and `package-linux` are both cloned inside PWD, one can run

```
./package-build/build package-linux
```

docs/01_developers/tutorials/How to build a package.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+
+Test that the you can execute it:
+
+```bash
+./build
+```
+
+You should get now an error message saying: ``unbound variable``, this is expected.
```
fwilhe, line 31: I would omit this section, I think this is confusing and does not explain much to the reader

docs/01_developers/tutorials/How to build a package.md
```diff
@@ -0,0 +1,129 @@
...
+ln -s ../package-linux/ ./package-linux
+```
+
+Now the package builder can be invoked with the parameter ``./package-linux``, and it will start a
+container, fetch the sources, apply our changes, and build the package, and in the end you will have
+a package within the ``.build`` folder of the ``package-linux``.
+
+
+```bash
+./build package-linux
```
fwilhe, line 71: `--arch` should be mentioned here as the default behaviour is to build for intel even if the user is running a machine with native arm (like modern macbooks)

PR #4210, opened 2026-01-21
docs: Add ADR 30 about the scope of python-gardenlinux-lib

Leon-hk:
- Write down outcome of the last team meeting (20.01.)
- Define the scope of the `python-gardenlinux-lib`
- I've added some really strict rules to make decisions easer, please review them and comment if you agree with them

### DIFF
--- a/docs/architecture/decisions/0030-python-gardenlinux-lib.md
+++ b/docs/architecture/decisions/0030-python-gardenlinux-lib.md
@@ -26,7 +26,7 @@ To provide a basis for these discussions, this ADR defines the scope of the `pyt
 ## Consequences
 
 - The only intended group using `python-gardenlinux-lib` are Garden Linux developers
-- The `python-gardenlinux-lib` does not need to have an official release process or official support
+- The `python-gardenlinux-lib` is not required to have an official release schedule or official support
 - Scripts targeted to users may contain duplicated code and have to be tested separately 
 - Pipeline scripts are stored indirectly
     - Therefore, if no version was specified, it will always use the latest script version and not the script version of the potentially older branch


docs/architecture/decisions/0030-python-gardenlinux-lib.md
```diff
@@ -0,0 +1,31 @@
...
+
+Therefore, pipeline and `/hack` scripts from the main repository have been refactored and moved to the `python-gardenlinux-lib`. In order to make the usage in pipelines still feasible, a GitHub Action to install the library is provided.
+
+However, as the `python-gardenlinux-lib` grows more and more, some scripts caused discussions about whether they fit into the scope of the `python-gardenlinux-lib` or not.
+
+To provide a basis for these discussions, this ADR defines the scope of the `python-gardenlinux-lib`, which should be taken into account, as long as there is no other ADR deprecating this one.
+
+## Decision
+
+- The `python-gardenlinux-lib` is an internal tool for the development and operations of GardenLinux, **not** a product
```
fwilhe: minor/style: The accepted spelling of gl is "Garden Linux", two words, both capitalised

docs/architecture/decisions/0030-python-gardenlinux-lib.md
```diff
@@ -0,0 +1,31 @@
...
+- Python scripts in this scope **should** be included in `python-gardenlinux-lib`, but do **not have to**, if a valid reason is provided
+- Python scripts used for pipelines, usually stored at `gardenlinux/.github/workflows`, **must always** be moved to the `python-gardenlinux-lib`
+- If the target group includes the users of Garden Linux, even if developers are the main target, the python script **must not** be included in `python-gardenlinux-lib`
+
+## Consequences
+
```
fwilhe, line 27: to add: how does this affect maintenance of releases (i.e. running the pipelines from rel-branches)

docs/architecture/decisions/0030-python-gardenlinux-lib.md
```diff
@@ -0,0 +1,31 @@
...
+- The only group using `python-gardenlinux-lib` are Garden Linux developers
```
fwilhe: We don't know, it is open source, people might use it for different purposes, but we don't want to support this. also I think this should be added to the main README of the lib (can you create a pr for this too, @Leon-hk ?)

docs/architecture/decisions/0030-python-gardenlinux-lib.md
Leon-hk: I've added a sentence, I hope it describes what you meant. Are there other consequences regarding this which I missed?

docs/architecture/decisions/0030-python-gardenlinux-lib.md
Leon-hk: I've corrected the sentence and proposed a small hint https://github.com/gardenlinux/python-gardenlinux-lib/pull/297

docs/architecture/decisions/0030-python-gardenlinux-lib.md
```diff
@@ -0,0 +1,32 @@
...
+
+- The `python-gardenlinux-lib` is an internal tool for the development and operations of Garden Linux, **not** a product
+- Python scripts in this scope **should** be included in `python-gardenlinux-lib`, but do **not have to**, if a valid reason is provided
+- Python scripts used for pipelines, usually stored at `gardenlinux/.github/workflows`, **must always** be moved to the `python-gardenlinux-lib`
+- If the target group includes the users of Garden Linux, even if developers are the main target, the python script **must not** be included in `python-gardenlinux-lib`
+
+## Consequences
+
+- The only intended group using `python-gardenlinux-lib` are Garden Linux developers
+- The `python-gardenlinux-lib` does not need to have an official release process or official support
```
ByteOtter: Maybe just a nitpick, but I don't think I agree with the wording here.
A release *process* would be good. As in a written down list of steps to take when creating a new release.
I think what we don't need though is a release schedule with maintenance cycles, milestones etc. But maybe I just interpret too much.

Leon-hk: That's a totally valid point. I meant that the lib is not required to have a process as there are no "end user" users, but it surely can still have one to make things more convenient. I've changed the wording a little bit and adjusted it to be *schedule*

ByteOtter: Yep that sounds much better, thank you! :)

PR #4262, opened 2026-02-05
firewall service compliance SRG-OS-000480-GPOS-00232

saubhikdattagithub:
Add a test_network to check for firewall service presence and active, for STIG compliance 

Fixes [160](https://github.com/gardenlinux/security/issues/160)

**Observations:**
pytest.mark.security_id(SRG-OS-000480-GPOS-00232) / pytest.mark.security_id(00232)
fails below:

### DIFF
--- a/tests/test_network.py
+++ b/tests/test_network.py
@@ -127,7 +127,7 @@ def test_that_nftables_firewall_service_is_running(systemd: Systemd):
 
 @pytest.mark.root(reason="Required to query systemd units")
 @pytest.mark.booted(reason="firewall service check requires booted system")
-@pytest.mark.feature("lima and gardener and server and ssh and fedramp ")
+@pytest.mark.feature("lima and gardener and server and ssh")
 def test_that_iptables_firewall_service_is_running(systemd: Systemd):
     """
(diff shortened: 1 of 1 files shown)


tests/test_network.py
```diff
@@ -108,3 +107,23 @@ def test_hostname_azure(shell):
...
         execution_time <= 10
     ), f"nslookup should not run into timeout: {result.stderr}"
+
+
+@pytest.mark.root(reason="Required to query systemd units")
+@pytest.mark.booted(reason="firewall service check requires booted system")
+@pytest.mark.feature(
+    "not gardener and not lima and not metal and not azure and not aws and not gcp and not gdch and not ali"
+)
+def test_firewall_nftables(systemd: Systemd):
```
Akendo: We should document in here the requierment. Writing it into a [docstring](https://peps.python.org/pep-0257/#multi-line-docstrings) why we have this need to separate between `iptables` and `nftables`.

Akendo: Also, we should rename the test to be more descriptive: `test_firewall_nftables` is generic. I would suggest that we call it `test_that_nftable_firewall_service_is_running`

PR #4238, opened 2026-02-02
ADR:  Static Feature/Test Coverage Analysis for Test Framework

yeoldegrove:
**What this PR does / why we need it**:

Outlined in https://github.com/gardenlinux/gardenlinux/issues/4233 and https://github.com/gardenlinux/gardenlinux/issues/4232.

**Which issue(s) this PR fixes**:
Fixes #4233

**Definition of Done:**
- [x] Write ADR documenting the static coverage analysis approach
- [x] Document the problem: variety of feature formats make it hard to track test coverage
- [x] Document the solution: setting identifiers and static analysis approach
- [x] Document consequences and trade-offs
- [x] Review ADR with the team
- [x] Address any feedback or concerns
- [ ] Update ADR status from "Draft" to "Accepted"

PR #4267, opened 2026-02-05
ADR: Garden Linux canonical name structure

### LINKED ISSUE #4263: Add ADR for Garden Linux canonical names
An ADR is required to align existing and new code to interpret correctly all possible variants of an Garden Linux canonical names.

**Why is this needed**:
While Garden Linux canonical names as a representation of flavors for specific architectures, versions and commits are commonly referred to an ADR is missing defining the parts of it.

NotTheEvilOne:
An ADR is required to align existing and new code to interpret correctly all possible variants of Garden Linux canonical names. This PR adds the ADR as 0032 as 0031 is in `Draft` state in PR #4238.

Closes: #4263

**Definition of Done:**
- [x] Write ADR documenting the expected Garden Linux canonical name structure
- [x] Document consequences
- [ ] Review ADR with the team
- [ ] Address any feedback or concerns
- [ ] Update ADR status from "Draft" to "Accepted"

### DIFF
--- a/docs/architecture/decisions/0032-garden-linux-canonical-name.md
+++ b/docs/architecture/decisions/0032-garden-linux-canonical-name.md
@@ -1,0 +1,30 @@
+# 32. Garden Linux canonical name structure
+
+Date: 2026-02-26
+
+## Status
+
+Draft
+
+## Context
+
+While Garden Linux canonical names (short `cname`) as a representation of flavors for specific architectures, versions and commits are commonly referred to an ADR is missing defining the parts of it. Historically the `cname` contains the platform, zero or more features, the architecture and version information. The platform and features part is referred to as "flavor".
+
+## Decision
+
+- `cname` contains exactly one platform as of [ADR 0020](0020-enforce-single-platform-by-default-in-builder.md) as long as it is not superseded.
+- The platform should but do not have to be at the first position of the flavor string.
+- Features are divided with `-`. `_` is used alternately as stated below.
+- Feature flags are prefixed with `_`. The `-` is omitted in this case.
+- Features are typically sorted by alphabet or a dependency and feature type based algorithm.
+- All features combined form a "flavor".
+- Appended to a "flavor" are in exact the following order, divided by a `-`:
+  - Architecture
+  - Semantic version string or `today` for local builds
+  - Commit ID (currently configured as first eight characters of an Git commit hash) or `local` for local build
+- Version information can be read from `COMMIT` and `VERSION` files of the `gardenlinux` Git repository to form the `cname` as described above.
+
+## Consequences
+
+- To ensure compatibility between tooling the `python-gardenlinux-lib` based tools should be used if applicable. The `python-gardenlinux-lib` contains unit-tests to ensure and conform to this ADR.
+- Additional generators and parsers must ensure the structure and all parts mentioned in this ADR are provided in the correct order.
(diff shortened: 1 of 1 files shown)


docs/architecture/decisions/0032-garden-linux-canonical-name.md
```diff
@@ -0,0 +1,30 @@
...
+Draft
+
+## Context
+
+While Garden Linux canonical names (short `cname`) as a representation of flavors for specific architectures, versions and commits are commonly referred to an ADR is missing defining the parts of it. Historically the `cname` contains the platform, zero or more features, the architecture and version information. The platform and features part is referred to as "flavor".
+
+## Decision
+
+- `cname` contains exactly one platform as of [ADR 0020](0020-enforce-single-platform-by-default-in-builder.md) as long as it is not superseded.
+- The platform should but do not have to be at the first position of the flavor string.
```
nkraetzschmar, line 16: If the platform occurs in the flavor string it absolutely has to be in first position. Only exception might be if an element is pulling in the platform via a feature include it thus gets elided => not appearing at first position (arguable if this edge case should explicitly be forbidden). But there should never be a case where the platform appears in the string and is not at first position.

docs/architecture/decisions/0032-garden-linux-canonical-name.md
```diff
@@ -0,0 +1,30 @@
...
+- Features are divided with `-`. `_` is used alternately as stated below.
+- Feature flags are prefixed with `_`. The `-` is omitted in this case.
+- Features are typically sorted by alphabet or a dependency and feature type based algorithm.
```
nkraetzschmar: are sorted by:
1. type: platforms, then elements, then flags
2. alphabetical

dependencies should not be relevant since the cname by design should not contain any features that are already given by includes (minimal form) => no features in the cname should have an include dependency on each other.

docs/architecture/decisions/0032-garden-linux-canonical-name.md
```diff
@@ -0,0 +1,30 @@
...
+- All features combined form a "flavor".
+- Appended to a "flavor" are in exact the following order, divided by a `-`:
+  - Architecture
+  - Semantic version string or `today` for local builds
+  - Commit ID (currently configured as first eight characters of an Git commit hash) or `local` for local build
+- Version information can be read from `COMMIT` and `VERSION` files of the `gardenlinux` Git repository to form the `cname` as described above.
```
nkraetzschmar: this should not come directly from the `VERSION` file. Instead the intermediary `./get_version` should be used to allow for arbitrary version formatting logic. Same for commit => local is not enforced to be the only valid non commit hash option here.

docs/architecture/decisions/0032-garden-linux-canonical-name.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Consequences
+
+- To ensure compatibility between tooling the `python-gardenlinux-lib` based tools should be used if applicable. The `python-gardenlinux-lib` contains unit-tests to ensure and conform to this ADR.
+- Additional generators and parsers must ensure the structure and all parts mentioned in this ADR are provided in the correct order.
```
nkraetzschmar, line 30: additional consequence: this will also influence in-place updates, as there the assumption is also that a cname does not contain the commit.

nkraetzschmar (changes requested):
I feel this is missing the point of a cname. The idea behind the cname is that it is the canonical shortest form that uniquely identifies a build target. Build target in this case being a target that the builder is able to produce.

yeoldegrove:
superseded by https://github.com/gardenlinux/gardenlinux/pull/4945

PR #4268, opened 2026-02-05
doc: adr for builder to glci interface spec

nkraetzschmar:
This ADR defines the intended interface between the gardenlinux image build and the meta data consumed by GLCI. This interface is intended as the spec that any upload logic (e.g. the python-gardenlinux-lib) should follow.

### DIFF
--- a/docs/architecture/decisions/0031-builder-glci-interface.md
+++ b/docs/architecture/decisions/0031-builder-glci-interface.md
@@ -103,22 +103,22 @@ Exactly one metadata document is produced per build, referred to as the *singles
 
 The metadata document contains the following information:
 
-| Metadata Key        | Source                                           |
-| ------------------- | ------------------------------------------------ |
-| `platform`          | `GARDENLINUX_PLATFORM`                           |
-| `platform_variant`  | `GARDENLINUX_PLATFORM_VARIANT` (if present)      |
-| `architecture`      | `arch` from requirements file                    |
-| `version`           | `GARDENLINUX_VERSION`                            |
-| `gardenlinux_epoch` | Major version component of `GARDENLINUX_VERSION` |
-| `build_committish`  | `GARDENLINUX_COMMIT_ID_LONG`                     |
-| `build_timestamp`   | Filesystem timestamp of the release file         |
-| `modifiers`         | Parsed list from `GARDENLINUX_FEATURES`          |
-| `require_uefi`      | `uefi` from requirements file                    |
-| `secureboot`        | `secureboot` from requirements file              |
-| `tpm2`              | `tpm2` from requirements file                    |
-| `paths`             | List of uploaded artifact descriptors            |
-| `s3_bucket`         | Target S3 bucket                                 |
-| `s3_key`            | S3 path of the metadata document                 |
+| Metadata Key        | Type            | Source                                           |
+| ------------------- | --------------- | ------------------------------------------------ |
+| `platform`          | string          | `GARDENLINUX_PLATFORM`                           |
+| `platform_variant`  | string          | `GARDENLINUX_PLATFORM_VARIANT` (if present)      |
+| `architecture`      | string          | `arch` from requirements file                    |
+| `version`           | string          | `GARDENLINUX_VERSION`                            |
+| `gardenlinux_epoch` | int             | Major version component of `GARDENLINUX_VERSION` |
+| `build_committish`  | string          | `GARDENLINUX_COMMIT_ID_LONG`                     |
+| `build_timestamp`   | timestamp[^1]   | Filesystem timestamp of the release file         |
+| `modifiers`         | List of strings | Parsed list from `GARDENLINUX_FEATURES`          |
+| `require_uefi`      | bool            | `uefi` from requirements file                    |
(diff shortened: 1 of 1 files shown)


docs/architecture/decisions/0031-builder-glci-interface.md
```diff
@@ -0,0 +1,176 @@
...
+The task of the upload logic is:
+
+- to upload files to S3, and
+- to translate builder output into metadata understood by GLCI.
+
+It must **not**:
+
+- enforce additional assumptions about structure,
+- reinterpret semantics,
+- or impose constraints beyond what is explicitly defined here.
```
NotTheEvilOne: Please refer to ADRs in general or "this and subsequent ADRs" instead of only this document. This would allow to revise needed changes / update without deprecating this ADR in general. See [RFC 9519 ](https://datatracker.ietf.org/doc/html/rfc9519) as an example: `[...] This document updates RFCs 4250, 4716, 4819, and 8308.`

docs/architecture/decisions/0031-builder-glci-interface.md
```diff
@@ -0,0 +1,176 @@
...
+```yaml
+- name: <artifact_base_name>.<artifact_file_extension>
+  suffix: .<artifact_file_extension>
+  md5sum: <md5sum_of_artifact>
+  sha256sum: <sha256sum_of_artifact>
+  s3_key: <path_to_uploaded_artifact_object>
+  s3_bucket_name: <s3_bucket>
+```
+
+No additional interpretation, normalization, or validation of these values is performed beyond format translation.
```
NotTheEvilOne: Please remove "validation" from here. As third-party input is handled by both GLCI and the above mentioned `python-gardenlinux-lib` input validation is an fundamental programming principle we can and should not prohibit.

docs/architecture/decisions/0031-builder-glci-interface.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Explicit Non-Responsibilities
+
+- Enforcing assumptions about naming conventions beyond documented extraction rules.
+- Inferring architecture from the canonical name or any other source than the requirements file.
+- Inferring platform semantics from feature graphs.
+- Validating builder output beyond presence of required fields.
+- Correcting, normalizing, or rejecting values based on internal heuristics.
+
+The upload logic is a **translator**, not a validator or policy engine.
```
NotTheEvilOne: Please remove or alter the last three lines. As described above an ADR should not forbid in any case core principles of security-aware programming including but not limiting to input validation.

NotTheEvilOne (changes requested):
I like the intention of this ADR even thought it seems to be too restrictive / focused on the current state without allowing even formally documented improvements / changes.

docs/architecture/decisions/0031-builder-glci-interface.md
nkraetzschmar: I see your point regarding security best practices, but I also see the following issue which led me to include this: if at some point we change the structure/format of any of the values here (obviously this should then be an ADR extending this) then assuming the upload logic does not validate/normalize only the build stage and GLCI would need to be updated accordingly. If the upload stage parses values then it too will always need to be adjusted in any such case.

So while I get the security aspect, which admittedly is likely the more important aspect here, I also see some value in keeping any upload logic somewhat generic to make future changes easier.

But I'd say we can drop this here => leaving wether to do validation/normalization open to the implementation for now; and in a subsequent ADR decide which direction we should go. Since both options imply a tradeoff it's probably a good idea to move it to its own decision.

docs/architecture/decisions/0031-builder-glci-interface.md
```diff
@@ -0,0 +1,172 @@
...
+```yaml
+- name: <artifact_base_name>.<artifact_file_extension>
+  suffix: .<artifact_file_extension>
+  md5sum: <md5sum_of_artifact>
+  sha256sum: <sha256sum_of_artifact>
+  s3_key: <path_to_uploaded_artifact_object>
+  s3_bucket_name: <s3_bucket>
+```
+
+No additional interpretation or normalization of these values is performed beyond format translation.
```
NotTheEvilOne, line 134: I think we need to remove `normalization` here as we currently remove S3 artifact unsupported characters from values before generating the YAML. The upload layer seems to be the right point to do these replacements / escaping.

docs/architecture/decisions/0031-builder-glci-interface.md
```diff
@@ -0,0 +1,174 @@
...
+
+## 2. Metadata Assembly
+
+Exactly one metadata document is produced per build, referred to as the *singles metadata*.
+
+### 2.1 Metadata Fields
+
+The metadata document contains the following information:
+
+| Metadata Key        | Type            | Source                                           |
```
NotTheEvilOne, line 106: Since the beginning of the [`upload_s3.sh`](https://github.com/gardenlinux/gardenlinux/blob/2932651b257814d77ac3719d0a2c85bcf457176d/.github/workflows/upload_to_s3.sh) we had a metadata key named "logs". I was thinking if we would like to split or additionally fill this key with life adding the same artifact descriptors as in `paths` if it is a log. What do you think?

PR #4277, opened 2026-02-09
feat: add test coverage markers to all features

yeoldegrove:
**What this PR does / why we need it**:

Add identifiers to all features following the format `GL-TESTCOV-<feature>-<category>-<setting-description>`.
Documentation on how to set the identifiers is also added.

**Which issue(s) this PR fixes**:
Fixes #4235

**Notes for reviewers**:

To report the feature/test coverage we need the script from https://github.com/gardenlinux/gardenlinux/pull/4276 (merged first).

PR #4278, opened 2026-02-09
feat: add test coverage markers all test files

### LINKED ISSUE #4236: Add setting identifier references to all test files
**Description:**

Update all test files in `tests-ng/test_*.py` to reference the setting identifiers defined in features. This establishes the mapping between features and tests for coverage analysis.

**Tasks:**

- [ ] Run coverage script to identify all setting IDs defined in features
- [ ] Add setting ID references to test files
- [ ] Ensure all referenced setting IDs follow the `GL-SET-*` pattern
- [ ] Run coverage script to verify references are detected
- [ ] Update test documentation to explain setting ID usage

yeoldegrove:

Update all test files in tests-ng/test_*.py to reference the test coverage markers defined in features. This establishes the mapping between features and tests for coverage analysis.

#### Test Coverage after This PR:

```
  Total markers:              574
  Tested markers:              97
  Untested:                   477
  Orphaned markers:             0

  Coverage:                 16.9%

  Total test functions:       173
  Tests with markers:           0
  Tests without markers:      173
```

**Coverage statistics explained:**

- **Total markers**: Number of `GL-TESTCOV-*` markers defined in feature files
- **Tested markers**: Markers that have corresponding tests with matching `@pytest.mark.testcov()` decorators
- **Untested**: Markers without corresponding tests
- **Orphaned markers**: Test markers referenced in tests but not defined in features (indicates potential typos or outdated references)
- **Tests with markers**: Test functions that include `@pytest.mark.testcov()` decorators (currently 0, as tests use markers but the reporting shows them differently)
- **Tests without markers**: Test functions without explicit marker decorators

Fixes #4236

**Notes for reviewers**:

- To report the feature/test coverage we need the script from https://github.com/gardenlinux/gardenlinux/pull/4276 (merged first).
- To match the feature identifiers, we need https://github.com/gardenlinux/gardenlinux/pull/4277 merged first.


tests/DEVELOPER-MARKER-GUIDE.md
```diff
@@ -145,8 +145,390 @@ _fips:
 
 ---
 
+## Marker Schema
+
+All setting ID markers follow this format:
+
+```
+GL-SET-$feature-$category-$description
```
vivus-ignis: Since we need these markers just for the coverage tool to report on a numeric KPI, I suggest to drop the feature, category and description parts. This will simplify the labeling process (no need to read another document) and will not pollute configuration files with redundant information, while feature name can be deduced from a config file path.
And to make the label matching work in the pytest world, we can use UUIDs that are easy to generate and unique -- which is, I would argue, is what we need.

I'm also not sure that "GL-SET" also explains the purpose of a label. I suggest something like GL-TESTCOV

Thus, a label would look like

```
# GL-TESTCOV-16180f6d-4493-45b7-9d5e-16105d2ce811
```

and 

```
@pytest.mark.setting_ids([
  "16180f6d-4493-45b7-9d5e-16105d2ce811"
```

in python code.

Unique IDs also help with nicer search (e.g. with something like `grep -A 5`), because they will point at the exact single place in features directory where we need test coverage.

yeoldegrove: > I'm also not sure that "GL-SET" also explains the purpose of a label. I suggest something like GL-TESTCOV

@NotTheEvilOne already raised this in https://github.com/gardenlinux/gardenlinux/pull/4238#discussion_r2755468050 
I would be fine with `GL-TESTCOV` or `GL-SET` but only having `GL-TEST` is to confusing for me.

yeoldegrove: > Since we need these markers just for the coverage tool to report on a numeric KPI, I suggest to drop the feature, category and description parts. This will simplify the labeling process (no need to read another document) and will not pollute configuration files with redundant information, while feature name can be deduced from a config file path.
> And to make the label matching work in the pytest world, we can use UUIDs that are easy to generate and unique -- which is, I would argue, is what we need.

I understand the "beauty" if the UUID suggestion here but I am not convinced that this is the right solution for the problem.

An example would be this `file.influce.ids.yaml`:

https://github.com/gardenlinux/gardenlinux/pull/4277/changes#diff-359b7abed3f0f6b595ef9c596762c94716e6a9e6295d0801653daee5ecf90667R1-R7

```
# file.influce.ids.yaml`
"setting_ids":
  "etc/cloud/cloud.cfg.d/01_debian-cloud.cfg":
  - "GL-SET-ali-config-cloud-apt-sources"
  - "GL-SET-ali-config-cloud-user-name"
  - "GL-SET-ali-config-cloud-user-shell"
  - "GL-SET-ali-config-cloud-user-lock-passwd"
  - "GL-SET-ali-config-cloud-user-sudo"
```

Each of the list items identifies a setting that is supposed to have a check. All settings are defined in a single config file. If we would add UUIDs here we would have no way of understanding which setting is actually meant.

vivus-ignis: Why do we need settings identifiers and additional configs in the first place?

> > Since we need these markers just for the coverage tool to report on a numeric KPI, I suggest to drop the feature, category and description parts. This will simplify the labeling process (no need to read another document) and will not pollute configuration files with redundant information, while feature name can be deduced from a config file path.
> > And to make the label matching work in the pytest world, we can use UUIDs that are easy to generate and unique -- which is, I would argue, is what we need.
> 
> I understand the "beauty" if the UUID suggestion here but I am not convinced that this is the right solution for the problem.
> 
> An example would be this `file.influce.ids.yaml`:
> 
> https://github.com/gardenlinux/gardenlinux/pull/4277/changes#diff-359b7abed3f0f6b595ef9c596762c94716e6a9e6295d0801653daee5ecf90667R1-R7
> 
> ```
> # file.influce.ids.yaml`
> "setting_ids":
>   "etc/cloud/cloud.cfg.d/01_debian-cloud.cfg":
>   - "GL-SET-ali-config-cloud-apt-sources"
>   - "GL-SET-ali-config-cloud-user-name"
>   - "GL-SET-ali-config-cloud-user-shell"
>   - "GL-SET-ali-config-cloud-user-lock-passwd"
>   - "GL-SET-ali-config-cloud-user-sudo"
> ```
> 
> Each of the list items identifies a setting that is supposed to have a check. All settings are defined in a single config file. If we would add UUIDs here we would have no way of understanding which setting is actually meant.

yeoldegrove: > Why do we need settings identifiers and additional configs in the first place?

I do not get your point. Could you elaborate on that?

vivus-ignis: I understand that we need to know the ratio of configuration settings [units] in features that are not covered by tests. To achieve this goal, there's no need for a full blown process, documentation and complex naming scheme. 

We need to get a number.

To get a number we need to match something in configuration [units] with something in unit tests. Unique IDs can easily solve this need. Benefit: simple workflow, no additional process (and a process WILL be broken, will incite discussions in PRs for every small change -- as soon as real people will start using that process), no additional documentation. 

My opinion is that this bureaucracy does not help to achieve the goal of measuring test coverage.

yeoldegrove: Please have a look at my reasoning why a UUID or a simple number does not work with the model: https://github.com/gardenlinux/gardenlinux/pull/4278#discussion_r2813099477
TLDR: A single file contains multiple settings and we cannot annotate IDs it in the file (as they will end up in the image).

vivus-ignis: There is no need for separate config files. I propose to comment the actual config files. What's the problem of comments ending up in image files, do we really have such a restriction? This is the same concept as comments for linters or similar tools, there is nothing wrong with committing code that has such comments.

yeoldegrove: Adding comments to files like these "won't work" and/or feels unnatural. These might be exceptions but we have those:
```
features/sap/file.include/usr/local/share/ca-certificates/SAP_Global_Root_CA.crt
features/sap/file.include/etc/motd
```

IMHO having these comments in our build artifacts feels wrong and does not justify the feature/test coverage use case.

If the team feels different, we can go for it :shrug:

tests/plugins/systemd.py
```diff
@@ -142,6 +142,57 @@ def is_active(self, unit_name: str) -> bool:
 
         return active
 
+    def is_inactive(self, unit_name: str) -> bool:
```
nkraetzschmar: Introducing new plugin functionality and new tests using it directly in the PR adding the coverage markers might make future traceability harder. maybe this should go into its own PR 🤔

tests/test_sysctls.py
```diff
@@ -1,13 +0,0 @@
-import pytest
-
-
-@pytest.mark.booted
-def test_sysctl_rp_filter(sysctl):
```
nkraetzschmar, line 5: Why drop this test? This test is really crucial to gardener and a very hard requirement for us to keep

tests/test_cloud_init.py
```diff
@@ -119,27 +200,35 @@ def test_gdch_ntp_settings(parse_file: ParseFile):
...
+)
 @pytest.mark.feature("openstack")
 def test_openstack_datasource_list(parse_file: ParseFile):
     file = "/etc/cloud/cloud.cfg.d/50-datasource.cfg"
     config = parse_file.parse(file, format="yaml")
     assert config["datasource_list"] == ["ConfigDrive", "OpenStack", "Ec2"]
 
 
-@pytest.mark.feature("openstack and cloud")
-def test_openstack_disable_network_config(parse_file: ParseFile):
```
nkraetzschmar, line 130: is there a reason to drop this test? not sure if we need it, but maybe make dropping it a dedicated PR if we don't to clearly document that decision.

tests/test_selinux.py
```diff
@@ -1,16 +0,0 @@
-from typing import List
-
-import pytest
-from plugins.file import File
-
-
-@pytest.mark.feature("_selinux")
-@pytest.mark.booted(reason="requires a running kernel to read /proc/cmdline")
```
nkraetzschmar, line 8: We should keep selinux tests, unless they are moved somewhere else?

tests/test_aide.py
```diff
@@ -3,6 +3,53 @@
 from plugins.systemd import Systemd
 
 
+@pytest.mark.testcov(["GL-TESTCOV-aide-service-aide-check-unit"])
+@pytest.mark.feature("aide")
+def test_aide_check_unit_exists(file):
```
nkraetzschmar: See previous comment on maybe creating a separate PR for introducing new tests and plugins.

tests/plugins/systemd.py
yeoldegrove: Yes, you're right. I suspect this broke during a rebase. Will remove the new functionality and add it in https://github.com/gardenlinux/gardenlinux/pull/4280.

tests/test_sysctls.py
yeoldegrove: Tests from this file are moved to `tests/test_kernel_parameters.py` and `tests/test_dmesg.py`.

tests/test_cloud_init.py
yeoldegrove: Good catch! This broken while rebasing for the OpenstaCloud feature. Re-added it.

tests/test_selinux.py
yeoldegrove: This is moved to `tests/test_lsm.py`.

tests/test_aide.py
yeoldegrove: See previous, removed the new tests and functionality.

PR #4289, opened 2026-02-09
SRG-OS-000755-GPOS-00220 

saubhikdattagithub:
Validate that an inactivity timeout is enforced for shell sessions via TMOUT configuration.

Fixes [161](https://github.com/gardenlinux/security/issues/161)

### DIFF
--- a/tests/test_network.py
+++ b/tests/test_network.py
@@ -114,7 +114,12 @@ def test_hostname_azure(shell):
 @pytest.mark.feature(
     "not gardener and not lima and not metal and not azure and not aws and not gcp and not gdch and not ali"
 )
-def test_firewall_nftables(systemd: Systemd):
+def test_that_nftables_firewall_service_is_running(systemd: Systemd):
+    """
+    As per DISA STIG requirement, either nftables or iptables firewall service should be active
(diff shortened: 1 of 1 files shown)


Akendo (changes requested):
You've committed the changes from the previous PR into this. Please remove the network-specific changes.

tests/test_profile.py
```diff
@@ -47,3 +49,135 @@ def test_profile_autologout_openstackbaremetal(parse_file: ParseFile):
...
+
+
+@pytest.mark.feature("cloud and openstackbaremetal")
+def test_shell_tmout_is_configured_cloud(parse_file: ParseFile):
+    """
+    As per DISA STIG requirement, this test verifies that TMOUT variable is configured.
+    Ref: SRG-OS-000755-GPOS-00220
+    """
+    lines = parse_file.lines(TMOUT_FILE_CLOUD)
+    lines_list = [
```
Akendo, line 142: I think we have a better way of dealing with this. 

I think you can use:
```
assert re.fullmatch(r"TMOUT\s*=\s*\d+") in parse_file.lines("/etc/profile.d/50-autologout.sh")
```

Also, we should test this if the shell environment has this parameter present.

PR #4309, opened 2026-02-13
doc: adr for pythonlib update regression tests

nkraetzschmar:


yeoldegrove (changes requested):
I also would vote against having an ADR for this.
I really love having these regression tests but this is merely another test in the already existing test suite of "python-gardenlinux-lib".

PR #4310, opened 2026-02-16
Enable SSH drop-ins

maxmoehl:
**What this PR does / why we need it**: Enable ssh / sshd config drop-ins and fix some branch names in the CONTRIBUTING guide.

Fixes n/a

**Definition of Done:**

**Special notes for your reviewer**: I'm happy to split this into two PRs if that makes more sense.

**Release note**:

```feature user
Enable drop-ins for ssh and sshd config.
```

### DIFF
--- a/features/ssh/file.include/etc/ssh/sshd_config
+++ b/features/ssh/file.include/etc/ssh/sshd_config
@@ -41,3 +41,5 @@ AuthenticationMethods publickey
 # Supported HostKey algorithms by order of preference.
 HostKey /etc/ssh/ssh_host_ed25519_key
 HostKey /etc/ssh/ssh_host_rsa_key
+
+Include /etc/ssh/sshd_config.d/*.conf
--- a/features/ssh/file.include/etc/ssh/ssh_config
+++ b/features/ssh/file.include/etc/ssh/ssh_config
@@ -1,3 +1,5 @@
+Include /etc/ssh/ssh_config.d/*.conf
+
 Host *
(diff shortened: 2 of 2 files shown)


features/ssh/file.include/etc/ssh/sshd_config
```diff
@@ -41,3 +41,5 @@ AuthenticationMethods publickey
 # Supported HostKey algorithms by order of preference.
 HostKey /etc/ssh/ssh_host_ed25519_key
 HostKey /etc/ssh/ssh_host_rsa_key
+
+Inlcude /etc/ssh/sshd_config.d/*.conf
```
fwilhe:
```suggestion
Include /etc/ssh/sshd_config.d/*.conf
```

typo

maxmoehl:
> I appreciate the boy-scout rule of fixing stuff as you go, but tbh I would have preferred a separate PR for the code of conduct fix

No problem, I've split the PR -> #4320.

> Regarding the ssh config: can you elaborate on your scenario for when this is useful? I don't see an issue with that, I'd just like to understand the change better.

When deploying configs automatically it's usually easier to just drop a file into the corresponding `.d` directory instead of trying prepend / append to the original config which is especially tricky with ssh configs as the ordering is important and the `Match` blocks leak (if I understood the semantics correctly). Besides that I'm kinda used to it from debian and the presence of the `ssd_config.d` directory suggests that it works out of the box although it doesn't right now.

PR #4311, opened 2026-02-16
Fix regular expression to not only change a commented out line

### LINKED ISSUE #4305: iscsi feature does not set CHAP algorithms
**What happened:**

https://github.com/gardenlinux/gardenlinux/blob/81d2743eed1d6d23e675033e7960734e32e8890d/features/iscsi/exec.config#L4 is supposed to set CHAP algorithms.

**Brief description of Bug**

In the final `/etc/iscsi/iscsid.conf` in the image, `node.session.auth.chap_algs` is commented out.

```
#node.session.auth.chap_algs = SHA3-256,SHA256,SHA1,MD5
```
Why do we want to set MD5 at all?

moritzkirchnersap:
but actually remove the '#' at the beginning of the line to comment the line in

To make the algorithm default choice explicit and to allow any changes to the algorithm to be respected in the config

Fixes #4305 

**Definition of Done:**
- [x ] The code is sufficiently documented
- [x ] Shared the changes with the Team so everyone is aware
- [x ] The code is appropriately tested
- [x ] Checked if the code needs to be backported to release branches of maintained versions (perform the actual backport after the merge to `main`)

**Special notes for your reviewer**:e

**Release note**:

```bugifx developer

make default algorithm choice explicit and also, should one change the line to get the desired results and not be surprised by the config line still being commented out

```

### DIFF
--- a/features/iscsi/exec.config
+++ b/features/iscsi/exec.config
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # config change requested by gardener
 sed -i 's/^\(node.session.scan\).*/\1 = manual/' /etc/iscsi/iscsid.conf
-sed -i 's/^.\(node.session.auth.chap_algs\).*/\1 = MD5/' /etc/iscsi/iscsid.conf
+sed -i 's/^#\(node.session.auth.chap_algs\).*/\1 = SHA3-256,SHA256,SHA1,MD5/' /etc/iscsi/iscsid.conf
 touch /var/lib/open-iscsi-configured0
 
 # Delete initiatorname.iscsi


features/iscsi/exec.config
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # config change requested by gardener
 sed -i 's/^\(node.session.scan\).*/\1 = manual/' /etc/iscsi/iscsid.conf
-sed -i 's/^\(node.session.auth.chap_algs\).*/\1 = MD5/' /etc/iscsi/iscsid.conf
+sed -i 's/^.\(node.session.auth.chap_algs\).*/\1 = MD5/' /etc/iscsi/iscsid.conf
```
vivus-ignis: What are trying to do with that exactly one character match (.) before the pattern?

moritzkirchnersap: https://github.com/open-iscsi/open-iscsi/blob/master/etc/iscsid.conf#L80C1-L82C56

The config comes with a commented out line that looks like this:

`#node.session.auth.chap_algs = SHA3-256,SHA256,SHA1,MD5`

What we did so far was use sed to create this line:

`#node.session.auth.chap_algs = MD5`

And now I would like to make it into a regular line instead of a comment, hence I would remove the first character before the group if there is one.

`node.session.auth.chap_algs = MD5`

vivus-ignis: If you know there's always a "#" symbol, why use the dot? Imo it's confusing the intention, I would replace with `#` or `#?`

yeoldegrove: @moritzkirchnersap Why do we want to set it to `MD5` and not leave it with a "secure" default ? I am just curious...

moritzkirchnersap: From the linked config file:
```

# Valid values are MD5, SHA1, SHA256, and SHA3-256.
# The default is MD5.

```
I don't know anything about that apart from what it says there, I don't care, I just "fixed" the regular expression to the best of my knowledge 😅 

If we trust our tests we might as well change it to sha256 or whatever

yeoldegrove: I guess this is less about our tests but about whoever uses this feature :shrug: 

The way it is currently implemented the default of `#node.session.auth.chap_algs = SHA3-256,SHA256,SHA1,MD5` is set. Which sound more reasonable than downgrading to `MD5` only.

Also I remember some discussions to drop md5 from our image... this is also why this came up to my mind.

PR #4319, opened 2026-02-16
feat: drop /boot/vmlinux* from the EROFS

nkraetzschmar:

Don't include the plain vmlinux on `/boot` in `_usi` flavours. In these flavours only the UKI on the ESP is used, so the vmlinux on `/boot` isn't required.

Fixes https://github.com/gardenlinux/internal-docs/issues/102


sh4sm (approved):
Elegant solution. Tested this branch with a local build and _/boot_ no longer contains the _vmlinuz_ file. 👍

PR #4322, opened 2026-02-16
SRG-OS-000348-GPOS-00136 ausearch|aureport

saubhikdattagithub:
As per DISA STIG requirement, this test verifies the operating system provides an audit reduction capability.

Fixes [163](https://github.com/gardenlinux/security/issues/163)

### DIFF
--- a/tests/test_disastig.py
+++ b/tests/test_disastig.py
@@ -1,6 +1,6 @@
 import pytest
-
 from plugins.shell import ShellRunner
+
 
 @pytest.mark.feature("not container")
 @pytest.mark.booted(reason="audit tools check requires booted system")


tests/test_disastig.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from plugins.shell import ShellRunner
+
+
+@pytest.mark.booted(reason="audit tools check requires booted system")
+@pytest.mark.root(reason="required to execute privileged tools")
```
Akendo, line 7: Just make sure that we do not run this in the container flag.

tests/test_disastig.py
```diff
@@ -0,0 +1,26 @@
...
+def test_audit_reduction_capability(shell: ShellRunner):
+    """
+    As per DISA STIG requirement, this test verifies the operating system
+    provides an audit reduction capability.
+    Ref: SRG-OS-000348-GPOS-00136
+    """
+
+    output = shell(
+        cmd="ausearch -k ssh_login -ts recent | aureport --summary",
```
yeoldegrove, line 16: As you use `ausearch` and `aureport` multiple times now in several tests, maybe it is woth to write a plugin? I am not sure though how complex this would be.

saubhikdattagithub: Thanks, not sure, how many of such requirements we would come across using `ausearch` | `aureport` further. Will contact in case of generalisation

PR #4328, opened 2026-02-17
feat: drop /boot/vmlinux and /boot/initrd from root partition

nkraetzschmar:


nkraetzschmar:
after discussion with @5kt this would break the pxe use case, so closing for now and maybe we can later re-open something that does not break pxe.

Issue #4369, opened 2026-02-24
Cleanup: Remove old lima upload to gcp and delete bucket


fwilhe:
- [x] We don't need [this workflow](https://github.com/gardenlinux/gardenlinux/blob/main/.github/workflows/lima_upload_oidc.yml) anymore because lima images are part of the 'proper' publishing workflow since #3945 
- [x] Also, the gcp bucket should be cleaned up (@saubhikdattagithub )

saubhikdattagithub:
Okay, the GCP bucket is now empty and cleaned up.

<img width="1584" height="786" alt="Image" src="https://github.com/user-attachments/assets/a250a721-cefc-4e36-b5ae-7424c86e0270" />

Good to close the ticket, IMO now

PR #4370, opened 2026-02-24
Delete .github/workflows/lima_upload_oidc.yml

fwilhe:
Part of #4369

We don't need this workflow anymore since lima images are published as part of the 'proper' pipeline


saubhikdattagithub (approved):
Agreed and approved

PR #4373, opened 2026-02-24
Attempt to add has_permissions

saubhikdattagithub:
Add granular file permissions for octal and text representation check 

Fixes [162](https://github.com/gardenlinux/security/issues/162)

image

### DIFF
--- a/tests/plugins/file.py
+++ b/tests/plugins/file.py
@@ -390,6 +390,7 @@ class File:
 
         return special | mode
 
+
 @pytest.fixture
 def file() -> File:
     """Fixture providing the ``File`` helper for file metadata operations."""


tests/plugins/file.py
```diff
@@ -159,6 +159,66 @@ def has_mode(self, path: str | Path, mode: str) -> bool:
...
+
+        if len(permissions) != 9:
+            raise ValueError(
+                "Symbolic permissions must be 9 characters like 'rwxr-x---'"
+            )
+
+        perm_map = {"r": 4, "w": 2, "x": 1, "-": 0}
+
+        mode = 0
+        for i in range(3):  # user, group, other
```
Akendo: We might have values with 4 digits, since subit might be placed here as well.

tests/test_disastig_00097.py
```diff
@@ -44,10 +41,20 @@ def test_audit_tools_permissions(file):
...
+        if not file.has_permissions(path, "755"):
+            invalid.append(f"{path} ({file.get_mode(path)})")
 
     assert not invalid, (
-        "stigcompliance: audit tools must have 755 or stricter (700): "
+        "stigcompliance: audit tools must have 755: "
         f"{', '.join(invalid)}"
     )
+
+
```
Akendo, line 52: Don't we miss any type of mark here?

tests/plugins/file.py
```diff
@@ -329,6 +329,67 @@ def is_owned_by(self, path: str | Path, user: str, group: str) -> bool:
         """
         return self.get_user(path) == user and self.get_group(path) == group
 
+    def has_permissions(self, path: str | Path, permissions: str | int) -> bool:
+        actual_mode = stat.S_IMODE(Path(path).stat().st_mode)
+        expected_mode = self._normalize_permissions(permissions)
+        return actual_mode == expected_mode
+
+    def _normalize_permissions(self, permissions: str | int) -> int:
```
Akendo, line 337: Please add some doc string here for more context why this func is.

tests/plugins/file.py
```diff
@@ -329,6 +329,67 @@ def is_owned_by(self, path: str | Path, user: str, group: str) -> bool:
         """
         return self.get_user(path) == user and self.get_group(path) == group
 
+    def has_permissions(self, path: str | Path, permissions: str | int) -> bool:
```
Akendo, line 332: So I was thinking a lot about this code and must admit I think we can improve how this code has been designed. 

I think this code makes it harder than it should, since it solves the problem in a very typical manner. 

My idea would be to keep track of this by using a subclass here. We might have to add a optional  constructor to the `File` class, but I think the code should look like this:

```
class Permissions(File):
    def __init__(self, file):
        super().__init__(file)
        self.mode = stat.S_IMODE(self.file.stat().st_mode)
    def __repr__(self):
        return(str(self.mode))
```

This will return the mode as the default value:
 
```
Permissions("flavors.yaml")
420
```

This makes the API here much cleaner. And when we want to check for the permissions, we just need to use this object: 

```
testing_file = Permissions("flavors.yaml")
testing_file.has_permissions("rwx")
True
testing_file.has_permissions(755)
True
```

Both use cases aren't implemented for `has_permissions`, but it should be seen as an example.

saubhikdattagithub: Ok, checking on the review to simplify

tests/test_disastig_00097.py
saubhikdattagithub: Yes, missed it. Initially thought to discuss if the idea with the plugin changes are fine, then would finalise. 
Would split it into 2 PRs as discussed, one for plugin and another for test to keep it clean post agreement.

PR #4389, opened 2026-02-26
feat: update local-k8s-lima for multi-node setup

fwilhe:
Just for fun and learning and experimentation

Follow up for #4381


gehoern (approved):
Not a lima veteran but lgtm

PR #4391, opened 2026-02-27
add SRG-OS-000257-GPOS-00098

saubhikdattagithub:
As per DISA STIG it is required to verify the operating system protects audit tools and audit log files from unauthorized modification.This test verifies whether Audit log directory exists, Directory is owned by root, Directory is not group/world writable. 
Ref: SRG-OS-000257-GPOS-00098

Fixes [167](https://github.com/gardenlinux/security/issues/167)

### DIFF
--- a/tests/DEVELOPER.md
+++ b/tests/DEVELOPER.md
@@ -613,6 +613,77 @@ def test_example(systemd: Systemd):
 > [!NOTE]
 > Have a look at the [user documentation](README.md#debugging-tests) if you want to know how to view those debug logs when running tests.
 
+### Debugging Tests in a Booted VM
+
+When you need to debug tests on a fully booted system (QEMU or a cloud VM), you can use the `--dev` flag to:
+
+- Quickly sync the local test distribution and test files into a VM.
+- Re-run tests on demand (oneshot) or automatically on file changes.
+
+#### QEMU VM Workflow
+
+##### Start VM
+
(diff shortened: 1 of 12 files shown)


tests/test_disastig_00098.py
```diff
@@ -0,0 +1,74 @@
...
+    As per DISA STIG it is required to verify the operating system protects
+    audit log files from unauthorized modification.
+    This test verifies:
+      - All regular files in /var/log/audit
+      - Owned by root
+      - Not group/world writable
+    Ref: SRG-OS-000257-GPOS-00098
+    """
+
+    if not file.exists(AUDIT_LOG_DIR):
```
Akendo, line 55: This shouldn't be here; either we test it on a flavor that has the necessary folders or we do not run the test.

saubhikdattagithub: Yes, as checked only `feature/stig` and `feature/cis` has this directory /var/log/audit

tests/test_disastig_00098.py
```diff
@@ -0,0 +1,82 @@
...
+        pytest.skip(f"{AUDIT_LOG_DIR} does not exist")
```
vivus-ignis, line 56: Why we need this check?

saubhikdattagithub: actually, this was for local test as its available only for `features/stig` and in `features/cis`

Issue #4335, opened 2026-02-18
Broken sshd config reference in `exec.config` for GCP and GDCH images


ByteOtter:
**What happened:**
 [`features/gcp/exec.config`](https://github.com/gardenlinux/gardenlinux/blob/bcaf09abe65325b8e8e35197cd55796b1c484920/features/gcp/exec.config) adds lines regarding `AuthorizedKeyCommand` and `AuthorizedKeyCommandUser` to a file called `sshd_conf`.
This file exists then in the image, but is not a valid sshd config.

We discovered this while working on https://github.com/gardenlinux/gardenlinux/pull/4237.

PR #4399, opened 2026-03-03
fix: wrong file path for sshd_config on GDCH

yeoldegrove:

GDCH feature uses wrong sshd config filename for oslogin configuration statements.
An extension of what was fixed in https://github.com/gardenlinux/gardenlinux/pull/4390.

Fixes https://github.com/gardenlinux/gardenlinux/issues/4335


ByteOtter (approved):
well spotted, thank you

PR #4403, opened 2026-03-04
feat: Add Debugging Tests in a Booted VM (QEMU and Cloud)

yeoldegrove:

This is an alternative approach to https://github.com/gardenlinux/gardenlinux/pull/4047 that does not need as many external dependencies and has a smaller footprint. It tries to solve the same issues and has a similar user experience. In addition to that it also works on Cloud VMs and chroot.

Fixes https://github.com/gardenlinux/gardenlinux/issues/3989, https://github.com/gardenlinux/gardenlinux/issues/3860 and replaces https://github.com/gardenlinux/gardenlinux/pull/3948 and https://github.com/gardenlinux/gardenlinux/pull/4047

**Definition of Done:**
- [x] Solution works both on x86 linux
- [ ] Solution works on arm64 macos (**TODO: verify**)
- [x] Solution works on QEMU and Cloud VMs
- [x] No modifications of GL VM images needed
- [x] No VM runners like vagrant or lima used, just plain qemu
- [x] Small footprint, no need to compile software, only inotify-tools, tar+ssh
- [x] No runtime rebuilds should be required if anything changes in tests directory
- [x] `inotifywait` stays in background and reacts on FS events to re-run tests on changes. It copies over files via tar+ssh.

**How to test:**

See [attached documentation](https://github.com/gardenlinux/gardenlinux/blob/ce9f00a616b691447c54553430dddb4307198c1b/tests/DEVELOPER.md#debugging-tests-in-a-booted-vm).

### DIFF
--- a/tests/DEVELOPER.md
+++ b/tests/DEVELOPER.md
@@ -613,6 +613,77 @@ def test_example(systemd: Systemd):
 > [!NOTE]
 > Have a look at the [user documentation](README.md#debugging-tests) if you want to know how to view those debug logs when running tests.
 
+### Debugging Tests in a Booted VM
+
+When you need to debug tests on a fully booted system (QEMU or a cloud VM), you can use the `--dev` flag to:
+
+- Quickly sync the local test distribution and test files into a VM.
+- Re-run tests on demand (oneshot) or automatically on file changes.
+
+#### QEMU VM Workflow
+
+##### Start VM
+
+Run the tests with `--dev` (or `--ssh --skip-cleanup --skip-tests --watch` for the long version) to start a QEMU VM in dev mode.
+
+```bash
+./test --dev .build/aws-gardener_prod-amd64-today-local.raw
+```
+
+You can also pass additional `--test-args`.
+
+```bash
+./test --dev --test-args "test_ssh.py -v" .build/aws-gardener_prod-amd64-today-local.raw
+```
+
+This will:
+
+- Start a QEMU VM.
+- Sync `.build/dist.tar.gz` into the VM.
+- Sync the current `tests/` tree.
+- Run `pytest` inside the VM with the provided `--test-args` once.
+- Wait for file changes in `tests/` and `features/`.
+- Re-sync the changed tests into the VM.
+- Re-run `pytest` with your given arguments after each change.
+
+You can stop watch mode at any time with `Ctrl+C`.
+
+#### Cloud VM Workflow
+
+##### Start VM
+
+Run the tests with `--dev` (or `--skip-cleanup --skip-tests --watch` for the lang version) to start a Cloud VM in dev mode.
+
+```bash
+./test --dev --cloud azure .build/azure-gardener_prod-amd64-today-local.raw
+# or use an already uploaded image
+./test --dev --cloud azure \
+    --cloud-image --image-requirements-file .build/azure-gardener_prod-amd64-today-local.requirements \
+    /CommunityGalleries/gardenlinux-13e998fe-534d-4b0a-8a27-f16a73aef620/Images/gardenlinux-nvme-gardener_prod-amd64/Versions/2150.0.0
+```
+
+You can also pass additional `--test-args`.
+
+```bash
+./test --dev --cloud azure --test-args "test_ssh.py -v" .build/aws-gardener_prod-amd64-today-local.raw
+```
+
+This will:
+
+- Deploy a Cloud VM.
+- Sync `.build/dist.tar.gz` into the VM.
+- Sync the current `tests/` tree.
+- Run `pytest` inside the VM with the provided `--test-args`.
+- Wait for file changes in `tests/` and `features/`.
+- Re-sync the changed tests into the VM.
+- Re-run `pytest` with your given arguments after each change.
+
+You can stop watch mode at any time with `Ctrl+C`.
+
+### Debugging Tests in the Chroot Environment
+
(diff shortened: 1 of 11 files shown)


tests/DEVELOPER.md
```diff
@@ -579,6 +579,103 @@ def test_example(systemd: Systemd):
...
+When you need to debug tests on a fully booted system (QEMU or a cloud VM), you can use the dev runner scripts to:
+
+- Quickly sync the local test distribution and test files into a VM.
+- Re-run tests on demand (oneshot) or automatically on file changes.
+
+#### QEMU VM Workflow
+
+##### Start VM
+
+Run the tests with `--ssh --skip-cleanup --skip-tests` to start a QEMU VM. This will simply boot the VM, enable ssh but not start any tests and keep your VM running until you hit `Ctrl+C`:
```
vivus-ignis: Why not to use a simple to remember `--dev` argument that will just do the same?

tests/DEVELOPER.md
```diff
@@ -579,6 +579,103 @@ def test_example(systemd: Systemd):
...
+- Sync `.build/dist.tar.gz` into the VM.
+- Sync the current `tests/` tree.
+- Run `pytest` inside the VM with the provided `--test-args`.
+
+##### Watch for file changes
+
+In a second shell, To automatically re-run tests when you change files under `tests/` or `features/`, just add `--watch`:
+
+```bash
+tests/util/run_dev_qemu.sh --test-args "test_ssh.py -v" --watch
```
vivus-ignis: It should be stated explicitly that the --watch mode relies on inotifywait, which may not be installed on the user system. For me it looked like this: the command above wrote "Waiting for first change to run tests..." and terminated right after.
I would show a proper error message if inotifywait is not found.

tests/DEVELOPER.md
```diff
@@ -579,6 +579,103 @@ def test_example(systemd: Systemd):
...
+- Re-run tests on demand (oneshot) or automatically on file changes.
+
+#### QEMU VM Workflow
+
+##### Start VM
+
+Run the tests with `--ssh --skip-cleanup --skip-tests` to start a QEMU VM. This will simply boot the VM, enable ssh but not start any tests and keep your VM running until you hit `Ctrl+C`:
+
+```bash
+./test --ssh --skip-cleanup --skip-tests .build/aws-gardener_prod-amd64-today-local.raw
```
vivus-ignis: Why not wrap `--ssh --skip-cleanup --skip-tests` into an easy to remember `--dev` argument?

tests/DEVELOPER.md
```diff
@@ -579,6 +579,103 @@ def test_example(systemd: Systemd):
...
+tests/util/run_dev_qemu.sh --test-args "test_ssh.py -v" --watch
+```
+
+The script will:
+
+- Wait for file changes in `tests/` and `features/`.
+- Re-sync the changed tests into the VM.
+- Re-run `pytest` with your given arguments after each change.
+
+You can stop watch mode at any time with `Ctrl+C`.
```
vivus-ignis, line 649: I think it would make sense from a UX experience if `--watch` will run tests once before starting to watch for changes. Or at least there should be an option to enable this behavior.

tests/DEVELOPER.md
yeoldegrove: I'll look into adding a `--dev` shortcut but would like to have the individual flags as well.

tests/DEVELOPER.md
yeoldegrove: Yepp...error handling is a good idea and can be added.

tests/DEVELOPER.md
yeoldegrove: I'll look into adding a `--dev` shortcut but would like to have the individual flags as well.

tests/DEVELOPER.md
yeoldegrove: Will look into this.

tests/DEVELOPER.md
yeoldegrove: @vivus-ignis took the approach you suggested and only a `./test --dev` is now needed to run/watch the tests. No seperate shell is needed.

tests/DEVELOPER.md
yeoldegrove: Added error handling.

tests/DEVELOPER.md
yeoldegrove: `--watch/dev` now runs the tests once.

tests/DEVELOPER.md
yeoldegrove: @vivus-ignis took the approach you suggested and only a ./test --dev is now needed to run/watch the tests. No seperate shell is needed.

tests/util/run_chroot.sh
```diff
@@ -41,6 +46,96 @@ log_file_junit="chroot.test.xml"
 
 mkdir -p "$log_dir"
 
+tmpdir=
+
+cleanup() {
+	get_logs
+	[ -z "$tmpdir" ] || [ ! -e "$tmpdir/chroot" ] || umount -l "$tmpdir/chroot" || rmdir "$tmpdir/chroot"
```
vivus-ignis, line 53: This is rather hard to unpack. I would rewrite it to make the intention clearer.

yeoldegrove: We might need a refactoring of those scripts at one point. The cleanup() function was just moved/not-touched in this PR.

PR #4408, opened 2026-03-05
audit plugin + STIG ID: SRG-OS-000466-GPOS-00210

vivus-ignis:
**What this PR does / why we need it**: we need to have a more abstract way to test against auditd rules to simplify mapping to STIG requirements. 

Fixes https://github.com/gardenlinux/security/issues/165 (https://stigaview.com/products/srg-gpos/v3r2/SRG-OS-000466-GPOS-00210/)


tests/plugins/audit.py
```diff
@@ -0,0 +1,129 @@
...
+        re.VERBOSE,
+    )
+    return [
+        syscall
+        for match in sc_re.findall(auditd_rule)
+        for syscall in match[0].split(",")
+    ]
+
+
+def _access_types_included(auditd_rule, access_types):
```
Akendo: Write a documentation string explaining the regex.

vivus-ignis: Explained the regex in the function's docstring.

tests/plugins/audit.py
```diff
@@ -0,0 +1,173 @@
...
+    results = [
+        match
+        for match in path_re.findall(auditd_rule)
+        if set(access_types).issubset(set(match[0]))
+    ]
+    return bool(results)
+
+
+@pytest.fixture
+def audit_rule():
```
Akendo, line 151: Can you turn this into a class?

vivus-ignis: Done

tests/plugins/audit.py
```diff
@@ -0,0 +1,169 @@
...
+from pathlib import Path
+
+import pytest
+
+logger = logging.getLogger(__name__)
+
+
+class AuditRule:
+    def __init__(self):
+        try:
```
Akendo, line 16: We should check whether the package is installed first.

vivus-ignis: We can certainly do that, but what value will this check have? Trying to call auditctl will raise an exception if the binary is not there or cannot be run, which in turn will fail the test and this will be a warning sign for anyone running tests. 

Testing for package to be present in the plugin will erode the single responsibility principle imo.

Akendo: That's a good question. The answer is simple: We ship the package; a Garden Linux user might have installed an `auditctl` on his system. Your error message suggests that the file wasn't found, but that the underlying package might be amiss isn't tested for.

vivus-ignis: If we need to test for presence of an auditd package, it should be a done in a separate unit test, not in a plugin. I think that putting this check here breaks the single responsibility idea.

vivus-ignis: Done

tests/integration/security/compliance/test_disastig_0210.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@pytest.fixture
+def sudoers_edit():
+    for line in fileinput.input("/etc/sudoers", inplace=True, backup=".bak"):
+        if not line.startswith("# Host alias specification"):
+            print(line, end="")
+    yield
+    shutil.copy2("/etc/sudoers.bak", "/etc/sudoers")
+    os.remove("/etc/sudoers.bak")
```
yeoldegrove, line 15: This needs to be a handler https://github.com/gardenlinux/gardenlinux/blob/main/tests/DEVELOPER.md#handlers-testshandlers.

Also, could be make it "more safe" by not using `/etc/sudoers` but somethig in `/etc/sudoers.d` ?

tests/integration/security/compliance/test_disastig_0210.py
```diff
@@ -0,0 +1,98 @@
...
+    for file in ["apparmor", "apparmor.d"]:
+        assert audit_rule(
+            fs_watch_path=f"/etc/{file}", access_types="wa"
+        ), f"stigcompliance: writing to or changing metadata of /etc/{file} should be audited"
+
+
+@pytest.mark.feature("not lima")
+@pytest.mark.booted(reason="audit rule validation requires running audit subsystem")
+@pytest.mark.root(reason="required to query audit logs")
+def test_attempt_to_delete_privileges_event_logged(audit_rule, shell, sudoers_edit):
```
yeoldegrove, line 86: This changes the system and needs to be marked https://github.com/gardenlinux/gardenlinux/blob/main/tests/DEVELOPER.md#pytestmarkmodifyreason.

tests/integration/security/compliance/test_disastig_0210.py
yeoldegrove: Or do we even really need to test this with such a sensitive file or could be use a dummy file?

vivus-ignis: > This needs to be a handler https://github.com/gardenlinux/gardenlinux/blob/main/tests/DEVELOPER.md#handlers-testshandlers.
> 
> Also, could be make it "more safe" by not using `/etc/sudoers` but somethig in `/etc/sudoers.d` ?

These tests are designed to be executed on a running system to assess compliancy.
We cannot trust sudoers not being manipulated to ignore sudoers.d, so I edit sudoers file itself.

vivus-ignis: > Or do we even really need to test this with such a sensitive file or could be use a dummy file?

We need to trigger a relevant audit event.

tests/integration/security/compliance/test_disastig_0210.py
vivus-ignis: Done.

PR #4413, opened 2026-03-06
docs(adr): Add retention policy for public cloud images

sh4sm:
Add an ADR for our retention policy for public cloud images. The amount of public cloud images you can provide on AWS for EC2 instances is limited and we just recently run into this limit. Hence, this ADR should ensure that we always have enough open slots for a release.

Related task: https://github.com/gardenlinux/gardenlinux/issues/4414


sh4sm:
Was accepted by the team during our team meeting (2026-04-14).

PR #4424, opened 2026-03-09
Add permissions to the file plugin

saubhikdattagithub:
What this PR does / why we need it:
Add granular file permissions for octal and text representation check (PLUGIN only)

Which issue(s) this PR fixes:
Fixes https://github.com/gardenlinux/security/issues/162


tests/plugins/file.py
```diff
@@ -334,3 +337,105 @@ def is_owned_by(self, path: str | Path, user: str, group: str) -> bool:
...
+
+PERMISSION_BITS = {
+    "r": 4,
+    "w": 2,
+    "x": 1,
+    "-": 0,
+}
+
+
+class Permissions(File):
```
vivus-ignis: I think that Permissions inheriting from File doesn't make sense: permissions are not a kind of a file.

tests/plugins/file.py
```diff
@@ -15,6 +15,9 @@ class File:
     existence, type, permissions, ownership, etc.
     """
 
+    def __init__(self, path: str | Path | None = None):
+        self.file = Path(path) if path else None
+
```
vivus-ignis: I think it's better to follow the convention already used in other plugins like "shell": arguments like the actual file path have to be passed to `__call__`.

tests/plugins/file.py
```diff
@@ -334,3 +337,105 @@ def is_owned_by(self, path: str | Path, user: str, group: str) -> bool:
...
+        Args:
+            file: Path to the file whose permissions will be evaluated.
+
+        Raises:
+            FileNotFoundError: If the provided path does not exist.
+            PermissionError: If the process lacks permission to stat the file.
+        """
+        super().__init__(file)
+        self.mode = stat.S_IMODE(self.file.stat().st_mode)
```
vivus-ignis: Same as my comment for File class: better to use `__call__` instead.

tests/plugins/file.py
```diff
@@ -334,3 +337,105 @@ def is_owned_by(self, path: str | Path, user: str, group: str) -> bool:
...
     """Fixture providing the ``File`` helper for file metadata operations."""
     return File()
+
+
+PERMISSION_BITS = {
+    "r": 4,
+    "w": 2,
+    "x": 1,
+    "-": 0,
+}
```
vivus-ignis: If you have a Permissions class, this has to be a part of it.

tests/plugins/file.py
vivus-ignis: There you would also read the permissions of a file.

tests/plugins/file.py
vivus-ignis: Andalso: no need to touch the filesystem here, it breaks your abstraction. Permissions class (by the way, a better name would be FilePermissions) should just represent permission bits.

tests/plugins/file.py
```diff
@@ -334,3 +337,105 @@ def is_owned_by(self, path: str | Path, user: str, group: str) -> bool:
...
+                - a symbolic string (e.g., ``"rw-r--r--"``).
+
+        Returns:
+            bool: ``True`` if the file's permissions match the expected
+            mode, otherwise ``False``.
+        """
+        expected = self._normalize(permissions)
+        return self.mode == expected
+
+    def _normalize(self, permissions: str | int) -> int:
```
vivus-ignis: The method name does not reflect what it does really. Better rename it to `__int__` so that this class would be used like this:
```
f = File("/etc/passwd")  

# File.__call__ would then do 
# self.mode = Permissions(stat.S_IMODE(self.file.stat().st_mode))

assert f.mode == Permissions("rwxrwr--")  
# here ^^^ __eq__ on Permissions object will be called 
# where you would call int(self) == int(other), 
# int() in turn calling __int__ on Permissions objects
```

tests/plugins/file.py
```diff
@@ -329,6 +329,116 @@ def is_owned_by(self, path: str | Path, user: str, group: str) -> bool:
...
+        Raises:
+        FileNotFoundError: If the path does not exist.
+        PermissionError: If permission to access the path is denied.
+        ValueError: If the permission specification is invalid.
+        """
+        actual_mode = stat.S_IMODE(Path(path).stat().st_mode)
+        expected_mode = self._normalize_permissions(permissions)
+        return actual_mode == expected_mode
+
+    def _normalize_permissions(self, permissions: str | int) -> int:
```
vivus-ignis, line 360: This function does not work as described in the docstring. When an integer 644 is passed to it, the return value is 644, not 420.

tests/plugins/file.py
```diff
@@ -329,6 +329,116 @@ def is_owned_by(self, path: str | Path, user: str, group: str) -> bool:
         """
         return self.get_user(path) == user and self.get_group(path) == group
 
+    def has_permissions(self, path: str | Path, permissions: str | int) -> bool:
+        """Check if a file has the specified permission mode.
+
+        This helper retrieves the current permission bits of the given path
+        and compares them with the expected permissions. The expected value
+        may be provided either as an octal representation (e.g., ``"0644"``
```
vivus-ignis:
```suggestion
        may be provided either as an octal **string** representation (e.g., ``"0644"``
```

vivus-ignis (changes requested):
Right now int values do not work as expected.

tests/plugins/file.py
saubhikdattagithub: Yes, this was comment for very initial work which had octal consideration, working on it

saubhikdattagithub: Edited the docstrings as suggested!

PR #4425, opened 2026-03-09
Add test for: SRG-OS-000256-GPOS-00097

saubhikdattagithub:
What this PR does / why we need it:
Add granular file permissions for octal and text representation checks test(TEST)

Which issue(s) this PR fixes:
Fixes [162](https://github.com/gardenlinux/security/issues/162)

### DIFF
--- a/tests/test_disastig_00097.py
+++ b/tests/test_disastig_00097.py
@@ -7,6 +7,7 @@ AUDIT_TOOL_PATHS = [
     "/usr/bin/journalctl",
 ]
 
+
 @pytest.mark.feature("not container")
 @pytest.mark.booted(reason="audit tools check requires booted system")
 @pytest.mark.root(reason="required to execute privileged tools")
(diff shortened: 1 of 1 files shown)


tests/test_disastig_00097.py
```diff
@@ -0,0 +1,88 @@
...
+        "stigcompliance: audit tools must have 755: " f"{', '.join(invalid)}"
+    )
+
+@pytest.mark.feature("not container")
+@pytest.mark.booted(reason="audit tools check requires booted system")
+@pytest.mark.root(reason="required to execute privileged tools")
+def test_sticky_bit_support(file, tmp_path):
+    """
+    As per DISA STIG requirement, the operating system must protect audit tools from unauthorized access.
+    Ref: SRG-OS-000256-GPOS-00097
```
Akendo, line 86: Since this does not contribute to the test, I feel we can remove the reference here.

saubhikdattagithub: As discussed, the has_permissions are now reverted to functions and test is adjusted

PR #4441, opened 2026-03-12
Add security mark for TMOUT test for SRG-OS-000029-GPOS-00010

saubhikdattagithub:
As per DISA STIG Compliance, it needs to be verified that the operating system initiates a session lock after a 15-minute period of inactivity for all connection types

Fixes #[175](https://github.com/gardenlinux/security/issues/175)

### DIFF
--- a/tests/test_profile.py
+++ b/tests/test_profile.py
@@ -1,7 +1,6 @@
 import re
 
 import pytest
-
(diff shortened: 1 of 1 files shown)


vivus-ignis (approved):
LGTM

PR #4448, opened 2026-03-13
STIG SRG-OS-000475-GPOS-00220

vivus-ignis:

Fixes https://github.com/gardenlinux/security/issues/177

### DIFF
--- a/tests/integration/security/compliance/test_disastig_00009.py
+++ b/tests/integration/security/compliance/test_disastig_00009.py
@@ -1,0 +1,18 @@
+import pytest
+
+
+@pytest.mark.feature("not container and not lima and not gardener and not baremetal")
+@pytest.mark.root(reason="required to verify PAM authentication enforcement")
+def test_session_lock_requires_reauthentication(lsm):
+    """
+    As per DISA STIG compliance requirements, the operating system must retain a
+    user's session lock until that user reestablishes access using established
+    identification and authentication procedures.
+    This test verifies that PAM configuration enforces authentication using
+    standard authentication modules and does not allow bypass mechanisms.
+    Ref: SRG-OS-000028-GPOS-00009
+    """
+
+    assert (
+        "selinux" in lsm
+    ), "stigcompliance: no LSM (AppArmor/SELinux) present to enforce session lock re-authentication"


tests/test_disastig_0220.py
```diff
@@ -0,0 +1,28 @@
...
+    )
+    assert (
+        "StandardOutput=journal" in result.stdout
+    ), f"sshd stdout not directed to journal: {result.stdout}"
+    assert re.search(
+        r"StandardError=(journal|inherit)", result.stdout
+    ), f"sshd stderr not directed to journal or inherit: {result.stdout}"
+
+
+def test_pam_unix_is_in_use():
```
Akendo: Please look at the PAM plugin we have.

tests/test_disastig_0220.py
```diff
@@ -0,0 +1,28 @@
+import re
+
+
+def test_sshd_unit_is_journald_friendly(shell):
+    """
+    The sshd.service unit must forward stdout/stderr to the journal.
+    """
+    result = shell(
+        cmd="systemctl show -p StandardOutput,StandardError sshd.service",
```
Akendo: Maybe you can add this to the systemd plugin?

tests/test_disastig_0220.py
vivus-ignis: Fixed

tests/test_disastig_0220.py
vivus-ignis: Done

tests/plugins/systemd.py
```diff
@@ -192,6 +193,15 @@ def wait_is_system_running(self) -> SystemRunningState:
         elapsed = time.time() - start_time
         return SystemRunningState(result.stdout.strip(), result.returncode, elapsed)
 
+    def get_unit_properties(self, service_name, *args) -> dict:
+        result = self._shell(
+            cmd=f"systemctl show -p {','.join(args)} {service_name}.service",
+            capture_output=True,
+        )
+        return {
+            k: v for kv in result.stdout.strip().split("\n") for k, v in [kv.split("=")]
```
yeoldegrove: Did you verify this works? Those lines might contains multiple `=`, e.g.:

```
❯ systemctl show wpa_supplicant.service | grep RootImagePolicy
RootImagePolicy=root=verity+signed+encrypted+unprotected+absent:usr=verity+signed+encrypted+unprotected+absent:home=encrypted+unprotected+absent:srv=encrypted+unprotected+absent:tmp=encrypted+unprotected+absent:var=encrypted+unprotected+absent
```

yeoldegrove: Maybe (did not thing much about it) we could use the existing parsers (or extend those)?

tests/plugins/systemd.py
```diff
@@ -192,6 +193,15 @@ def wait_is_system_running(self) -> SystemRunningState:
         elapsed = time.time() - start_time
         return SystemRunningState(result.stdout.strip(), result.returncode, elapsed)
 
+    def get_unit_properties(self, service_name, *args) -> dict:
```
yeoldegrove: Maybe name this simply `show_properties`? Also it would be nice if this would work with both  sockets/services and also without passing one of those and getting the global properties.

tests/plugins/systemd.py
vivus-ignis: Fixed.

tests/plugins/systemd.py
vivus-ignis: I would prefer to leave the name as it is because it matches the current naming: list_units, list_failed_units etc

tests/integration/security/compliance/test_disastig_0220.py
```diff
@@ -0,0 +1,67 @@
...
+Verify the operating system generates audit records for all direct access to the information system.
+"""
+
+
+@pytest.mark.booted(reason="requires running systemd")
+def test_journald_should_not_store_logs_in_memory(shell):
+    """
+    Confirm that journald is not configured to store logs in memory
+    """
+    result = shell(
```
yeoldegrove: Please add something in `tests/plugins/systemd.py` for the `cat-config`, too.
I can imagine many tests that will benefit and it will have way nicer parsing than what you're currently doing.

tests/plugins/systemd.py
```diff
@@ -267,6 +267,22 @@ def wait_is_system_running(self) -> SystemRunningState:
...
+        return {
+            k: v
+            for kv in result.stdout.strip().split("\n")
+            for k, v in [kv.split("=", 1)]
+        }
+
+    def get_config(self, config_path):
+        return self._shell(
+            f"systemd-analyze cat-config {config_path}", capture_output=True
+        ).stdout
```
yeoldegrove: This should already do the heavy lifting of parsing via existing parsing plugins. This way we can make nice and easy asserts in the actual tests.

vivus-ignis: This would be a violation of the single responsibility principle. Here one function does one simple thing, it's up to a user what to do next with the output.

yeoldegrove: The idea of plugins is to provide an interface to write easy tests. We had these discussions before and updated https://github.com/gardenlinux/gardenlinux/blob/main/tests/DEVELOPER.md#6-use-abstractions-judiciously-to-hide-implementation-details as a guideline.

Both `get_unit_properties` and `get_config` have to provide a return value that needs no interpretation but can just be compared against in an assert.

e.g. `assert result["StandardOutput"] == "journal"` from `get_unit_properties` is nice but this is bad and has to move to the plugin:

```
    result = systemd.get_config("/etc/systemd/journald.conf")
    config = [
        v
        for line in result.splitlines()
        if line.startswith("Storage=")
        for k, v in [line.split("=")]
    ]
    if config:
        assert config[-1] != "volatile"
```

The assert should be as simple as `assert result["Storage"] == "persistent"` or `assert not result["Storage"] == "volatile"` .

tests/integration/security/compliance/test_disastig_0220.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+
+
+@pytest.mark.booted(reason="requires running systemd")
+def test_journald_should_not_store_logs_in_memory(systemd):
+    """
+    Confirm that journald is not configured to store logs in memory
+    """
+    result = systemd.get_config("/etc/systemd/journald.conf")
+    if "Storage" in result.keys():
```
yeoldegrove, line 16: This is not ideal, as the test will never run if the key is missing (which should be a failed test?).

tests/plugins/systemd.py
```diff
@@ -267,6 +267,30 @@ def wait_is_system_running(self) -> SystemRunningState:
         elapsed = time.time() - start_time
         return SystemRunningState(result.stdout.strip(), result.returncode, elapsed)
 
+    def get_unit_properties(self, service_name, *args) -> dict:
```
yeoldegrove: Do we really need `args` or is it simpler to just always return the whole dict? If we pass args, it should be optional.

tests/plugins/systemd.py
```diff
@@ -267,6 +267,30 @@ def wait_is_system_running(self) -> SystemRunningState:
...
+        result = self._shell(
+            cmd=f"systemctl show -p {','.join(args)} {service_name}.service",
+            capture_output=True,
+        )
+        return {
+            k: v
+            for kv in result.stdout.strip().split("\n")
+            for k, v in [kv.split("=", 1)]
+        }
```
yeoldegrove: https://github.com/gardenlinux/gardenlinux/blob/448955531f6d93a928b4db77b8b11276130aaf5c/tests/plugins/parse.py#L339 with `keyval` could be used here to have this even shorter.

tests/plugins/systemd.py
```diff
@@ -267,6 +267,30 @@ def wait_is_system_running(self) -> SystemRunningState:
...
+            f"systemd-analyze cat-config {config_path}", capture_output=True
+        )
+        config_lines = [
+            line
+            for line in result.stdout.split("\n")
+            if not line.startswith("#")
+            if "=" in line
+        ]
+
+        return {k: v for kv in config_lines for k, v in [kv.split("=", 1)]}
```
yeoldegrove: https://github.com/gardenlinux/gardenlinux/blob/448955531f6d93a928b4db77b8b11276130aaf5c/tests/plugins/parse.py#L339 with `keyval` could be used here to have this even shorter.

yeoldegrove (changes requested):

tests/integration/security/compliance/test_disastig_0220.py
vivus-ignis: No, it won't be a failed test, because if key is not there, the default value is used. And the default value is good for this test. This test actually only looks for violations.

tests/plugins/systemd.py
vivus-ignis: Fixed.

tests/plugins/systemd.py
vivus-ignis: Done.

tests/plugins/systemd.py
vivus-ignis: Done

tests/plugins/systemd.py
vivus-ignis: Done

tests/integration/security/compliance/test_disastig_0220.py
yeoldegrove: Let me rephrase it.. My point was, that this assert/test never runs (is not reachable) if "Storage" is no Key inside `result`. Shouldn't this throw a failed test already if the key does not exist?

vivus-ignis: No it shouldn't.

tests/integration/security/compliance/test_disastig_0220.py
```diff
@@ -0,0 +1,59 @@
...
+    The sshd.service unit must forward stdout/stderr to the journal.
+    """
+    result = systemd.get_unit_properties("ssh")
+    assert (
+        result["StandardOutput"] == "journal"
+    ), f"sshd stdout not directed to journal: {result['StandardOutput']}"
+    assert result["StandardError"] in [
+        "journal",
+        "inherit",
+    ], f"sshd stderr not directed to journal or inherit: {result['StandardError']}"
```
yeoldegrove, line 42: Thanks for reworking the function `systemd.get_unit_properties`. I wonder why this isn't catched by the CI (failed test) ...

vivus-ignis: What do you mean?

yeoldegrove: Is the new function `systemd.get_unit_properties` really returning `StandardOutput` as a key? If yes, all good....just looked odd to me.

yeoldegrove: It seems the dict returns this key, nevermind.

PR #4474, opened 2026-03-18
ssh timeout test

vivus-ignis:

Required by https://stigaview.com/products/srg-gpos/v3r2/SRG-OS-000163-GPOS-00072/

Fixes https://github.com/gardenlinux/security/issues/247

### DIFF
--- a/.github/workflows/build_flavor.yml
+++ b/.github/workflows/build_flavor.yml
@@ -48,7 +48,7 @@ jobs:
         with:
           submodules: true
       - name: Install python-gardenlinux-lib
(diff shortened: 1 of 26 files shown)


features/stig/file.include/etc/ssh/sshd_config
Akendo, line 1: Why place a configuration file here? How big is the difference?

vivus-ignis: What are the alternatives to update sshd config?

PR #4484, opened 2026-03-19
Add test_disastig_00216.py and check for module 802.1q

saubhikdattagithub:
As per DISA STIG compliance requirement, the audit system must be configured to audit the loading and unloading of dynamic kernel modules.
Ref: SRG-OS-000471-GPOS-00216

Fixes [171](https://github.com/gardenlinux/security/issues/171)

### DIFF
--- a/tests/integration/security/compliance/test_disastig_00216.py
+++ b/tests/integration/security/compliance/test_disastig_00216.py
@@ -34,4 +34,8 @@ def test_kernel_module_load_logged(kernel_module, shell):
 
     logs = (dmesg_result.stdout + journal_result.stdout).lower()
 
-    assert "802.1q" in logs, "stigcompliance: kernel module load event not logged"
+    expected = MODULE.replace("8021q", "802.1q")
+
+    assert (
+        expected in logs
+    ), f"stigcompliance: kernel module load event for {MODULE} not logged"


tests/integration/security/compliance/test_disastig_00216.py
```diff
@@ -0,0 +1,47 @@
...
+    Ref: SRG-OS-000471-GPOS-00216
+    """
+
+    MODULE = "8021q"
+
+    assert kernel_module.is_module_available(
+        MODULE
+    ), f"stigcompliance: module {MODULE} not available on this system"
+
+    result = kernel_module._shell("dmesg -C", capture_output=True)
```
vivus-ignis: Don't do that. All the underscore methods and attributes should be considered an implementation detail and are not to be called externally. Consider passing shell fixture to this test case.

tests/integration/security/compliance/test_disastig_00216.py
```diff
@@ -0,0 +1,47 @@
...
+        "journalctl -k --since '1 min ago' --no-pager",
+        capture_output=True,
+    )
+    assert journal_result.returncode == 0, "stigcompliance: failed to read journal"
+
+    logs = (dmesg_result.stdout + journal_result.stdout).lower()
+
+    assert logs.strip() != "", "stigcompliance: no kernel logs captured"
+
+    assert "8021q" in logs, "stigcompliance: module load not logged in kernel logs"
```
vivus-ignis: Please try to come up with a substring that is not just a module name, but actually contains information on the kind of event that is logged: loading or unloading of a kernel module as required by the STIG rule.

vivus-ignis: And then, why not use MODULE variable here?

tests/integration/security/compliance/test_disastig_00216.py
```diff
@@ -0,0 +1,47 @@
...
+import pytest
+
+
+@pytest.mark.feature(
+    "not container and not gardener and not lima and not capi and not baremetal"
+)
+@pytest.mark.booted(reason="requires kernel logging")
+@pytest.mark.root(reason="required to manage kernel modules")
+@pytest.mark.modify(reason="modifies kernel module state")
+def test_module_load(kernel_module):
```
vivus-ignis: This module name does not reflect what it's testing. You are not testing module loading as the comment below suggests.

tests/integration/security/compliance/test_disastig_00216.py
```diff
@@ -0,0 +1,47 @@
...
+    that the audit system is configured to audit the loading and
+    unloading of dynamic kernel modules
+    Ref: SRG-OS-000471-GPOS-00216
+    """
+
+    MODULE = "8021q"
+
+    assert kernel_module.is_module_available(
+        MODULE
+    ), f"stigcompliance: module {MODULE} not available on this system"
```
vivus-ignis: This should not fail the test, because the test is not testing if you can load 8021q kernel module.

tests/integration/security/compliance/test_disastig_00216.py
```diff
@@ -0,0 +1,47 @@
...
+
+    result = kernel_module._shell("dmesg -C", capture_output=True)
+    assert result.returncode == 0, "stigcompliance: failed to clear dmesg"
```
vivus-ignis: This should not fail the test as it is not a prerequisite for a successful test as described in the docstring.

tests/integration/security/compliance/test_disastig_00216.py
```diff
@@ -0,0 +1,47 @@
...
+
+    loaded = kernel_module.load_module(MODULE)
+    assert loaded, f"stigcompliance: failed to load module {MODULE}"
+
+    time.sleep(1)
+
+    dmesg_result = kernel_module._shell("dmesg", capture_output=True)
+    assert dmesg_result.returncode == 0, "stigcompliance: failed to read dmesg"
```
vivus-ignis: What's the point in calling dmesg here?

tests/integration/security/compliance/test_disastig_00216.py
```diff
@@ -0,0 +1,47 @@
...
+
+    journal_result = kernel_module._shell(
+        "journalctl -k --since '1 min ago' --no-pager",
+        capture_output=True,
+    )
+    assert journal_result.returncode == 0, "stigcompliance: failed to read journal"
+
+    logs = (dmesg_result.stdout + journal_result.stdout).lower()
+
+    assert logs.strip() != "", "stigcompliance: no kernel logs captured"
```
vivus-ignis: You are not testing if kernel logging works in principle. 1 unit test == 1 test goal.

tests/integration/security/compliance/test_disastig_00216.py
saubhikdattagithub: Yes, thanks for pointing it out. Tried improving it, pls check

tests/integration/security/compliance/test_disastig_00216.py
```diff
@@ -0,0 +1,41 @@
...
+
+    dmesg_result = shell("dmesg", capture_output=True)
+    journal_result = shell(
+        "journalctl -k --since '1 min ago' --no-pager",
+        capture_output=True,
+    )
+
+    logs = (dmesg_result.stdout + journal_result.stdout).lower()
+
+    expected = MODULE.replace("8021q", "802.1q")
```
Akendo, line 37: Don't we unload that module if it wasn't loaded already?

tests/integration/security/compliance/test_disastig_00216.py
```diff
@@ -0,0 +1,41 @@
...
+    unloading of dynamic kernel modules
+    Ref: SRG-OS-000471-GPOS-00216
+    """
+
+    MODULE = "8021q"
+
+    if not kernel_module.is_module_available(MODULE):
+        return
+
+    kernel_module.load_module(MODULE)
```
Akendo, line 25: Keep in mind we might have a system that already has this module loaded.

saubhikdattagithub: Doesnt, the test kit, build all the flavors and test on them only specifically 🤔

tests/integration/security/compliance/test_disastig_00216.py
saubhikdattagithub: It was discussed to get the artifacts verified from `dmesg` and `journalctl`

PR #4485, opened 2026-03-19
Add test_disastig_00218.py for concurrent login audit check

saubhikdattagithub:
As per DISA STIG compliance requirement, The operating system must generate audit
records when concurrent logons to the same account occur from different sources.
Ref: SRG-OS-000472-GPOS-00218

Fixes [173](https://github.com/gardenlinux/security/issues/173)

image

### DIFF
--- a/.github/workflows/manual_gh_release_page.yml
+++ b/.github/workflows/manual_gh_release_page.yml
@@ -26,13 +26,6 @@ jobs:
     uses: ./.github/workflows/download_workflow_data.yml
     with:
       run_id: ${{ inputs.run_id }}
-  flavors_matrix:
-    needs: workflow_data
-    name: Generate flavors matrix (marked for publishing)
-    uses: ./.github/workflows/build_flavors_matrix.yml
-    with:
-      flags: "--no-arch --json-by-arch --publish"
-      ref: ${{ needs.workflow_data.outputs.commit_id }}
   glrd:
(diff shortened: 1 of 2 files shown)


tests/integration/security/compliance/test_disastig_00218.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    As per DISA STIG compliance requirement, the operating system must generate audit
+    records when concurrent logons to the same account occur from different sources.
+    Ref: SRG-OS-000472-GPOS-00218
+    """
+
+    shell(f"date '+%H:%M:%S' > {TIME_FILE}")
+
+    shell(f"su - {TEST_USER} -c 'sleep 30' &")
+    shell(f"su - {TEST_USER} -c 'sleep 30' &")
```
vivus-ignis: I think we need to discuss what "different sources" means here.

vivus-ignis (changes requested):

tests/integration/security/compliance/test_disastig_00218.py
saubhikdattagithub: Sure, lets clarify

saubhikdattagithub: Improved to a tty login + another ssh@localhost to emulate 2 different sources, as per the discussion yesterday

PR #4489, opened 2026-03-19
Add test_disastig_00096.py audit of identity

saubhikdattagithub:
As per DISA STIG compliance requirement, it is needed to verify the operating system produces audit records containing information to establish the identity of any individual or process associated with the event.
Ref: SRG-OS-000255-GPOS-00096 

Fixes [224](https://github.com/gardenlinux/security/issues/224)


saubhikdattagithub:
We now have a common `test_disastig_auditd` test file, where this check would be moved

PR #4533, opened 2026-03-26
mark.modify the check which was missed

saubhikdattagithub:
missed mark.modify for a check for STIG-000218

Fixes  https://github.com/gardenlinux/security/issues/173


saubhikdattagithub:
Covered by other PR (multiple tests)

PR #4540, opened 2026-03-27
motd compliance check

vivus-ignis:

DISA STIG check https://stigaview.com/products/srg-gpos/v3r2/SRG-OS-000228-GPOS-00088/

Fixes https://github.com/gardenlinux/security/issues/252


Akendo (changes requested):
Again, things like this I wouldn't like to add. We need to clarify this first. Please close this PR and document it in the ticket.

PR #4544, opened 2026-03-30
contains/icontains functions for the file plugin

vivus-ignis:
**What this PR does / why we need it**: test framework improvement that was created for a now closed PR.


yeoldegrove (changes requested):
The file plugin is intentionally not doing any parsing. We have the `parse_file` plugin for that.

yeoldegrove (changes requested):
My point is, I am not against having that "parse a raw file" function, I just think it needs to be integrated with the existing parsing plugins and not added to a plugin that is not supposed to do any parsing.

e.g. `SUPPORTED_FORMATS` https://github.com/gardenlinux/gardenlinux/blob/819c6f6a5d64f545802dac8113118280f59b92bd/tests/plugins/parse.py#L13 could be extended with a raw key. and _parse_conent could than call your proposed containsfunction.
[parse.py](https://github.com/gardenlinux/gardenlinux/blob/819c6f6a5d64f545802dac8113118280f59b92bd/tests/plugins/parse.py)

A requirement to add this would of course have to have a use-case...

PR #4546, opened 2026-03-30
Add test_disastig_00146.py for timedatectl check of UTC/GMT

saubhikdattagithub:
As per DISA STIG requirement, the operating system must record audit
timestamps that can be mapped to UTC or GMT.
Ref: SRG-OS-000359-GPOS-00146

Fixes [301](https://github.com/gardenlinux/security/issues/301)

### DIFF
--- a/tests/integration/security/compliance/test_disastig_00149.py
+++ b/tests/integration/security/compliance/test_disastig_00149.py
@@ -1,0 +1,75 @@
+import pytest
+from plugins.file import File
+
+PKG_BINARIES = [
+    "/usr/bin/apt",
+    "/usr/bin/apt-get",
+    "/usr/bin/dpkg",
+]
+
+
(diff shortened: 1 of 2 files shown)


tests/integration/security/compliance/test_disastig_00146.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    As per DISA STIG requirement, the operating system must record audit
+    timestamps that can be mapped to UTC or GMT.
+    Ref: SRG-OS-000359-GPOS-00146
+    """
+    result = shell(
+        "timedatectl show -p Timezone",
+        capture_output=True,
+        ignore_exit_code=True,
+    )
```
yeoldegrove, line 18: We already have a plugin that could be extended.
https://github.com/gardenlinux/gardenlinux/blob/main/tests/plugins/timedatectl.py

PR #4555, opened 2026-03-31
Add test_disastig_00152.py for audit enforcement test with a handler

saubhikdattagithub:
As per DISA STIG compliance requirement, the operating system must audit the
enforcement actions used to restrict access associated with changes to the system.
Ref: SRG-OS-000365-GPOS-00152

Fixes [306](https://github.com/gardenlinux/security/issues/306)


tests/handlers/enforcement.py
```diff
@@ -0,0 +1,26 @@
...
+    shell(f"rm -f {TEST_FILE} {TIME_FILE}", ignore_exit_code=True)
+    shell(f"touch {TEST_FILE}")
+    shell(f"chmod 600 {TEST_FILE}")
+
+    yield TEST_FILE, TIME_FILE
+
+    shell(f"rm -f {TEST_FILE} {TIME_FILE}", ignore_exit_code=True)
+
+
+@pytest.fixture
```
yeoldegrove: If we need this handler, could we call it something more descriptive?

tests/integration/security/compliance/test_disastig_00152.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+
+    FILE_TEST, FILE_TIME = enforcement_test
+
+    shell(f"date '+%H:%M:%S' > {FILE_TIME}")
+
+    shell(
+        f"su nobody -s /bin/sh -c 'cat {FILE_TEST}'",
+        ignore_exit_code=True,
+    )
```
yeoldegrove, line 18: Can't we e.g. try to access `/etc/shadow` or a similar already existing restricted file and have a test without any preparation/handler?

yeoldegrove (changes requested):

tests/integration/security/compliance/test_disastig_00152.py
saubhikdattagithub: Yes, thanks for the simplification

yeoldegrove (changes requested):
https://stigaview.com/products/srg-gpos/v3r1/SRG-OS-000365-GPOS-00152/ and e.g. https://www.tenable.com/audits/items/DISA_STIG_Oracle_Linux_9_v1r5.audit:f48ee72965773db3b715cab09683e8c9 reads more like this is meant in a more general way.

Akendo (approved):

PR #4570, opened 2026-04-01
Add test_disastig_00173.py to validate weak_macs in SSHD config

saubhikdattagithub:
As per DISA STIG compliance requirements, the operating system must implement
cryptographic mechanisms to protect the integrity of nonlocal maintenance and
diagnostic communications, when used for nonlocal maintenance sessions.
This test verifies that SSH is configured with strong MAC algorithms to ensure
integrity protection for remote sessions.
Ref: SRG-OS-000393-GPOS-00173

Fixes [173](https://github.com/gardenlinux/security/issues/321)

### DIFF
--- a/tests/integration/security/compliance/test_disastig_0067.py
+++ b/tests/integration/security/compliance/test_disastig_0067.py
@@ -1,0 +1,60 @@
+import pwd
+
+import pytest
+
+"""
+Ref: SRG-OS-000132-GPOS-00067
+
+Verify the operating system separates user functionality (including user
+interface services) from operating system management functionality.
+
+Ref: SRG-OS-000134-GPOS-00068
+
+Verify the operating system isolates security functions from nonsecurity
+functions.
+
+Ref: SRG-OS-000326-GPOS-00126
+
+Verify that the operating system prevents all software from executing at higher
+privilege levels than users executing the software.
+"""
+
+SETUID_BINARIES_ALLOWLIST = {
+    "default": {
+        "/usr/bin/chfn",
+        "/usr/bin/chsh",
+        "/usr/bin/gpasswd",
+        "/usr/bin/passwd",
+        "/usr/bin/su",
+        "/usr/bin/sudo",
+        "/usr/bin/sudoedit",
+    },
+    "lima": {"/usr/bin/fusermount3", "/usr/bin/fusermount"},
+}
+
+
+def test_only_root_user_has_uid_zero():
+    adm_users = [u.pw_name for u in pwd.getpwall() if u.pw_uid == 0]
+    assert adm_users == [
+        "root"
+    ], f"only root user should have uid 0, instead {adm_users} found"
+
+
+@pytest.mark.feature("not lima")
+def test_only_setuid_binaries_from_the_list_are_allowed(exposed_setuid_binaries):
+    if exposed_setuid_binaries:
+        diff = exposed_setuid_binaries - SETUID_BINARIES_ALLOWLIST["default"]
+        assert (
+            not diff
+        ), f"unexpected setuid binaries with too broad exec permissions: {exposed_setuid_binaries=} {diff=}"
+
+
+@pytest.mark.feature("lima")
+def test_only_lima_setuid_binaries_from_the_list_are_allowed(exposed_setuid_binaries):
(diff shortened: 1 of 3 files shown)


tests/integration/security/compliance/test_disastig_00173.py
```diff
@@ -0,0 +1,92 @@
+import pytest
+from plugins.dpkg import Dpkg
+
+
+@pytest.mark.feature(
+    "not container and not lima and not gardener and not capi and not baremetal"
+)
```
yeoldegrove: Just saw that a lot of markers like these are already merged. Isn't the actual setting tied to a single feature and would it make sense to have a positive list instead of this long negative list?

tests/integration/security/compliance/test_disastig_00173.py
```diff
@@ -0,0 +1,92 @@
...
+    diagnostic communications, when used for nonlocal maintenance sessions.
+    This test verifies that SSH is configured with strong MAC algorithms to ensure
+    integrity protection for remote sessions.
+    Ref: SRG-OS-000393-GPOS-00173
+    """
+
+    if not dpkg.package_is_installed("openssh-server"):
+        pytest.skip(
+            "openssh-server not installed; no nonlocal maintenance mechanism present"
+        )
```
yeoldegrove: Can't we tie this to a feature marker like "ssh"?

yeoldegrove: If not, the `sshd` plugin should handle this.

tests/integration/security/compliance/test_disastig_00173.py
```diff
@@ -0,0 +1,92 @@
...
+
+    config = sshd.get_config()
+    macs = config.get("macs")
+
+    assert macs is not None, "stigcompliance: SSH MACs configuration not defined"
```
yeoldegrove: Do we need this assert or is it covererd by the next assert? If we need it, split the tests, please.

tests/integration/security/compliance/test_disastig_00173.py
```diff
@@ -0,0 +1,92 @@
...
+        mac_list.extend([m.strip() for m in entry.split(",")])
+
+    weak_macs = {
+        "hmac-md5",
+        "hmac-md5-96",
+    }
+
+    assert not any(
+        mac in weak_macs for mac in mac_list
+    ), "stigcompliance: weak MAC algorithms present in SSH configuration"
```
yeoldegrove: Same questions than above.

tests/integration/security/compliance/test_disastig_00173.py
saubhikdattagithub: Basically, idea is to check if we can have the tests to comply with our existing flavors already. Since the feature `stig` has no flavor yet and cannot be tested currently and ask is to expedite the tests for STIG, its a break-fix mechanism to see what works best for the tests. 
Agree on the observation here, if long negatives, better to have the `stig` feature to have it on board. Shall incorporate.

tests/integration/security/compliance/test_disastig_00173.py
saubhikdattagithub: Yes both strong mac and weak mac checks are needed, however removed the redundant `presence` check

tests/integration/security/compliance/test_disastig_00173.py
```diff
@@ -0,0 +1,72 @@
...
+    strong_macs = {
+        "hmac-sha2-256",
+        "hmac-sha2-512",
+        "hmac-sha2-256-etm@openssh.com",
+        "hmac-sha2-512-etm@openssh.com",
+    }
+
+    assert any(
+        mac in strong_macs for mac in mac_list
+    ), "stigcompliance: no strong MAC algorithms configured for SSH integrity protection"
```
yeoldegrove: I read this like: "Any MAC needs to be strong". Do we really want this or do we want that every algorithms is strong (reads more like it when I read the test description above).

saubhikdattagithub: Although STIG requires one of the strong MACs present, however we could test stricter compliance

PR #4575, opened 2026-04-01
Add test_disastig_00187.py to check integrity of transmitted info

saubhikdattagithub:
As per DISA STIG compliance requirement, the operating system must protect
the confidentiality and integrity of transmitted information.
This test verifies that insecure network services (which transmit data
without encryption) are not enabled on the system.
Ref: SRG-OS-000423-GPOS-00187

Fixes [330](https://github.com/gardenlinux/security/issues/330)

### DIFF
--- a/features/stig/file.include/etc/ssh/sshd_config
+++ b/features/stig/file.include/etc/ssh/sshd_config
@@ -56,3 +56,8 @@ AuthenticationMethods publickey
 
 HostKey /etc/ssh/ssh_host_ed25519_key
 HostKey /etc/ssh/ssh_host_rsa_key
+
+# Requisite Ciphers, MAC, KexAlgorithms
+Ciphers aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes128-ctr
+MACs hmac-sha2-512,hmac-sha2-256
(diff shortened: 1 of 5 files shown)


tests/integration/security/compliance/test_disastig_00187.py
```diff
@@ -0,0 +1,32 @@
...
+INSECURE_SERVICES = [
+    "telnet.service",
+    "rsh.service",
+    "rexec.service",
+    "rlogin.service",
+    "vsftpd.service",
+]
+
+
+@pytest.mark.feature("not container and not lima")
```
yeoldegrove: Is lima starting one of these services? If not, can we remove it?

yeoldegrove: Will container ever be booted? If not, can we remove it?

tests/integration/security/compliance/test_disastig_00187.py
```diff
@@ -0,0 +1,32 @@
...
+    Ref: SRG-OS-000423-GPOS-00187
+    """
+
+    for service in INSECURE_SERVICES:
+        assert not systemd.is_enabled(
+            service
+        ), f"stigcompliance: insecure service {service} is enabled"
+
+        assert not systemd.is_active(
+            service
```
yeoldegrove: We want one assert per test (with exceptions) so this needs to be splitted.
Also, I would vote for either using parametrize here or just writing down stupidly tests for each service. This way we make sure to only have a single assert per test (an not having loops which might return to early).

yeoldegrove (approved):

PR #4590, opened 2026-04-07
Add test_disastig_00057.py for validating SSHD presence

saubhikdattagithub:
As per DISA STIG compliance requirement, the operating system must implement
replay-resistant authentication mechanisms for network access to privileged accounts.
This test verifies that SSH (a replay-resistant authentication mechanism)
is available for network access.
Ref: SRG-OS-000112-GPOS-00057

Fixes [234](https://github.com/gardenlinux/security/issues/234)

### DIFF
--- a/tests/integration/security/compliance/test_disastig_00057.py
+++ b/tests/integration/security/compliance/test_disastig_00057.py
@@ -1,11 +1,10 @@
 import pytest
-from plugins.sshd import Sshd
 
 
 @pytest.mark.feature("not container")
 @pytest.mark.booted(reason="requires sshd effective configuration")
 @pytest.mark.root(reason="required to inspect ssh configuration")
-def test_ssh_available_for_network_access(sshd: Sshd):
+def test_ssh_available_for_network_access(sshd):
     """
     As per DISA STIG compliance requirement, the operating system must implement
     replay-resistant authentication mechanisms for network access to privileged accounts.


tests/integration/security/compliance/test_disastig_00057.py
```diff
@@ -0,0 +1,18 @@
...
+    As per DISA STIG compliance requirement, the operating system must implement
+    replay-resistant authentication mechanisms for network access to privileged accounts.
+    This test verifies that SSH (a replay-resistant authentication mechanism)
+    is available for network access.
+    Ref: SRG-OS-000112-GPOS-00057
+    """
+
+    config = sshd.get_config()
+
+    assert config, "stigcompliance: sshd configuration not available"
```
yeoldegrove, line 18: I do not really get what this test is supposed to do? "ssh available for network access" does not make much sense to me.

yeoldegrove (changes requested):

tests/integration/security/compliance/test_disastig_00057.py
saubhikdattagithub: Yes that is correct, basically we were parked for this test. However we had a discussion and as here https://github.com/gardenlinux/security/issues/234#issuecomment-4162464526 it was decided that we only check for the presence of SSHD as we already test other basic and core working in our test framework, well.

yeoldegrove: https://github.com/gardenlinux/security/issues/234#issuecomment-4162464526 talks about "SSH enabled" which is not the check we have here. Also the test description is way too nondescript.

The test is not designed well... just reading it I would assume that `config==true` (a boolean) which is not true as it returns a way more complex structure.

PR #4591, opened 2026-04-08
Add test_disastig_00154.py to prevent program execution local policies

saubhikdattagithub:
As per DISA STIG requirement, the operating system must prevent program
execution in accordance with local policies regarding software program usage.
Ref: SRG-OS-000368-GPOS-00154

Fixes [308](https://github.com/gardenlinux/security/issues/308)

### DIFF
--- a/tests/integration/security/compliance/test_disastig_00154.py
+++ b/tests/integration/security/compliance/test_disastig_00154.py
@@ -1,0 +1,276 @@
+import os
+
+import pytest
+from plugins.dpkg import Dpkg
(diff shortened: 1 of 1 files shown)


saubhikdattagithub:
Closing in favour of #4729 which uses the correct feature marker (_selinux) and targets the right STIG ref (SRG-OS-000312-GPOS-00154). The _selinux flag is an opt-in feature in Garden Linux and should not be assumed present in disaSTIGmedium builds.

PR #4617, opened 2026-04-13
Fix sources list file adaption in latest image

SujanaSubr:
Nightly build fails since sources.list doesn't exist anymore
Fixes #
https://github.com/gardenlinux/gardenlinux/actions/runs/24329367219/job/71033160247
**Definition of Done:**

**Release note**:

```feature user

```

### DIFF
--- a/images/kmodbuild/Containerfile
+++ b/images/kmodbuild/Containerfile
@@ -7,8 +7,23 @@ FROM $base
 COPY --from=snapshot /etc/apt/sources.list /tmp/sources.list
 RUN grep '^deb ' /tmp/sources.list >> /etc/apt/sources.list && rm /tmp/sources.list
 
-RUN version="$(awk '/^URIs:.*https:\/\/packages\.gardenlinux\.io\/gardenlinux/ { found=1 } found && /^Suites:/ { print $2; exit }' /etc/apt/sources.list.d/gardenlinux.sources)" && \
+RUN version="$(awk ' \
+    /^$/ { \
+        if (found && suite) { print suite; exit } \
+        found=0; suite="" \
+    } \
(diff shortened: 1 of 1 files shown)


images/kmodbuild/Containerfile
```diff
@@ -6,7 +6,10 @@ FROM $snapshot AS snapshot
 FROM $base
 COPY --from=snapshot /etc/apt/sources.list /tmp/sources.list
 RUN grep '^deb ' /tmp/sources.list >> /etc/apt/sources.list && rm /tmp/sources.list
-RUN version="$(awk '$2 == "https://packages.gardenlinux.io/gardenlinux" { print $3 }' /etc/apt/sources.list)" && printf 'Package: *\nPin: release n=%s\nPin-Priority: 900\n' "$version" > /etc/apt/preferences.d/gardenlinux
+
+RUN version="$(awk '/^URIs:.*https:\/\/packages\.gardenlinux\.io\/gardenlinux/ { found=1 } found && /^Suites:/ { print $2; exit }' /etc/apt/sources.list.d/gardenlinux.sources)" && \
```
nkraetzschmar: This is probably fine, but if the order in the file happens to be `Suites` first, then `URIs` this will break. Maybe you could store both a found state and a suite variable and then on end of entry (blank line or end of file) print the suite var if found is true.

images/kmodbuild/Containerfile
```diff
@@ -6,7 +6,10 @@ FROM $snapshot AS snapshot
...
+    printf 'Package: *\nPin: release n=%s\nPin-Priority: 900\n' "$version" > /etc/apt/preferences.d/gardenlinux
```
nkraetzschmar, line 25: Also for added safety it might be good to check that `$version` is not empty before using it

nkraetzschmar (approved):
Small suggestions, but am also fine merging this as is :+1:

PR #4619, opened 2026-04-15
Fix/fips gnu tls generate hmac

Akendo:
In this PR, we compute the HMAC of the dependency of GnuTLS manually. This is necessary since Debian has published a change to one of GnuTLS's dependencies that wasn't open-sourced yet. 

The main problem is that the build environment in which GnuTLS is built remains fairly static and computes the HMAC for the dependency at build-time. However, when one of the underlying dependencies is changed the HMAC test will fail. In the past, this could have been addressed by rebuilding the GnuTLS package. Since the packages from the build environment would be in the nightly. 

With this change, we will compute it every time we build. This ensures that the HMAC is computed for the libraries that are present to make it independent of the package built environment.


yeoldegrove (approved):
LGTM

PR #4647, opened 2026-04-16
Add a test that check that BTRFS is disabled.

Akendo:
BTRFS makes use of the xxhash64. Since it is considered not an approved algorithm by the NIST. Hence, we have to disable it. We also have to remove BTRFS, since it uses this.

Relates to: https://github.com/gardenlinux/security/issues/405


vivus-ignis (approved):
LGTM

PR #4648, opened 2026-04-17
Tests/check that md5 is disabled

Akendo:
We have to ensure that MD5 isn't loaded. This test uses Python's `hashlib` to ensure that this isn't the case. Since it uses OpenSSL to load MD5.

### DIFF
--- a/tests/integration/security/compliance/test_fips.py
+++ b/tests/integration/security/compliance/test_fips.py
@@ -3,6 +3,8 @@ import hmac
 import os
 from ctypes import CDLL, c_int, c_void_p
 from ctypes.util import find_library
+from hashlib import _hashlib # pyright: ignore
+from hashlib import md5 as MD5
 from hashlib import sha256 as SHA256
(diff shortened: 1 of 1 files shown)


tests/integration/security/compliance/test_fips.py
```diff
@@ -20,6 +20,22 @@
...
+    We try to load the MD5 constructor to see if OpenSSL disallows the usage of MD5.
+    Fails when we have a vaild MD5 object in a Security senstive context.
+    """
+    try:
+        MD5()
+    except hashlib._hashlib.UnsupportedDigestmodError:
+        assert True
+        return
+
+    assert False, "MD5 can be loaded!"
```
vivus-ignis:
```suggestion
    pytest.raises(hashlib._hashlib.UnsupportedDigestmodError, MD5)
```

vivus-ignis (changes requested):

vivus-ignis (approved):
LGTM

PR #4649, opened 2026-04-17
Check that the kernel has the right FIPS kernel module name

Akendo:
In this PR, we add a test to the kernel that the right name of the FIPS module has been set. 

Relates to: https://github.com/gardenlinux/security/issues/368


tests/integration/security/compliance/test_fips.py
```diff
@@ -491,3 +491,20 @@ def test_kernel_hmac_file_is_correct(
...
+    """
+    Ensure that the FIPS module name is set correctly for the kernel.
+    It should look like as follows:
+       - SAP SE Garden Linux [RELEASE] Kernel Cryptographic Module
+    """
+    for config in kernel_configs.get_installed():
+        parsed_config = parse_file.parse(config.path, format="keyval")
+        assert (
+            parsed_config["CONFIG_CRYPTO_FIPS_NAME"]
+            == "SAP SE Garden Linux nightly Kernel Cryptographic Module"
```
commel, line 509: Is this supposed to be fixed on "nightly"? With a modified `config.version` it should be possible to test again releases as well.

Image

Akendo: Yes that's intended that way. Needs to be adapteted for 2150 and 1877.

5kt (approved):

PR #4650, opened 2026-04-17
Add check for FIPS-module name for libgcrypt

Akendo:
In this PR, we add a test that should ensure that libgcrypt has the correct FIPS-Module name set. 

Relates to: https://github.com/gardenlinux/security/issues/368


tests/integration/security/compliance/test_fips.py
```diff
@@ -321,6 +321,37 @@ def test_libgcrypt_is_in_fips_mode():
...
+    shared_lib_name = find_library("gcrypt")
+    libgcrypt = CDLL(shared_lib_name)
+
+    libgcrypt.gcry_get_config.restype = c_char_p
+    libgcrypt.gcry_get_config.argtypes = [c_int, c_char_p]
+
+    config = libgcrypt.gcry_get_config(0, None)
+
+    if not config:
+        assert False, "Can't load configuration from libgcrypt via gcry_get_config"
```
vivus-ignis:
```suggestion
    assert not config, "Can't load configuration from libgcrypt via gcry_get_config"
```

tests/integration/security/compliance/test_fips.py
```diff
@@ -321,6 +321,37 @@ def test_libgcrypt_is_in_fips_mode():
...
+
+    for entry in config.decode().splitlines():
+        if "fips-mode" in entry:
+            assert "fips-mode:y::SAP SE Garden Linux nightly Libgcrypt Cryptographic Module:"
```
vivus-ignis:
```suggestion
    found = [entry for entry in config.decode().splitlines() if "fips-mode" in entry]
    assert found, "fips-mode:y::SAP SE Garden Linux nightly Libgcrypt Cryptographic Module:"
```

vivus-ignis (changes requested):

tests/integration/security/compliance/test_fips.py
Akendo: This will not work.

Akendo: Technically, I should throw an exception here.

vivus-ignis: Then it probably should be 

`assert config, "Can't load configuration from libgcrypt via gcry_get_config"`

Akendo: I think the code is redundant. Thanks for point out!

vivus-ignis (approved):
LGTM

PR #4651, opened 2026-04-17
Update ToDo, re-enable test

Akendo:
Re-enable the ARM64 test to compute the kernel HMAC. This test works fine, unless build crossplatfrom.


tests/integration/security/compliance/test_fips.py
```diff
@@ -429,10 +429,11 @@ def test_kernel_configs_btrfs_is_disabled(
         ), f"CONFIG_BTRFS is set in {config.path}"
 
 
-# TODO: check why this does not work on arm64
+# TODO: check why this does not work on arm64, when build crossplatfrom.
+# When run podman on x86_64 and you try to build this via podman for arm64 it will fail.
 @pytest.mark.testcov(["GL-TESTCOV-_fips-config-kernel-hmac"])
 @pytest.mark.feature("_fips")
-@pytest.mark.arch("not arm64")
+@pytest.mark.arch("not _usi")
```
vivus-ignis: The only supported values for the arch annotation are: amd64, arm64 (see https://github.com/gardenlinux/gardenlinux/blob/b0249f34b5034325a4623cd67f0e9fdc04a6b264/tests/plugins/arch.py#L17). `_usi` though should work with feature annotation.

vivus-ignis (changes requested):

tests/integration/security/compliance/test_fips.py
Akendo: Thanks for catching this.

Akendo: Fixed.

vivus-ignis (approved):
LGTM

PR #4653, opened 2026-04-20
Tests/validate fips openssl module name

Akendo:
In this PR, we add a test to ensure that OpenSSL has the FIPS vendor set. 

Related to: https://github.com/gardenlinux/security/issues/368


vivus-ignis (approved):
Ship it!

PR #4654, opened 2026-04-20
harmonized disa stig test file names

vivus-ignis:
**What this PR does / why we need it**: we have slightly different naming approaches for disa stig test files; this is an attempt to harmonize them


saubhikdattagithub (approved):
LGTM

PR #4668, opened 2026-04-21
checking for password-changing tools

vivus-ignis:

Fixes https://github.com/gardenlinux/security/issues/360


yeoldegrove (changes requested):
The test description does not match the test.

tests/integration/security/compliance/test_disastig_00170.py
```diff
@@ -0,0 +1,39 @@
...
+    assert len(results) == 1, "pam_unix module should be used for account checking"
+
+
+def test_password_recovery_tools_available(shell):
+    """
+    passwd sets a password
+    chage controls expiration date
+    """
+    assert shell("which passwd")
+    assert shell("which chage")
```
yeoldegrove: I do not think that this test does make much sense since these are very basic tools that will always exist. Checking if they are in the path does not add much quality... 

If you think we need this, could you use the `shutils.which()` call like in the other tests?

yeoldegrove (changes requested):

tests/integration/security/compliance/test_disastig_00170.py
vivus-ignis: Agree. Removed the test.

yeoldegrove (approved):

PR #4678, opened 2026-04-21
Add features/disaSTIGlow — low severity STIG compliance feature

saubhikdattagithub:
Add `disaSTIGlow` feature — low severity STIG compliance configurations including auditd config, faillock, limits, pwquality, and sysctl settings.
Make features/stig working + make a flavor of it for test framework


features/disaSTIGlow/exec.config
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+set -x
+
+ensure_line() {
```
Akendo: Do we need this? This feels somehow duplicated? Such tests should be located within the tests.

features/disaSTIGlow/file.include/etc/sysctl.d/99-disaSTIG.conf
```diff
@@ -0,0 +1,2 @@
+# https://www.stigviewer.com/stig/canonical_ubuntu_20.04_lts/2023-09-08/finding/V-255913
+kernel.dmesg_restrict = 1
```
Akendo: This should be already presented in the [`server`](https://github.com/gardenlinux/gardenlinux/blob/4ae57fcf6251c7e95565e398f3ec7c91f76df56a/features/server/file.include/etc/sysctl.d/40-restrict-dmesg.conf) feature.

features/disaSTIGlow/file.include/etc/audit/auditd.conf
```diff
@@ -0,0 +1,21 @@
+# https://www.stigviewer.com/stig/canonical_ubuntu_20.04_lts/2023-09-08/finding/V-238307
+space_left_action = email
```
Akendo: I think we do not have any MTA so far.

saubhikdattagithub: Removed

features/disaSTIGlow/file.include/etc/sysctl.d/99-disaSTIG.conf
saubhikdattagithub: Great, thanks

features/disaSTIGlow/file.include/etc/security/pwquality.conf
```diff
@@ -0,0 +1,10 @@
+# https://www.stigviewer.com/stig/canonical_ubuntu_20.04_lts/2023-09-08/finding/V-238222
```
Akendo: The URL is not working. We either fix this or remove this URL and keep only the reference.

Akendo:
One detail I notice here is that almost all the changes have reference to the Ubuntu 20.04 STIG, but none to the STIG we're working on. How are those changes 'relevant' to the work we did?

saubhikdattagithub:
> One detail I notice here is that almost all the changes have reference to the Ubuntu 20.04 STIG, but none to the STIG > we're working on. How are those changes 'relevant' to the work we did?

Yes, I agree on this point. The reference URL should be updated to the ones we are working with instead. I am working to update them with relevance.

features/disaSTIGlow/file.include/etc/security/pwquality.conf
saubhikdattagithub: Yes, I agree on this point. The reference URL should be updated to the ones we are working with instead. I am working to update them with relevance.

features/disaSTIGlow/file.include/etc/sysctl.d/99-disaSTIG.conf
saubhikdattagithub: Commented the requirement and kept the file which doesnot look convincing, however, keeping a comment would assure we are covered for that STIG ID

Akendo (approved):
LGTM

PR #4686, opened 2026-04-21
test: update test_stig.py for disaSTIG feature restructure

saubhikdattagithub:
Update `test_stig.py` to reference the new `disaSTIGmedium`/`disaSTIGlow` feature names and restructured compliance checks.
Make features/stig working + make a flavor of it for test framework


saubhikdattagithub:
The agreed way is using "test_disastig_XXXXX.py" and this change is not needed anymore

PR #4727, opened 2026-04-27
Add test_disastig_00051.py to verify no duplicate UIDs in /etc/passwd

saubhikdattagithub:

Add `test_disastig_00051.py` to verify no duplicate UIDs in `/etc/passwd`.

As per DISA STIG compliance requirements (SRG-OS-000104-GPOS-00051), all accounts assigned to interactive users must be given unique User IDs (UIDs). This test checks that no two entries in `/etc/passwd` share the same UID.

### DIFF
--- a/tests/integration/security/compliance/test_disastig_00051.py
+++ b/tests/integration/security/compliance/test_disastig_00051.py
@@ -1,0 +1,28 @@
+import pytest
+from plugins.linux_etc_files import Passwd
+
+
+@pytest.mark.feature("disaSTIGmedium")
+@pytest.mark.root(reason="requires access to /etc/passwd")
+def test_no_duplicate_uids_in_passwd(passwd_entries: list[Passwd]):
+    """
+    As per DISA STIG compliance requirements, all accounts assigned to
+    interactive users must be given unique User IDs (UIDs).
+    This test verifies that no two entries in /etc/passwd share the same UID.
+    Ref: SRG-OS-000104-GPOS-00051
+    """
+    seen = {}
+    duplicates = []
+
+    for entry in passwd_entries:
+        if entry.uid in seen:
+            duplicates.append(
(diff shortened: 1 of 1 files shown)


tests/integration/security/compliance/test_disastig_00051.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+from plugins.linux_etc_files import Passwd
+
+
+@pytest.mark.feature("disaSTIGmedium")
+@pytest.mark.root(reason="requires access to /etc/passwd")
```
yeoldegrove, line 6:
```
root@bc8471ea02b4:/# ls -la /etc/passwd 
-rw-r--r-- 1 root root 837 Feb 18 00:00 /etc/passwd
```

tests/integration/security/compliance/test_disastig_00051.py
```diff
@@ -0,0 +1,27 @@
...
+        if entry.uid in seen:
+            duplicates.append(
+                f"{entry.name} (uid={entry.uid}) duplicates {seen[entry.uid]}"
+            )
+        else:
+            seen[entry.uid] = entry.name
+
+    assert (
+        not duplicates
+    ), "stigcompliance: duplicate UIDs found in /etc/passwd:\n" + "\n".join(duplicates)
```
yeoldegrove, line 27: `tests/integration/core/test_users_groups.py` has the same test using a different way of doing this. Please let's align if we need test duplication at all.

yeoldegrove (changes requested):

tests/integration/security/compliance/test_disastig_00051.py
yeoldegrove: Please do the test the same way as it already implemented...elsewise we might have problems matching those later.

saubhikdattagithub:
@yeoldegrove thanks for the cross review, will close it for now. 👍

PR #4732, opened 2026-04-27
Add test_disastig_00227.py to verify X11 forwarding is disabled in SSH

saubhikdattagithub:

Add `test_disastig_00227.py` to verify that `X11Forwarding` is set to `no` and `X11UseLocalhost` is set to `yes` in the SSH daemon configuration.

As per DISA STIG compliance requirements (SRG-OS-000480-GPOS-00227), the operating system must be configured so that the SSH daemon does not allow X11 Forwarding.


yeoldegrove (approved):
LGTM

PR #4735, opened 2026-04-27
os verifies correct status of all security functions

vivus-ignis:

Fixes https://github.com/gardenlinux/security/issues/338


tests/integration/security/compliance/test_disastig_00199.py
```diff
@@ -0,0 +1,31 @@
...
+
+@pytest.mark.feature("selinux")
+def test_selinux_enabled(shell):
+    result = shell("sestatus", capture_output=True)
+    status = [
+        line.split(":")[1].trim()
+        for line in result.stdout.splitlines()
+        if line.split(":")[0] == "SELinux status"
+    ]
+    assert status[0] == "enabled"
```
yeoldegrove: As we already have a plugin for LSMs...
Do you think we could use `tests/plugins/lsm.py` directly or extend it to e.g. output/parse:
```
root@localhost:~# getenforce 
Permissive
```

yeoldegrove (changes requested):

yeoldegrove (approved):

PR #4736, opened 2026-04-28
tests: add /sbin/nologin to the list of nologin shells

5kt:
_/sbin/nologin_ should also be part of the list of no login shells.


saubhikdattagithub (approved):
LGTM, as discussed today for nightly failure for `_dhcpcd`

moritzkirchnersap (approved):

PR #4737, opened 2026-04-28
error messages should not be revealed

vivus-ignis:

Fixes https://github.com/gardenlinux/security/issues/251


saubhikdattagithub (approved):
LGTM

PR #4738, opened 2026-04-28
fixed stig feature annotations

vivus-ignis:
**What this PR does / why we need it**: I've used a temporary "passthrough" stig feature name for some tests


saubhikdattagithub (approved):
LGTM

PR #4743, opened 2026-04-30
renamed stig tests to conform to the naming convention

vivus-ignis:
**What this PR does / why we need it**: fixes test file names


saubhikdattagithub (approved):
Thank you

PR #4746, opened 2026-04-30
a way to mark disa stig rules covered by tests somewhere else

vivus-ignis:
**What this PR does / why we need it**: it is possible that requirements for a certain DISA STIG rule are already covered by tests written for another rule; I propose a way to handle it so that it's easy to lookup tests by a rule number


saubhikdattagithub (approved):
LGTM naming convention is standard with the proposal

PR #4776, opened 2026-05-11
re-enable inactivity-disabled (not intentional ones) workflows in package repos daily

saubhikdattagithub:
## Summary

GitHub automatically disables scheduled workflows after 60 days of repository inactivity. This affects all \`gardenlinux/package-*\` repos which run a nightly \`build.yml\` — they gets disabled 

**Note**: This change does not touch **intentionally** disabled repos as below:

```bash
  package-libyang2
  package-docker-machine
  package-neofetch
  package-ostree
  package-syslinux
  package-tar
  package-fail2ban
  package-uid-wrapper
  package-usr-is-merged
  package-dropbear-static
```

### DIFF
--- a/.github/workflows/enable_disabled_workflows.yml
+++ b/.github/workflows/enable_disabled_workflows.yml
@@ -1,0 +1,64 @@
+name: Re-enable inactivity-disabled workflows in package repos
+
+on:
+  schedule:
+    - cron: '17 5 * * *'  # daily at 05:17 UTC
+  workflow_dispatch:        # allow manual trigger
+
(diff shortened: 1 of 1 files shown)


vivus-ignis:
Is this using a personal access token?

saubhikdattagithub:
> Is this using a personal access token?

Not really, Currently the token is a dummy. If the Idea and logic is fine, would work on using the actual token used in other workflows/create new

saubhikdattagithub:
Not needed anymore post NeoNephos integration

PR #4786, opened 2026-05-12
disa stig 00206

vivus-ignis:

Fixes https://github.com/gardenlinux/security/issues/344


saubhikdattagithub (approved):
LGTM

PR #4789, opened 2026-05-13
Update nightly.yml and add timezone

moritzkirchnersap:
The schedule works, but it rarely seems to finish before our daily. If we want to have time to investigate before the meeting, we need to start earlier and I believe it will be the least confusing if we add a timezone.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onschedule

### DIFF
--- a/.github/workflows/build_lima_yaml_container.yml
+++ b/.github/workflows/build_lima_yaml_container.yml
@@ -3,6 +3,8 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '0 0 * * 0'
+      timzone: 'UTC'
+
 jobs:
   lima_manifest_container:
     name: Build Container for lima manifest generation
--- a/.github/workflows/build_lima_yaml_container.yml
+++ b/.github/workflows/build_lima_yaml_container.yml
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '0 0 * * 0'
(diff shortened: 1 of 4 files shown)


yeoldegrove (changes requested):
I really like this but could you change all our cron workflows please?

```
❯ grep "cron:" -C2 .github
.github/workflows/test_update_python_runtime.yml
5-  schedule:
6-    # Run at 12:00 noon CEST (10:00 UTC)
7:    - cron: "0 10 * * *"
8-
9-jobs:

.github/workflows/nightly.yml
5-on:
6-  schedule:
7:    - cron: '0 4 * * *'
8-jobs:
9-  build:

.github/workflows/build_lima_yaml_container.yml
3-  workflow_dispatch:
4-  schedule:
5:    - cron: '0 0 * * 0'
6-jobs:
7-  lima_manifest_container:

.github/workflows/cloud_test_cleanup.yml
2-on:
3-  schedule:
4:    - cron: "0 2 * * *" # Run daily at 2 AM UTC
5-  # manual trigger
6-  workflow_dispatch:
```

moritzkirchnersap:
Good idea, thank you!

moritzkirchnersap:
@yeoldegrove : could you re-review, please?

yeoldegrove (approved):
LGTM

PR #4819, opened 2026-05-18
disa stig 00142

vivus-ignis:

Fixes https://github.com/gardenlinux/security/issues/297


yeoldegrove (changes requested):
SRG-OS-000354-GPOS-00142 read more like "when creating a report (do we have a tool that does this? e.g. `aureport`?) all logfiles are in a timely order and contain the same contents as from the original source".

Copying `audit.log` and checking its checksum twice does not fulfill this IMO.

vivus-ignis:
> SRG-OS-000354-GPOS-00142 read more like "when creating a report (do we have a tool that does this? e.g. `aureport`?) all logfiles are in a timely order and contain the same contents as from the original source".
> 
> Copying `audit.log` and checking its checksum twice does not fulfill this IMO.

The STIG rule description says (https://stigaview.com/products/srg-gpos/v3r2/SRG-OS-000354-GPOS-00142/): `Verify the operating system does not alter original content or time ordering of audit records when it provides a report generation capability. ` 

Which in my opinion means: "if you generate a report, the report generation process is not altering the original data on which report is based". And this is what I test here.

tests/integration/security/compliance/test_disastig_00142.py
```diff
@@ -0,0 +1,23 @@
...
+audit records when it provides a report generation capability.
+"""
+
+
+@pytest.mark.feature("log")
+@pytest.mark.booted(reason="Needs running auditd")
+@pytest.mark.root(reason="Needs to access root-owned audit.log")
+def test_audit_report_does_not_alter_original_audit_logs(shell, file):
+    with TemporaryDirectory(prefix="pytest-disa-stig-") as tmpd:
+        shutil.copy("/var/log/audit/audit.log", f"{tmpd}/audit.log")
```
Akendo, line 19: Maybe use `from shutils import copy` here.

Akendo (approved):
This test looks awesome, thank you for the great work here!

yeoldegrove:
> > SRG-OS-000354-GPOS-00142 read more like "when creating a report (do we have a tool that does this? e.g. `aureport`?) all logfiles are in a timely order and contain the same contents as from the original source".
> > Copying `audit.log` and checking its checksum twice does not fulfill this IMO.
> 
> The STIG rule description says (https://stigaview.com/products/srg-gpos/v3r2/SRG-OS-000354-GPOS-00142/): `Verify the operating system does not alter original content or time ordering of audit records when it provides a report generation capability. `
> 
> Which in my opinion means: "if you generate a report, the report generation process is not altering the original data on which report is based". And this is what I test here.

To get this approved:
- Decide if cheking for an enabled/running `auditd` would also be sufficient (IMHO: if log files are not in order this is an auditd bug).

OR

- We do not know how large `/var/log/audit/audit.log` is. It is potentially harmful to copy it to another fs or memory. Taking this into account, please copy only last x lines and run check.

Akendo (approved):

PR #4834, opened 2026-05-19
disa stig 00214

vivus-ignis:

Fixes https://github.com/gardenlinux/security/issues/169


yeoldegrove (changes requested):
From SRG-OS-000470-GPOS-00214 I read this to be an active check but this does not check anything login related.

Akendo (approved):

yeoldegrove:
To get this approved:

- Consider adding an active check like

```
su -l -c test_202606041413
grep test_202606041413 /var/log/audit/audit.log
```

OR

- decide we do not need it

Akendo:
Based on our discussion, we're not going to do the thing.

PR #4854, opened 2026-05-19
fix: remove aide service enables from disaSTIGmedium/exec.config

saubhikdattagithub:
## Problem

Building `kvm-disaSTIGmedium` fails with:
```
Failed to enable unit: aide-init.service does not exist
```

`disaSTIGmedium/exec.config` calls `systemctl enable aide-init.service` and `systemctl enable aide-check.timer`, but these units don't exist unless the `aide` feature is explicitly included in the build flavor. The `disaSTIGmedium` feature only depends on `disaSTIGlow`, not `aide`.


vivus-ignis (approved):
LGTM

PR #4878, opened 2026-05-22
Fix/try to fix pip fixture to clean up dependencies

### LINKED ISSUE #4868: failed nightly 2026-05-22
Nightly failed, see thread

moritzkirchnersap:
Nightly is broken due to sysdiff test failure

the test failed because since recently the idna package, a dependency of requests, places the file idna venv/bin/idna OR if no venv is used in /usr/local/bin/idna . While the fixture backs the python package dir and up and restores it, it does not carefully deinstall packages that were installed as dependencies for requests but removes the package dir. This didn't work in that case as the binary was places outside of the package dir. This change implements a more proper cleanup.

Fixes https://github.com/gardenlinux/gardenlinux/issues/4868

### DIFF
--- a/tests/handlers/pip.py
+++ b/tests/handlers/pip.py
@@ -12,6 +12,7 @@ def pip_requests(shell: ShellRunner):
         capture_output=True,
     )
     package_dir = out.stdout.strip("\n")
+    shell("/bin/pip3 freeze >> /tmp/pip_freeze_before_install")
 
     restore_backup = False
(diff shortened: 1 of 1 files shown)


NotTheEvilOne (approved):
LGTM so far, would be great if @yeoldegrove could have another look at it.

Technically I'ld love to see the removal of the system-wide installation and move everything just into a venv. But that's out of scope for this PR. Maybe another issue to work on in the future?

PR #4892, opened 2026-05-27
docs: fix: choosing flavors to be less repetitive

### LINKED ISSUE #4871: [ER] [Docs] flavors choosing workflow is confusing
Flavors choosing workflow (https://gardenlinux-docs.netlify.app/how-to/choosing-flavors.html) is confusing, because all the features from step 1 are then listed again in step 3. This should imply that features from Step 1 are semantically different, which is not the case. Step 1 is also covered once again in Step 4 ("Common flavors by use case") adding to the confusion.

I propose to remove Step 1, start with what's in Step 2 (determining target platform) and extend Step 3 if needed with "Go to..." links from Step 1.

yeoldegrove:

docs: fix: choosing flavors to be less repetitive

image

Fixes #4871

### DIFF
--- a/docs/how-to/choosing-flavors.md
+++ b/docs/how-to/choosing-flavors.md
@@ -41,16 +41,18 @@ complete list of all published flavors, see the
 Determine which of the primary Garden Linux use cases matches your workload.
 Each use case maps to a specific feature set.
 
-| Use Case                                             | Key Features                                                                     | Go To                                                                                               |
-| ---------------------------------------------------- | -------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------- |
-| Gardener-managed Kubernetes nodes                    | [`gardener`](/reference/features/gardener), [`_prod`](/reference/features/_prod) | [Gardener Kubernetes Nodes](/explanation/use-cases#gardener-kubernetes-nodes)                       |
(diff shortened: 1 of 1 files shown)


docs/how-to/choosing-flavors.md
```diff
@@ -41,16 +41,18 @@ complete list of all published flavors, see the
...
-| Vanilla (non-Gardener) Kubernetes nodes              | [`khost`](/reference/features/khost)                                             | [Vanilla Kubernetes Nodes](/explanation/use-cases#vanilla-kubernetes-nodes)                         |
-| OCI container base image                             | [`container`](/reference/features/container) or `bare-*`                         | [Container Base Images](/explanation/use-cases#container-base-images)                               |
-| KVM/libvirt virtualization host                      | [`vhost`](/reference/features/vhost)                                             | [Virtualization Host](/explanation/use-cases#virtualization-host)                                   |
+| Use Case                                             | Description | Key Features                                                                     | Go To                                                                                               |
```
ByteOtter: I would suggest renaming this as it references a fitting feature for the suggested use case?

```suggestion
| Use Case                                             | Description | Suitable Features                                                                     | Go To                                                                                               |
```

ByteOtter (approved):
Just one optional suggestion, but I like this change overall

docs/how-to/choosing-flavors.md
yeoldegrove: Renamed it and also changed the Step3 introduction again to match the naming.

vivus-ignis (approved):
LGTM

PR #4897, opened 2026-05-27
feat: add ci for docs

yeoldegrove:

feat: add ci for docs

This adds a CI for the new documentation hub that can be run locally and by github actions.
It uses the same checks as https://github.com/gardenlinux/docs-ng (https://github.com/gardenlinux/docs-ng/blob/07c93e83e37e22afaf0463ee1b7f3c61160d0ea5/.github/workflows/docs-checks.yml), running the aggregate with the ref/commit from the PR and running the checks.
It dispatches the https://github.com/gardenlinux/docs-ng/blob/07c93e83e37e22afaf0463ee1b7f3c61160d0ea5/.github/workflows/docs-pr.yml workflow so a PR is created/updated in the docs-ng repository.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -7,12 +7,24 @@ on:
     branches:
       - main
     paths-ignore:
-      - "docs/**"
-      - "**/README.md"
+      - ".gitignore"
+      - ".github/docs*.yml"
+      - 'docs/**'
+      - 'features/*/README.md'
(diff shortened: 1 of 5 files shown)


ByteOtter (approved):
Looks good codewise but brings up another question: Can we prevent the full image build/test CI for PRs that only touch documentation?
Would make it cleaner in my opinion ...

PR #4911, opened 2026-06-08
adjust test_server.py for nightly fix after sudo package update

### LINKED ISSUE #4910: nightly broken after package-sudo update
nightly workflow failures seen from 06-Jun https://github.com/gardenlinux/gardenlinux/actions/runs/27113540039/job/80020192116

test assertion restricts to [env_reset](https://github.com/gardenlinux/gardenlinux/blob/a893328616dfa3a40bc5ec114c85e0e589a71660/tests/integration/core/test_server.py#L271) which is now missing since `1.9.17p2-6`

saubhikdattagithub:
nightly broken on 08-Jun due to package-sudo update to `1.9.17p2-6`

Fixes #4910 

Change in upstream causing the issue: https://salsa.debian.org/sudo-team/sudo/-/commit/363f95083974fb2bd0242917d899111890d7d9c5

### DIFF
--- a/tests/integration/core/test_server.py
+++ b/tests/integration/core/test_server.py
@@ -266,9 +266,9 @@ def test_server_sudoers_keepssh_content(parse_file: ParseFile):
 
 @pytest.mark.feature("server")
 @pytest.mark.root
-def test_sudo_resets_user_environment(parse_file: ParseFile):
+def test_sudo_secure_path_is_set(parse_file: ParseFile):
     lines = parse_file.lines("/etc/sudoers")
-    assert "Defaults env_reset" in lines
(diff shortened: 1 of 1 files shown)


vivus-ignis (approved):
LGTM

yeoldegrove (approved):
Thanks for the fix. Could you reference the upstream changes in the issue somehow so that we have some record? Otherwise good to merge.

saubhikdattagithub:
> Thanks for the fix. Could you reference the upstream changes in the issue somehow so that we have some record? Otherwise good to merge.

Yes, the upstream change in the `sudo` package is now referenced in the Issue Description

PR #4945, opened 2026-06-12
add ADR for features and cname definitions

### LINKED ISSUE #4600: [Docs] standardize flavor/feature/platform/cname terms

The terms `flavor`, `feature`, `platform`, `cname` terms are used throughout the Garden Linux and need to be documented in the efforts of https://github.com/gardenlinux/gardenlinux/issues/4332.

I would like to standardize and document these terms to reference them in other parts of the user facing documentation.
Writing an ADR should be part of the documentation process.

repo: https://github.com/gardenlinux/gardenlinux
path: `docs/reference`

**What is the current state**:

The terms `flavor`, `feature`, `platform`, `cname` terms are used throughout the Garden Linux project but they are not well defined and implemented differently in repos like https://github.com/gardenlinux/builder, https://github.com/gardenlinux/python-gardenlinux-lib and https://github.com/gardenlinux/gardenlinux.

**Why is this needed**:

We need a common understanding and implementation of those terms to be transparent to the user reading documentation.

nkraetzschmar:
Resolves #4600

### DIFF
--- a/docs/architecture/decisions/0034-feature-terminology.md
+++ b/docs/architecture/decisions/0034-feature-terminology.md
@@ -23,7 +23,7 @@ Features are not independent: they form a **directed acyclic graph (DAG)**. Each
 - `features.include`: other features that are automatically pulled in when this feature is selected.
 - `features.exclude`: other features that are removed from the build if they were only transitively included (explicitly including an excluded feature is a hard build error).
 
-The DAG is resolved and validated at build time. For the precise mechanics of how feature content (package lists, file overlays, build time scripts such as `exec.config` and `exec.early`) is applied during the build, this ADR defers to the `gardenlinux/builder` documentation.[^1]
+For the precise mechanics of how the feature DAG is handled and how feature content (package lists, file overlays, build time scripts such as `exec.config` and `exec.early`) is applied during the build, this ADR defers to the `gardenlinux/builder` documentation.[^1]
 
 ### 2. Feature types
 
@@ -33,7 +33,7 @@ Every feature has exactly one type, declared in its `info.yaml`. The type is pri
 
 A feature of type `platform` represents a deployment target — the combination of hardware, firmware, and cloud or hypervisor environment the image is intended to run on. Examples: `aws`, `azure`, `gcp`, `kvm`, `baremetal`, `container`.
 
-In a well-formed build exactly one platform feature SHOULD be present. If zero or more than one platform feature is present after DAG resolution, the builder will by default treat this as an error. Explicitly opting in via `--allow-frankenstein` downgrades this to a warning and allows the build to continue; the resulting image is referred to as a **frankenstein image** and is not considered a supported configuration.
+In a well-formed build exactly one platform feature SHOULD be present. If zero or more than one platform feature is present, the build is ordinarily treated as invalid. Implementations may provide an explicit override that allows such builds to continue; the resulting image is referred to as a **frankenstein image** and is not considered a supported configuration.
 
 #### 2.2 `element`
 
@@ -41,7 +41,7 @@ A feature of type `element` represents a functional component or capability adde
 
 #### 2.3 `flag`
 
-A feature of type `flag` represents a lightweight modifier. Flags are distinguished by a leading underscore in their name (e.g. `_prod`, `_fips`, `_trustedboot`). They are intended for minor behavioural changes that do not warrant a full element and SHOULD NOT include other features.
+A feature of type `flag` represents a lightweight modifier. Flags are distinguished by a leading underscore in their name (e.g. `_prod`, `_fips`, `_trustedboot`). They are intended for minor behavioural changes that do not warrant a full element and MUST NOT include non-flag features.
 
 ### 3. Feature set
 
@@ -49,7 +49,7 @@ The **feature set** of a build is the complete set of features present after DAG
 
 ### 4. `GARDENLINUX_FEATURES_*` and `GARDENLINUX_PLATFORM` in `/etc/os-release`
 
-The builder exposes the resolved feature set to feature scripts via environment variables. These are written verbatim into `/etc/os-release` by `features/base/exec.config`:
+The build process writes the following keys into `/etc/os-release`:
 
 | Key | Content |
 |---|---|
@@ -68,32 +68,6 @@ The builder exposes the resolved feature set to feature scripts via environment 
 ### Benefits
 
 A shared written definition eliminates the ambiguity that has caused silent bugs (e.g. the `openstackbaremetal` regression described in [ADR 0020](0020-enforce-single-platform-by-default-in-builder.md)) and divergent implementations. All tooling can now be validated against a single reference.
-
-### Impact on individual components
-
-#### `gardenlinux/builder`
-
-No functional changes required. The builder's existing behaviour is consistent with the definitions in this ADR. Documentation in `docs/features.md` should be updated to cross-reference this ADR.
-
-#### `gardenlinux/gardenlinux`
-
-`features/base/exec.config` already implements the `GARDENLINUX_PLATFORM` / frankenstein logic correctly. `features/README.md` contains outdated content (notably listing `metal` as a platform when it is in fact an element; `baremetal` is the platform) and must be updated to match this ADR.
-
-#### `gardenlinux/python-gardenlinux-lib`
-
-This library contains a parallel implementation of the feature type system. Several adjustments are required:
-
-- The `CName.platform` property returns `platforms[0]` and falls back to a `frankenstein` constant controlled by the `GL_ALLOW_FRANKENSTEIN` environment variable. This logic duplicates `features/base/exec.config` but with different environment variable names and without being tied to the builder's `BUILDER_FEATURES_PLATFORMS` output. It should be aligned: the single-platform-or-frankenstein resolution belongs in the build layer (i.e. `exec.config`), not in a parsing library. The library should read `GARDENLINUX_PLATFORM` from release files rather than re-deriving it from the feature list.
-- The library's `Parser.filter_as_dict()` correctly separates features by type, which is consistent with this ADR and can be retained.
-- The `GL_ALLOW_FRANKENSTEIN` and `GL_ALLOW_MULTIPLE_PLATFORMS` environment variables in the library are not defined by the builder and should be removed in favour of reading the already-resolved `GARDENLINUX_PLATFORM` value from the release file.
-
-#### `gardenlinux/glci`
-
-glci uses the word `platform` to mean cloud provider (e.g. `"AWS"`, `"Azure"`). This is a distinct concept from the feature-type `platform` defined here. To avoid confusion it may be worth considering renaming glci's internal config field (e.g. to `cloud_provider` or `target`) in a future cleanup, though this is not required for correctness given that glci reads `GARDENLINUX_PLATFORM` from the manifest and that contract is defined in [ADR 0031](0031-builder-glci-interface.md).
-
-#### `gardenlinux/gardenlinux-update`
-
-No changes required. This tool reads `GARDENLINUX_PLATFORM` from `/etc/os-release` indirectly (via the OCI index annotations populated from the manifest). The meaning is consistent with this ADR.
 
 
 
--- a/docs/architecture/decisions/0034-feature-terminology.md
+++ b/docs/architecture/decisions/0034-feature-terminology.md
@@ -1,0 +1,100 @@
+# 34. Garden Linux Feature Terminology
+
+Date: 2026-06-12
+
+## Status
+
+Accepted
+
+## Context
+
+Garden Linux images are assembled from composable units called *features*. The main Garden Linux repository, `python-gardenlinux-lib`, and downstream tooling all use the terms *feature*, *platform*, *element*, and *flag*, but without a shared, written definition. This has led to inconsistencies — most notably in how `platform` is understood (a feature sub-type in the build system, a cloud provider name in publishing tooling, and an OCI image spec concept in update tooling) and in what the `GARDENLINUX_PLATFORM` and `GARDENLINUX_FEATURES_PLATFORMS` keys in `/etc/os-release` mean.
+
+This ADR establishes shared definitions for these terms.
+
+## Decision
+
+### 1. Feature
+
+A **feature** is the atomic building block of a Garden Linux image. It is a named directory inside the `features/` tree of the Garden Linux repository. A feature must contain an `info.yaml` file that declares at minimum its `type`. It may also contain package lists, file overlays, and build time scripts that are applied during the build.
+
+Features are not independent: they form a **directed acyclic graph (DAG)**. Each feature may declare:
+
+- `features.include`: other features that are automatically pulled in when this feature is selected.
+- `features.exclude`: other features that are removed from the build if they were only transitively included (explicitly including an excluded feature is a hard build error).
+
+The DAG is resolved and validated at build time. For the precise mechanics of how feature content (package lists, file overlays, build time scripts such as `exec.config` and `exec.early`) is applied during the build, this ADR defers to the `gardenlinux/builder` documentation.[^1]
+
+### 2. Feature types
+
+Every feature has exactly one type, declared in its `info.yaml`. The type is primarily a descriptive and user-facing classification; it does not change how the feature participates in the DAG or how its content is applied during the build. The three types are:
+
+#### 2.1 `platform`
+
+A feature of type `platform` represents a deployment target — the combination of hardware, firmware, and cloud or hypervisor environment the image is intended to run on. Examples: `aws`, `azure`, `gcp`, `kvm`, `baremetal`, `container`.
+
+In a well-formed build exactly one platform feature SHOULD be present. If zero or more than one platform feature is present after DAG resolution, the builder will by default treat this as an error. Explicitly opting in via `--allow-frankenstein` downgrades this to a warning and allows the build to continue; the resulting image is referred to as a **frankenstein image** and is not considered a supported configuration.
+
+#### 2.2 `element`
+
+A feature of type `element` represents a functional component or capability added on top of the platform. Elements are composable and multiple elements may be present in a single build. Examples: `gardener`, `server`, `cloud`, `cis`, `firewall`.
+
+#### 2.3 `flag`
+
+A feature of type `flag` represents a lightweight modifier. Flags are distinguished by a leading underscore in their name (e.g. `_prod`, `_fips`, `_trustedboot`). They are intended for minor behavioural changes that do not warrant a full element and SHOULD NOT include other features.
+
+### 3. Feature set
+
+The **feature set** of a build is the complete set of features present after DAG resolution — the transitive closure of all included features minus all excluded ones. This is distinct from the *minimal* feature set used for naming (see [ADR 0035](0035-cname-flavor-artifact-naming.md)).
+
+### 4. `GARDENLINUX_FEATURES_*` and `GARDENLINUX_PLATFORM` in `/etc/os-release`
+
+The builder exposes the resolved feature set to feature scripts via environment variables. These are written verbatim into `/etc/os-release` by `features/base/exec.config`:
+
+| Key | Content |
+|---|---|
+| `GARDENLINUX_FEATURES` | Comma-separated list of all features in the resolved feature set |
+| `GARDENLINUX_FEATURES_PLATFORMS` | Comma-separated list of all features of type `platform` |
+| `GARDENLINUX_FEATURES_ELEMENTS` | Comma-separated list of all features of type `element` |
+| `GARDENLINUX_FEATURES_FLAGS` | Comma-separated list of all features of type `flag` |
+| `GARDENLINUX_PLATFORM` | The single authoritative platform identifier for this image |
+
+`GARDENLINUX_PLATFORM` is derived from `GARDENLINUX_FEATURES_PLATFORMS` as follows: if exactly one platform feature is present, `GARDENLINUX_PLATFORM` is set to that feature's name. If more than one platform feature is present (frankenstein build), `GARDENLINUX_PLATFORM` is set to the literal string `frankenstein`.
+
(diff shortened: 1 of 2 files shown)


docs/architecture/decisions/0035-cname-flavor-artifact-naming.md
```diff
@@ -0,0 +1,144 @@
...
+
+#### `gardenlinux/builder`
+
+The builder's `parse_features` script already implements the feature encoding, sorting, and minimisation algorithm that produces the cname as defined here — what the builder currently calls `cname_base`. The following terminology changes are required:
+
+- The builder's internal variable `cname_base` maps to *cname* under this ADR.
+- The builder's internal variable `cname` (currently `{cname_base}-{arch}-{version}`) maps to *builder target* under this ADR.
+- `BUILDER_CNAME` passed to feature scripts currently contains the builder target. It should be updated to contain only the cname (feature encoding), with `BUILDER_FLAVOR` (`{cname}-{arch}`) and `BUILDER_TARGET` (`{cname}-{arch}-{version}`) introduced as separate variables.
+- The make target rule and artifact naming in `Makefile` use the builder target form and require no functional change, only renaming of internal variables.
+- `docs/features.md` and `docs/getting_started.md` should be updated to reflect the new terminology.
```
yeoldegrove: We should frame this more open, as the new documentation hub already provides way more context in regards to features.

e.g. https://gardenlinux-docs.netlify.app/explanation/flavors.html

docs/architecture/decisions/0035-cname-flavor-artifact-naming.md
```diff
@@ -0,0 +1,144 @@
...
+- `flavors.yaml` entries and the `flavors/` tooling currently produce strings of the form `{platform}-{features}-{arch}`, which matches the *flavor* definition in this ADR. No change required there.
+- References in CI workflows and scripts that construct `{cname}-{arch}-{version}-{short_commit}` and call this "the cname" should be updated to use the term *artifact base name*.
+
+#### `gardenlinux/python-gardenlinux-lib`
+
+This library has the most significant divergence from the definitions in this ADR and requires the most substantial rework:
+
+1. **`CName` class and the role of `platform`**: The library's `CName` class treats `platform` as a first-class field, exposing `CName.platform` and `CName.features["platform"]` as primary identifiers. Under this ADR, platform is simply a feature of type `platform` that sorts to the front of the cname — it has no special status in the cname string itself. The `CName` class should drop `platform` as a first-class accessor and instead expose it via the general feature-set breakdown if needed.
```
yeoldegrove: What is the harm of keeping it the way it is? As you say yourself the platform might be a feature in the end but is kind of special (e.g. exposed as `GARDENLINUX_PLATFORM` in `/etc/os-release`).
I guess @NotTheEvilOne will have an opinion on that, too ;)

docs/architecture/decisions/0035-cname-flavor-artifact-naming.md
```diff
@@ -0,0 +1,144 @@
...
+
+3. **`flavor` property**: `CName.flavor` currently returns the feature-encoding prefix (what this ADR calls the *cname*). It should be renamed to `cname`, and a new `flavor` property returning `{cname}-{arch}` introduced.
+
+4. **`flavors/parser.py`**: This separate module parses `flavors.yaml` and generates strings of the form `{target}-{feature}-{arch}`, which match the *flavor* definition here. The module name and its output are consistent with this ADR and require no renaming. However, the relationship between this module and `features/parser.py` (which also deals with something it calls "flavor") should be clarified in documentation.
+
+5. **Sorting**: The library's `get_flavor_from_feature_set` uses a `reduce` with simple underscore/hyphen logic. This does not reproduce the lexicographical topological sort defined here. For correctness, cname construction in the library must use the same algorithm as the builder, or — preferably — delegate to the builder's `parse_features` script rather than reimplementing the sort, e.g. by consuming `GARDENLINUX_CNAME` from `/etc/os-release` which is already correctly sorted and minimised by the builder.
```
yeoldegrove: Agreed that we need to use the same algorithm for sorting here. I am not a fan of pulling in the builder into the library. The library was thought the other way around, to be consumed by everything else (e.g. the builder).

docs/architecture/decisions/0035-cname-flavor-artifact-naming.md
```diff
@@ -0,0 +1,144 @@
...
+
+### 3. Builder Target
+
+A **builder target** is the flavor qualified with a version — the string passed to the builder to invoke a build:
+
+```
+builder target = {cname}-{arch}-{version}
+```
+
+Examples: `aws-gardener_prod-amd64-1877.3`, `container-amd64-1877.3`.
```
yeoldegrove, line 70: I am not completely happy with the naming "builder target" and leaving out _the source_ that is used for the build (the _commit_).
If the builder is not needing a commit as input (it might actually need it but gets it itself from the checked out directory), could we call this "builder input(s)"?

yeoldegrove: Or to make it even more universal, we could call it "build inputs", optionally allowing to add the _commit_ (as not having a commit as build input is very `builder` specific and in case another builder is used [e.g. new build system]), and calling the "artifact base name" the "build result(s) (basename)".

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+### 1. Feature
+
+A **feature** is the atomic building block of a Garden Linux image. It is a named directory inside the `features/` tree of the Garden Linux repository. A feature must contain an `info.yaml` file that declares at minimum its `type`. It may also contain package lists, file overlays, and build time scripts that are applied during the build.
+
+Features are not independent: they form a **directed acyclic graph (DAG)**. Each feature may declare:
+
+- `features.include`: other features that are automatically pulled in when this feature is selected.
+- `features.exclude`: other features that are removed from the build if they were only transitively included (explicitly including an excluded feature is a hard build error).
+
+The DAG is resolved and validated at build time. For the precise mechanics of how feature content (package lists, file overlays, build time scripts such as `exec.config` and `exec.early`) is applied during the build, this ADR defers to the `gardenlinux/builder` documentation.[^1]
```
yeoldegrove: Actually, it is not only resolved at build time... e.g. when the `python-gardenlinux-lib` resolves features.

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+
+In a well-formed build exactly one platform feature SHOULD be present. If zero or more than one platform feature is present after DAG resolution, the builder will by default treat this as an error. Explicitly opting in via `--allow-frankenstein` downgrades this to a warning and allows the build to continue; the resulting image is referred to as a **frankenstein image** and is not considered a supported configuration.
+
+#### 2.2 `element`
+
+A feature of type `element` represents a functional component or capability added on top of the platform. Elements are composable and multiple elements may be present in a single build. Examples: `gardener`, `server`, `cloud`, `cis`, `firewall`.
+
+#### 2.3 `flag`
+
+A feature of type `flag` represents a lightweight modifier. Flags are distinguished by a leading underscore in their name (e.g. `_prod`, `_fips`, `_trustedboot`). They are intended for minor behavioural changes that do not warrant a full element and SHOULD NOT include other features.
```
yeoldegrove: Could you add here why the shouldn't include other features? This is done a lot currently in `gardenlinux/features`. Shall those all change or re-typed to `element`?

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Benefits
+
+A shared written definition eliminates the ambiguity that has caused silent bugs (e.g. the `openstackbaremetal` regression described in [ADR 0020](0020-enforce-single-platform-by-default-in-builder.md)) and divergent implementations. All tooling can now be validated against a single reference.
+
+### Impact on individual components
+
+#### `gardenlinux/builder`
+
+No functional changes required. The builder's existing behaviour is consistent with the definitions in this ADR. Documentation in `docs/features.md` should be updated to cross-reference this ADR.
```
yeoldegrove: We should frame the docs update more open, as the new documentation hub already provides way more context in regards to features.

e.g. https://gardenlinux-docs.netlify.app/explanation/flavors.html#feature-based-design

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+
+#### `gardenlinux/python-gardenlinux-lib`
+
+This library contains a parallel implementation of the feature type system. Several adjustments are required:
+
+- The `CName.platform` property returns `platforms[0]` and falls back to a `frankenstein` constant controlled by the `GL_ALLOW_FRANKENSTEIN` environment variable. This logic duplicates `features/base/exec.config` but with different environment variable names and without being tied to the builder's `BUILDER_FEATURES_PLATFORMS` output. It should be aligned: the single-platform-or-frankenstein resolution belongs in the build layer (i.e. `exec.config`), not in a parsing library. The library should read `GARDENLINUX_PLATFORM` from release files rather than re-deriving it from the feature list.
+- The library's `Parser.filter_as_dict()` correctly separates features by type, which is consistent with this ADR and can be retained.
+- The `GL_ALLOW_FRANKENSTEIN` and `GL_ALLOW_MULTIPLE_PLATFORMS` environment variables in the library are not defined by the builder and should be removed in favour of reading the already-resolved `GARDENLINUX_PLATFORM` value from the release file.
```
yeoldegrove: @NotTheEvilOne Would be interested in your opinion here. I would say we have the use case to e.g. parse flavors/platforms also without the builder (access to it).

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+
+`GARDENLINUX_PLATFORM` is always a single token and is the value to be used for all platform-routing decisions (publishing, testing, image naming). `GARDENLINUX_FEATURES_PLATFORMS` is informational and reflects the raw feature graph output; it MUST NOT be used as input to publishing or routing decisions (see also [ADR 0031](0031-builder-glci-interface.md)).
+
+## Consequences
+
+### Benefits
+
+A shared written definition eliminates the ambiguity that has caused silent bugs (e.g. the `openstackbaremetal` regression described in [ADR 0020](0020-enforce-single-platform-by-default-in-builder.md)) and divergent implementations. All tooling can now be validated against a single reference.
+
+### Impact on individual components
```
yeoldegrove: The GitHub workflows of `gardenlinux` are not taken into account here. The heavilly use `python-gardenlinux-lib` and it's CLIs and need to be adapted, too. We might also need to backport this to older release branches. @NotTheEvilOne will have more insights.

yeoldegrove (changes requested):
Thanks for writing this all down and also for having the complete picture in mind what needs to be adapted.
I have a few comments and suggestions that I would like to address.

docs/architecture/decisions/0035-cname-flavor-artifact-naming.md
nkraetzschmar: The reason the builder target should not contain the commit is that a different commit is not something the builder can target. The commit is fixed by whatever happens to be currently checked out. Whereas everything else: feature set, architecture, and version, can be decided at build time. So the builder target is everything that fully specifies what the builder is supposed to build. Including the commit in the builder target would be ambiguous in cases where the commit provided there does not match what's actually checked out.

docs/architecture/decisions/0034-feature-terminology.md
nkraetzschmar: There we might want to re-align if having multiple places resolving this is a good idea. I'd say it should be enough for only the builder to resolve this and then the python-lib to simply use the already resolved outputs that the builder produces in the `os-release` file. Re-resolving this after the build is pretty fragile.

docs/architecture/decisions/0034-feature-terminology.md
nkraetzschmar: It was originally intended that flags should not include other features[^1], but it's not actually enforced, thus only a _"should not"_ and not a _"must not"_ 😅

[^1]: https://github.com/gardenlinux/builder/blob/6026df526f40948aab159fd8847a327560556f64/docs/features.md?plain=1#L12

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Decision
+
+### 1. Feature
+
+A **feature** is the atomic building block of a Garden Linux image. It is a named directory inside the `features/` tree of the Garden Linux repository. A feature must contain an `info.yaml` file that declares at minimum its `type`. It may also contain package lists, file overlays, and build time scripts that are applied during the build.
+
+Features are not independent: they form a **directed acyclic graph (DAG)**. Each feature may declare:
+
+- `features.include`: other features that are automatically pulled in when this feature is selected.
```
NotTheEvilOne, line 23: `features.include` and `features.exclude` are implementation specific and should therefore not be part of this ADR.

docs/architecture/decisions/0034-feature-terminology.md
NotTheEvilOne: It should more be the other way around. As the python-lib can not differentiate between trusted and untrusted input it *has to* validate all input in it's own. The builder implementation works great for it's purpose but can not be reused. It's code was merged to the python-lib long time ago.

NotTheEvilOne: Same as for above. The ADR section contains too much implementation detail about specific files. Please consider moving it to a separate document refering to this ADR.

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+- `features.exclude`: other features that are removed from the build if they were only transitively included (explicitly including an excluded feature is a hard build error).
+
+The DAG is resolved and validated at build time. For the precise mechanics of how feature content (package lists, file overlays, build time scripts such as `exec.config` and `exec.early`) is applied during the build, this ADR defers to the `gardenlinux/builder` documentation.[^1]
+
+### 2. Feature types
```
NotTheEvilOne, line 28: Section 1 already defines a feature. Section 2 is adding details but reiterates 1 as well.

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+
+Every feature has exactly one type, declared in its `info.yaml`. The type is primarily a descriptive and user-facing classification; it does not change how the feature participates in the DAG or how its content is applied during the build. The three types are:
+
+#### 2.1 `platform`
+
+A feature of type `platform` represents a deployment target — the combination of hardware, firmware, and cloud or hypervisor environment the image is intended to run on. Examples: `aws`, `azure`, `gcp`, `kvm`, `baremetal`, `container`.
+
+In a well-formed build exactly one platform feature SHOULD be present. If zero or more than one platform feature is present after DAG resolution, the builder will by default treat this as an error. Explicitly opting in via `--allow-frankenstein` downgrades this to a warning and allows the build to continue; the resulting image is referred to as a **frankenstein image** and is not considered a supported configuration.
```
NotTheEvilOne: The frankenstein parameter is a funny addition but again implementation specific.

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+
+A feature of type `flag` represents a lightweight modifier. Flags are distinguished by a leading underscore in their name (e.g. `_prod`, `_fips`, `_trustedboot`). They are intended for minor behavioural changes that do not warrant a full element and SHOULD NOT include other features.
+
+### 3. Feature set
+
+The **feature set** of a build is the complete set of features present after DAG resolution — the transitive closure of all included features minus all excluded ones. This is distinct from the *minimal* feature set used for naming (see [ADR 0035](0035-cname-flavor-artifact-naming.md)).
+
+### 4. `GARDENLINUX_FEATURES_*` and `GARDENLINUX_PLATFORM` in `/etc/os-release`
+
+The builder exposes the resolved feature set to feature scripts via environment variables. These are written verbatim into `/etc/os-release` by `features/base/exec.config`:
```
NotTheEvilOne: Implementation detail: It should not be part where values come from.

docs/architecture/decisions/0034-feature-terminology.md
NotTheEvilOne: Implementation details oft this ADR should be part of a design documentation file in the `builder` repository.

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+
+No functional changes required. The builder's existing behaviour is consistent with the definitions in this ADR. Documentation in `docs/features.md` should be updated to cross-reference this ADR.
+
+#### `gardenlinux/gardenlinux`
+
+`features/base/exec.config` already implements the `GARDENLINUX_PLATFORM` / frankenstein logic correctly. `features/README.md` contains outdated content (notably listing `metal` as a platform when it is in fact an element; `baremetal` is the platform) and must be updated to match this ADR.
+
+#### `gardenlinux/python-gardenlinux-lib`
+
+This library contains a parallel implementation of the feature type system. Several adjustments are required:
```
NotTheEvilOne: As the ADR describes what values in os-release have what meaning the implementation detail should not be part of it. Furthermore environment variables do not need alignment as they are a black box for the outcome of this ADR.

docs/architecture/decisions/0034-feature-terminology.md
```diff
@@ -0,0 +1,100 @@
...
+
+- The `CName.platform` property returns `platforms[0]` and falls back to a `frankenstein` constant controlled by the `GL_ALLOW_FRANKENSTEIN` environment variable. This logic duplicates `features/base/exec.config` but with different environment variable names and without being tied to the builder's `BUILDER_FEATURES_PLATFORMS` output. It should be aligned: the single-platform-or-frankenstein resolution belongs in the build layer (i.e. `exec.config`), not in a parsing library. The library should read `GARDENLINUX_PLATFORM` from release files rather than re-deriving it from the feature list.
```
NotTheEvilOne: A parsing library intended for reuse has to validate input no matter of the source for consistency and validation of parsed data. It ensures the correct implementation in sources used.

NotTheEvilOne (changes requested):

docs/architecture/decisions/0034-feature-terminology.md
nkraetzschmar: I'm not sure I follow what's meant by _"untrusted input"_. If the python lib is used in tasks like `publish_s3.yml` or `publish_oci_containers.yml` then its job should be to publish the artifacts the builder produced. For this job the meta data the builder produced, as in what features it built, what cname it is, etc. are to be treated as authoritative. The upload action should NOT overrule the cname or other info determined during build.

NotTheEvilOne: As long as a release file is part of the build artifacts no revalidation is done. In the other hand the Python library supports basic feature set parsing with and without a `features` directory. Some minor validations are deactivated without a `features` directory in that case.

docs/architecture/decisions/0035-cname-flavor-artifact-naming.md
NotTheEvilOne: As the builder may not check out a different commit it can validate that a given commit (`!= local`) matches the repository tree. Tags and versions are moving targets not ensuring reproducibility. Only a commit hash from `./get_commit` or `COMMIT` ensures the reproducability expected from Garden Linux builds. Therefore a given commit should be mandatory and should be prefixed by a tag for easier reference.

docs/architecture/decisions/0035-cname-flavor-artifact-naming.md
```diff
@@ -0,0 +1,144 @@
...
+
+### 4. Artifact Base Name
+
+The **artifact base name** is the builder target plus the short commit hash:
```
NotTheEvilOne: The name seems a bit long. We could refer to artifact name here instead?

docs/architecture/decisions/0035-cname-flavor-artifact-naming.md
NotTheEvilOne: While the whole class name might need change the platform property is still part of a cname and allows programatically access to it.

docs/architecture/decisions/0035-cname-flavor-artifact-naming.md
NotTheEvilOne: The library distiguishes between a mode including vor not using a `features` directory. The lexicographical topological sort is used when the features directory is present. Otherwise some functions expect input is already the *minimal* feature set described above.

NotTheEvilOne (changes requested):

docs/architecture/decisions/0034-feature-terminology.md
NotTheEvilOne: It depends in what is required and used for building (= no backporting) or publishing (e.g. release file). A versioned release file might help preventing any backports.

docs/architecture/decisions/0034-feature-terminology.md
NotTheEvilOne: I missed this is part of the `info.yaml` - ignoring :)

yeoldegrove (approved):
Thanks for the adaptions. This is good to merge from my side now.

docs/architecture/decisions/0035-cname-flavor-artifact-naming.md
```diff
@@ -0,0 +1,104 @@
...
+Accepted
+
+## Context
+
+The terms `cname`, `flavor`, and related identifiers are used across the Garden Linux project — in the builder, the main repository, `python-gardenlinux-lib`, `glci`, and `gardenlinux-update` — but each component has developed its own interpretation. The following inconsistencies exist today:
+
+### `cname`
+
+- **`gardenlinux/builder`**: defines cname as `{feature-encoding}-{arch}-{version}`.
+- **`gardenlinux/gardenlinux`**: two conflicting usages within the same repo — `GARDENLINUX_CNAME` in `/etc/os-release` correctly contains `{feature-encoding}-{arch}-{version}` (no commit), matching the builder; but `.build/` output filenames use `{feature-encoding}-{arch}-{version}-{commit}` as their prefix, and this form has informally also been referred to as "the cname" in CI logs and scripts.
```
NotTheEvilOne, line 16: We should replace "... correctly contains ..." with just "... contains ..." as ADRs are neutral and most likely previous decisions were intentionally valid for the time being.

PR #5009, opened 2026-07-10
create issue if nightly fails

vivus-ignis:
**What this PR does / why we need it**: To create GH issues for tracking failed nighly runs

**Special notes for your reviewer**: The job creating GH issue tries to enrich it with some data that can useful for troubleshooting.

Example of created issue: https://github.com/vivus-ignis/gardenlinux-dev/issues/6

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -12,7 +12,7 @@ jobs:
     uses: ./.github/workflows/build.yml
     permissions:
       id-token: write # TODO: 3215 elevated permissions for included workflow
-      actions: write  # TODO: 3215 elevated permissions for included workflow
+      actions: write # TODO: 3215 elevated permissions for included workflow
       packages: write # TODO: 3215 elevated permissions for included workflow
     with:
       version: now
@@ -23,6 +23,7 @@ jobs:
       aws_kms_role: ${{ secrets.KMS_SIGNING_IAM_ROLE }}
       aws_oidc_session: ${{ secrets.AWS_OIDC_SESSION }}
       secureboot_db_kms_arn: ${{ secrets.SECUREBOOT_DB_KMS_ARN }}
+
   build_retry:
     needs: build
     if: ${{ failure() && needs.build.result == 'failure' }}
@@ -40,6 +41,7 @@ jobs:
 
             const gitHubRef = "${{ github.head_ref == '' && github.ref_name || github.head_ref }}";
             return await gitHubLib.dispatchRetryWorkflow(core, github.rest.actions, context, gitHubRef, 5);
+
   test:
     needs: build
     name: Test
@@ -69,6 +71,7 @@ jobs:
       alibaba_cloud_oidc_provider_arn: ${{ secrets.ALIBABA_CLOUD_OIDC_PROVIDER_ARN }}
       alibaba_cloud_region: ${{ secrets.ALIBABA_CLOUD_REGION }}
       tf_encryption: ${{ secrets.TF_ENCRYPTION }}
+
   test_retry:
     needs: test
     if: ${{ failure() && needs.test.result == 'failure' }}
@@ -86,6 +89,7 @@ jobs:
 
             const gitHubRef = "${{ github.head_ref == '' && github.ref_name || github.head_ref }}";
             return await gitHubLib.dispatchRetryWorkflow(core, github.rest.actions, context, gitHubRef, 3);
+
   upload_workflow_data:
     needs: [build, test]
     name: Store workflow data
@@ -107,6 +111,7 @@ jobs:
           name: workflow-data
           path: workflow_data.json
           if-no-files-found: error
+
   publish_ghcr:
     needs: upload_workflow_data
     name: Publish nightly to ghcr.io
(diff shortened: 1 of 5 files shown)


Akendo (commented):
Looks good to me, but I do not have much say about the code. I perfe to have a Python script and JavaScript is something I keep my hands off. Maybe @yeoldegrove can assist here if that code does well.

Akendo:
If possible, I would rather *not* see this logic be exclusive to GitHub worker, and instead is a local script that is only invoked by GHA.

.github/workflows/nightly.yml
```diff
@@ -107,24 +111,175 @@ jobs:
...
     uses: ./.github/workflows/publish_s3.yml
     with:
       run_id: ${{ github.run_id }}
     permissions:
-      actions: write  # TODO: 3215 elevated permissions for included workflow
+      actions: write # TODO: 3215 elevated permissions for included workflow
       id-token: write # TODO: 3215 elevated permissions for included workflow
     secrets: inherit
+
+  create_issue_on_failure:
```
NotTheEvilOne, line 138: This is the wrong place for the creation of issues. As the retry is triggered after a run failed this should result in multiple issues being created with each failing run. The logic must be placed in `.github/workflows/github.mjs` under `retryWorkflow()` triggered by `.github/workflows/github_rerun_workflow.yml`.

.github/workflows/nightly.yml
```diff
@@ -107,24 +111,175 @@ jobs:
...
+
+            const apiHeaders = {
+              'Authorization': `Bearer ${process.env.GITHUB_TOKEN}`,
+              'Accept': 'application/vnd.github+json',
+              'X-GitHub-Api-Version': '2022-11-28',
+            };
+
+            const getLastLogLines = async (jobId) => {
+              try {
+                const resp = await fetch(
```
NotTheEvilOne: Access to GitHub API is provided by e.g. `github.rest.actions` abstraction layer (called "octokit.js") and should be used.

.github/workflows/nightly.yml
```diff
@@ -107,24 +111,175 @@ jobs:
...
+            echo EOF
+          } >> "$GITHUB_OUTPUT"
+      - uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
+        env:
+          COMPARE_OUTPUT: ${{ steps.compare_versions.outputs.compare_output }}
+          NEW_VERSION: ${{ steps.compare_versions.outputs.new_version }}
+          OLD_VERSION: ${{ steps.compare_versions.outputs.old_version }}
+        with:
+          script: |
+            const needs = JSON.parse(process.env.NEEDS_JSON);
```
NotTheEvilOne, line 177: This code faces multiple issues:
- no separation of logic and presentation
- good intention of adding additional details about what happened within the last 24 hours but I think it may produce too much data
- it makes too many assumptions that an issue is caused and will be fixed within the first 24 hours (the second automatically created issue after >24 hours will not contain relevant changes anymore)
- most of the time the relevant log data gives a first hint where to look for issues
- as API calls may fail the less calls needed to create an issue the more useful issues are automatically opened for failing nightly runs

NotTheEvilOne (commented):
While the idea may be helpful the solution should be changed as following. I'll not finish this review with "request changes" as I'll most likely not review changes in the next three weeks but would recommend to further fine tune the PR before reviewing it again by a colleague.

.github/workflows/nightly.yml
vivus-ignis: Added a check that prevents multiple issues from appearing for the same pipeline run.

yeoldegrove (changes requested):
I would like to suggest, as already discussed, to drop the gh actions code completely.
An easily extensible and debuggable python script is the way to go. In the current implementation the python part is merely an entrypoint that calls node/Oktokit. It was meant to be the other way around.

What I like:
- `hack/compare-apt-repo-versions.sh` is called and added to the issue (could be easily implemented in python, too)
- "Pull requests merged in the last 24 hours" --> Maybe do not deliver a list but an url/query that filters this in on github.

What I would like to add/change:
- we need a `--dry-run` to test things.
- Failed jobs might potentially have many API calls and produce a large output. This might need filtering in addition to limit it to 10 lines per job.
- We already have a test artifact that contains the pytest JUnit XML logs. This could easily be parsed and added and would provide great insights (see screenshots).

Image

Image

- for accessing github and version related parsing, python-gardenlinux-lib already implements a lot that could be used

I imagine a UI with similar flags that you already suggested:
```
❯ bin/create-nightly-issue --help
usage: create-nightly-issue [-h] [--repo OWNER/REPO] [--workflow FILE] [--run-id ID] [--branch NAME] [--dry-run] [--update] [--label LABEL] [--title TEXT] [--token TOKEN] [-v]

Open a GitHub issue for the latest nightly workflow run.

options:
  -h, --help         show this help message and exit
  --repo OWNER/REPO  Target repository (default: gardenlinux/gardenlinux)
  --workflow FILE    Workflow file name (default: nightly.yml)
  --run-id ID        Use a specific run instead of the latest
  --branch NAME      Restrict latest-run lookup to this branch (default: main)
  --dry-run          Print the issue title and body; do not create or update an issue
  --update           Update an existing issue for this run instead of skipping
  --label LABEL      Extra labels to apply (repeatable; 'bug' is always applied)
  --title TEXT       Override the computed issue title
  --token TOKEN      GitHub token (defaults to GITHUB_TOKEN env var)
  -v, --verbose      Enable debug logging
```

PR #5020, opened 2026-07-12
Use 0666 for /dev/kvm for the qemu tests

5kt:
Use 0666 permissions for /dev/kvm so that our qemu tests can use kvm acceleration. (nested virtualization is allowed on the runners[^1])

Fixes #

[^1]: https://github.com/orgs/community/discussions/160591#discussioncomment-9018826


yeoldegrove (approved):
Thanks for improving this. Let's have one more green CI run before merging, please.

PR #5034, opened 2026-07-14
ssh key lower for openssh 10.4 compatibility

### LINKED ISSUE #5033: openssh 10.4 tests/plugin issue while`2150.7.0` release
As openssh is apt_src (Debian Snapshot) and backport is via Salsa, an openssh compatibilty issue is discovered in 10.4p1-1 with our tests/plugins/sshd.py

saubhikdattagithub:
As openssh is apt_src (Debian Snapshot) and backport (rel-2150) is via Salsa, an `openssh` compatibility issue is discovered in 10.4p1-1 with our `tests/plugins/sshd.py`

Fixes #5033

### DIFF
--- a/tests/plugins/sshd.py
+++ b/tests/plugins/sshd.py
@@ -18,6 +18,7 @@ class Sshd:
             parts = line.split(maxsplit=1)
             if len(parts) == 2:
                 key, value = parts
+                key = str.casefold(key)
                 # If key already exists, convert to set and append
                 if key in self._sshd_config:
                     if isinstance(self._sshd_config[key], set):
@@ -27,7 +28,7 @@ class Sshd:
                 else:
                     self._sshd_config[key] = value
             elif len(parts) == 1:
-                key = parts[0]
+                key = str.casefold(parts[0])
                 self._sshd_config[key] = None
 
     def get_config(self) -> dict:
--- a/tests/plugins/sshd.py
+++ b/tests/plugins/sshd.py
@@ -18,7 +18,7 @@ class Sshd:
             parts = line.split(maxsplit=1)
             if len(parts) == 2:
                 key, value = parts
-                key = str.casefold(key)
+                key = key.lower()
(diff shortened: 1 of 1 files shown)


tests/plugins/sshd.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, shell: ShellRunner):
             parts = line.split(maxsplit=1)
             if len(parts) == 2:
                 key, value = parts
+                key = str.casefold(key)
```
moritzkirchnersap: Had to look up what "casefold" does, but after reading the documentation it is a "more aggressive" lower - which ... I don't know what it means but I guess let's go with casefold then.

moritzkirchnersap (approved):

tests/plugins/sshd.py
saubhikdattagithub: OpenSSH 10.4p1 changed the output of `sshd -T` from lowercase keys (e.g. hostkey, permitrootlogin) to mixed case (e.g. HostKey, PermitRootLogin). The tests look up keys with lowercase strings like get_config_section("hostkey"), so we need to normalize the keys at storage time.

Before (broken for 10.4):
  - sshd -T outputs HostKey /etc/ssh/ssh_host_ed25519_key
  - Plugin stored it as-is: {"HostKey": "/etc/ssh/ssh_host_ed25519_key"}
  - Test calls get_config_section("hostkey") → looks up "hostkey" in dict → not found → returns None
  - All SSH config assertions failed

  After (fixed for both 10.3 and 10.4):
  - sshd -T outputs HostKey /etc/ssh/ssh_host_ed25519_key
  - Plugin casefolks at storage time: {"hostkey": "/etc/ssh/ssh_host_ed25519_key"}
  - Test calls get_config_section("hostkey") → looks up "hostkey" → found → returns correct value

saubhikdattagithub: Thanks for the review, let me check using "lower" only

PR #5036, opened 2026-07-15
material event to check with auditd

vivus-ignis:
**What this PR does / why we need it**: Introduces a material event to be checked with auditd for the disa stig rule 00191 compliance test

Fixes failing pipeline runs like here https://github.com/gardenlinux/gardenlinux/actions/runs/29408189927/job/87331402585?pr=4703


tests/integration/security/compliance/test_disastig_00218.py
```diff
@@ -6,25 +6,20 @@
...
 JOURNAL_FILE = "/tmp/journal_output_gl.txt"
 TIME_FILE = "/tmp/audit_start_time"
 
 
 @pytest.fixture
-def concurrent_login_environment(shell: ShellRunner):
-    shell(f"id {TEST_USER} || useradd -m {TEST_USER}")
+def concurrent_login_environment(shell: ShellRunner, audit_user):
     yield
     shell(f"pkill -9 -u {TEST_USER}", ignore_exit_code=True)
```
yeoldegrove: How about adding the `pkill -9` to the `audit_user` fixture before deleting the user?

tests/integration/security/compliance/test_disastig_00218.py
```diff
@@ -6,25 +6,20 @@
...
     shell("sleep 1")
-    shell(f"userdel -r {TEST_USER}")
     shell(f"rm -f {OUTPUT_FILE} {JOURNAL_FILE} {TIME_FILE}")
```
yeoldegrove: We could potentially eliminate this whole fixture by using https://docs.python.org/3/library/tempfile.html and then just call the new `audit_user` directly in the test.

tests/handlers/audit_user.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@pytest.fixture
+def audit_user(shell: ShellRunner):
+    shell(f"id {TEST_USER} || useradd -m -s /bin/sh {TEST_USER}")
+    yield TEST_USER
+    shell(f"userdel -r {TEST_USER}", ignore_exit_code=True)
+    shell(
+        "rm -f /etc/group- /etc/gshadow- /etc/passwd- /etc/shadow- /etc/subgid- /etc/subuid-"
+    )
```
yeoldegrove, line 16: Great that this is now it's own handler!

tests/integration/security/compliance/test_disastig_00191.py
```diff
@@ -1,22 +1,26 @@
...
 @pytest.mark.security_id(203752)
-@pytest.mark.feature("not container and not lima")
+@pytest.mark.feature("disaSTIGmedium")
 @pytest.mark.booted(reason="requires audit subsystem running")
-@pytest.mark.root(reason="required to read audit logs")
-def test_invalid_input_handling_is_audited(shell):
-    """Verify ausearch -ts recent output contains 'success=no', 'res=failed' or 'invalid'."""
-    result = shell("ausearch -ts recent", capture_output=True)
-    stdout = result.stdout
+@pytest.mark.root(reason="useradd and ausearch require root privileges")
```
yeoldegrove, line 17: Please add the `@pytest.mark.modify` marker https://docs.gardenlinux.org/reference/testing/developing-tests.html#_3-be-very-strict-about-declaring-if-they-mutate-system-state.

tests/integration/security/compliance/test_disastig_00218.py
yeoldegrove, line 18: Please add the `@pytest.mark.modify` marker https://docs.gardenlinux.org/reference/testing/developing-tests.html#_3-be-very-strict-about-declaring-if-they-mutate-system-state.

yeoldegrove (changes requested):

yeoldegrove (approved):
Thanks for the adaptions.

PR #5051, opened 2026-07-21
suppress stdout for diki verbose

### LINKED ISSUE #5050: suppress stdout for diki
as discussed in the meeting, stdout gets leaked for few tests and we need to suppress stdout for diki

saubhikdattagithub:
few tests are leaking stdout appearing in diki verbose output

Fixes #5050

### DIFF
--- a/.github/workflows/build_requirements.yml
+++ b/.github/workflows/build_requirements.yml
@@ -38,7 +38,7 @@ jobs:
           submodules: true
       - name: Set signing environment
         id: signing_environment
-        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
         with:
           script: |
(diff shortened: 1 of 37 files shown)


vivus-ignis (changes requested):
I propose to do this instead:

1. avoid doing multiple "in" comparisons with boolean logic all in one assert call (like in `tests/integration/security/compliance/test_disastig_00022.py`) and use simple regexes instead
2. add something like this to conftest.py:
```python
def pytest_assertrepr_compare(op, left, right):
    if op == "in" and isinstance(right, str) and len(right) > 200:
        return [
            f"Assertion failed: '{left}' not found in comparison string.",
            f"Comparison string (truncated): {right[:500]}...",
        ]
```
this will make nicer output without obscuring tests code with throwaway variables
more on `pytest-assertrepr_compare`: https://docs.pytest.org/en/7.1.x/how-to/assert.html#defining-your-own-explanation-for-failed-assertions

vivus-ignis (approved):
LGTM

PR #5053, opened 2026-07-21
Various small fixes

5kt:
For the install feature, the ESP should use a sector size of 512 otherwise syslinux will fail the install step in case of legacy mode.

Building _kvm_iso_ for example is failing as the dracut dmsquash-live module pulls in the dm module which requires dmsetup, so it's now included via pkg.include.

When disabling ignition on first boot, add a conditional based on the kernel cmdline - the old conditional based on the existence of _/run/ignition.env_ is not really correct as that file always exists - it's created by an ignition generator.


Akendo (approved):

yeoldegrove (approved):
As discussed, please add a small documentation in `features/_install/README.md`. Otherwise good to merge.

PR #5101, opened 2026-08-03
try again for disa_stig_version in diki report

saubhikdattagithub:
What this PR does / why we need it:
as discussed with diki team, we need to append a the DISA STIG revision info in our compliance tests

Which issue(s) this PR fixes:
Fixes https://github.com/gardenlinux/gardenlinux/issues/5043

PR #5104, opened 2026-08-03
fix: disa_stig_version showing as None in diki JUnit XML reports

saubhikdattagithub:
## Summary

- Fixes `disa_stig_version` property showing as `None` in diki JUnit XML reports
- Root cause: `config.addini()` does not exist in pytest 9.x — ini options must be registered via `parser.addini()` in `pytest_addoption`
- Adds a hardcoded default (`"General Purpose Operating System STIG V3R2"`) so the value is always present even when `pytest.ini` is not included in the test distribution

## Changes

- `tests/plugins/security_id.py`: adds `pytest_addoption(parser)` hook using `parser.addini()` to register the `disa_stig_version` ini option with a default value; reads it in `pytest_collection_modifyitems` via `config.getini()`

## Test plan

- [ ] Verify diki report shows `disa_stig_version = "General Purpose Operating System STIG V3R2"` in `<properties>` for `test_disastig_*` tests
- [ ] Verify CI lint/test-utils passes (no `AttributeError: 'Config' object has no attribute 'addini'`)

### DIFF
--- a/tests/plugins/security_id.py
+++ b/tests/plugins/security_id.py
@@ -2,6 +2,16 @@ from typing import List
 
 import pytest
 
+DISA_STIG_VERSION = "General Purpose Operating System STIG V3R2"
+
+
+def pytest_addoption(parser: pytest.Parser):
+    parser.addini(
+        "disa_stig_version",
+        default=DISA_STIG_VERSION,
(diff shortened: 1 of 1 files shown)


saubhikdattagithub:
Closing — changes belong in PR #5101 instead.

PR #5116, opened 2026-08-04
Add ADR037 superceding ADR0031

Gchbg:

This PR extends the interface between the builder and GLCI, paving the way to OCI multiarch image publishing support in GLCI. It adds a single field called `publishing_group`:

The canonical example is containers. Say the goal is to publish a multiarch container. That is not built as *one*artifact; rather it is built as *N* artifacts, one per architecture, each with its own manifest - e.g. the amd64 portion and the arm64 portion. This requires new semantics for GLCI - it has to publish all, and then fuse their outputs which in the container case means create a container OCI-manifest that points to all - that's how multiarch containers work.

GLCI needs to be told to do this for some artifacts, but not most, and it needs to be told which ones go together. The way to tell it is to add to each manifest (= for each arch) of the container a `publishing_group` string. It will be equal to the cname, so that the amd64+arm64 containers get the same value, while any other pair of containers get a different value.

*Example*

Flavors `container-some_features-amd64` and `container-some_features-arm64` both get `publishing_group: container-some_features`. Same string, published together and fused! On the other hand `aws-gardener_prod-amd64` and `aws-gardener_prod-arm64` get no/empty `publishing_group`. They are treated as completely independent images as usual.

Adding this `publishing_group` field is the entire amendment presented here.

**Definition of Done:**

### DIFF
--- a/docs/reference/adr/0037-builder-glci-interface.md
+++ b/docs/reference/adr/0037-builder-glci-interface.md
@@ -1,18 +1,20 @@
 ---
-title: "ADR 0031: S3 Artifact and Metadata Publishing Contract Between Builder and GLCI"
+title: "ADR 0037: S3 Artifact and Metadata Publishing Contract Between Builder and GLCI"
 github_org: gardenlinux
 github_repo: gardenlinux
-github_source_path: docs/reference/adr/0031-builder-glci-interface.md
-github_target_path: docs/reference/adr/0031-builder-glci-interface.md
+github_source_path: docs/reference/adr/0037-builder-glci-interface.md
+github_target_path: docs/reference/adr/0037-builder-glci-interface.md
 ---
 
-# ADR 0031: S3 Artifact and Metadata Publishing Contract Between Builder and GLCI
+# ADR 0037: S3 Artifact and Metadata Publishing Contract Between Builder and GLCI
 
-Date: 2026-02-05
+Date: 2026-08-04
 
 ## Status
 
-Superseded by [ADR 37](./0037-builder-glci-interface.md)
+Accepted
+
+Supersedes [ADR 31](./0031-builder-glci-interface.md)
 
 ## Context
 
@@ -27,9 +29,13 @@ Superseded by [ADR 37](./0037-builder-glci-interface.md)
   + Which builder outputs are authoritative inputs for that metadata
 - Currently, the upload to S3 and metadata generation is handled by `python-gardenlinux-lib`.
 - Misalignment between builder output and how metadata is derived has caused issues, most notably while consolidating `openstack` / `openstackbaremetal` into `openstack` and `openstack-metal`.
+- [ADR 31](./0031-builder-glci-interface.md) established this interface. It assumed a one-to-one relationship between a build and a published artifact, and it defined the default value of any non-required requirements field as `false`.
+- Two needs have since emerged that the ADR 31 contract cannot express:
+  + Some artifacts must be published as a single combined artifact assembled from several builds — the motivating case being container images, published as one multi-architecture image. The interface must carry abuild-provided signal of which builds belong together.
+  + Expressing that signal requires an optional field that is not a boolean, but the ADR 31 default-value rule only defines a default for booleans.
 - A clear documented interface between **build artifacts** and **GLCI inputs** is required to avoid future ambiguity and regressions.
 
-This ADR defines that interface.
+This ADR supersedes ADR 31 and restates the interface in full so that it remains the single authoritative document.
 
 ## Decision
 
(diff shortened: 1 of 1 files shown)


docs/reference/adr/0037-builder-glci-interface.md
```diff
@@ -0,0 +1,192 @@
...
+
+#### Platform semantics
+
+- `GARDENLINUX_PLATFORM` is the **one and only platform value** used for publishing.
+- This guarantees exactly one platform value, or a deliberate “frankenstein” value if a non-standard image was built.
+- `GARDENLINUX_FEATURES_PLATFORMS` **must not** be used as input to any publishing or routing decision:
+  + It may contain zero, one, or multiple values.
+  + It reflects feature graph structure, not publishing semantics.
+
+### 1.2 Requirements File (`*.requirements`)
```
Akendo, line 91: I feel like this ADR should only contain the differences or changes to the existing ADR? Makes it harder to see what has changed from the previous ADR. It would be more helpful in my view to at least highlight the changes.

docs/reference/adr/0037-builder-glci-interface.md
```diff
@@ -0,0 +1,192 @@
...
+- Currently, the upload to S3 and metadata generation is handled by `python-gardenlinux-lib`.
+- Misalignment between builder output and how metadata is derived has caused issues, most notably while consolidating `openstack` / `openstackbaremetal` into `openstack` and `openstack-metal`.
+- [ADR 31](./0031-builder-glci-interface.md) established this interface. It assumed a one-to-one relationship between a build and a published artifact, and it defined the default value of any non-required requirements field as `false`.
+- Two needs have since emerged that the ADR 31 contract cannot express:
+  + Some artifacts must be published as a single combined artifact assembled from several builds — the motivating case being container images, published as one multi-architecture image. The interface must carry abuild-provided signal of which builds belong together.
+  + Expressing that signal requires an optional field that is not a boolean, but the ADR 31 default-value rule only defines a default for booleans.
+- A clear documented interface between **build artifacts** and **GLCI inputs** is required to avoid future ambiguity and regressions.
```
Akendo, line 36: You add the `publishing_group` but I feel like this should explain in the context here.

docs/reference/adr/0037-builder-glci-interface.md
```diff
@@ -0,0 +1,192 @@
...
+- For this workflow to be reliable, a **clear, documented interface** is required that specifies:
+  + Which files are uploaded
+  + Where they are uploaded in S3
+  + Which metadata is produced
+  + Which builder outputs are authoritative inputs for that metadata
+- Currently, the upload to S3 and metadata generation is handled by `python-gardenlinux-lib`.
+- Misalignment between builder output and how metadata is derived has caused issues, most notably while consolidating `openstack` / `openstackbaremetal` into `openstack` and `openstack-metal`.
+- [ADR 31](./0031-builder-glci-interface.md) established this interface. It assumed a one-to-one relationship between a build and a published artifact, and it defined the default value of any non-required requirements field as `false`.
+- Two needs have since emerged that the ADR 31 contract cannot express:
+  + Some artifacts must be published as a single combined artifact assembled from several builds — the motivating case being container images, published as one multi-architecture image. The interface must carry abuild-provided signal of which builds belong together.
```
Akendo, line 34: Can you maybe write down an example?

docs/reference/adr/0037-builder-glci-interface.md
```diff
@@ -0,0 +1,192 @@
...
+
+#### Fields
+
+| Field              | Type   | Required | Meaning                                                                                               |
+| ------------------ | ------ | -------- |-------------------------------------------------------------------------------------------------------|
+| `arch`             | string | yes      | Architecture of the image                                                                             |
+| `publishing_group` | string | no       | Opaque group identifier, builds sharing a value are published as one artifact, empty means standalone |
+| `secureboot`       | bool   | no       | Secure Boot is required                                                                               |
+| `tpm2`             | bool   | no       | TPM 2.0 is required                                                                                   |
+| `uefi`             | bool   | no       | UEFI is required (not merely supported, for hybrid legacy and UEFI images this must be false)         |
```
Akendo, line 103: Should we maybe add another placeholder field to have the ability to provide additional data at will? Changes will occur, and having something like a placeholder in place could be helpful?

docs/reference/adr/0037-builder-glci-interface.md
Gchbg: In terms of review yes, but in terms of having one document with the current interface, it makes sense to have the entire interface complete in one spot. I can tell you definitively that there are exactly TWO changes:
1) The addition of the publishing_group field.
2) The change of the default value of all optional fields from `false` (because before publishing_group all requirements-based optional fields were booleans) to _the zero value of the type_ which for booleans is still `false` but accommodates the empty string for strings.

docs/reference/adr/0037-builder-glci-interface.md
Gchbg: Yes, added it to the PR description. Let me know if it clarifies it.

docs/reference/adr/0037-builder-glci-interface.md
Gchbg: I'd argue that this is counterproductive to the idea of this ADR (and its predecessor). The idea is - here is the interface, here is every possible field, and here is what each means. Adding a placeholder would allow someone to shortcut extending the interface by encoding extra data in that placeholder, with implementation-defined semantics - which would be going back to where we were before this ADR.

docs/reference/adr/0037-builder-glci-interface.md
Gchbg: I amended the description. If you meant extending the ADR itself, let me know.

NotTheEvilOne (approved):
LGTM in general. I wonder if GLCI can be further improved to read the release file as all additional translation steps for S3 metadata may not depend on information not covered in the release file.

Gchbg:
> LGTM in general. I wonder if GLCI can be further improved to read the release file as all additional translation steps for S3 metadata may not depend on information not covered in the release file.

Not sure I fully understood the double-negative tbh. In general I'm all in favor of reducing intermediate steps, and I'm all in favor of clear interfaces. Sometimes those two pull in opposite directions ;) Happy to discuss possible enhancements.

NotTheEvilOne:
> > LGTM in general. I wonder if GLCI can be further improved to read the release file as all additional translation steps for S3 metadata may not depend on information not covered in the release file.
> 
> Not sure I fully understood the double-negative tbh. In general I'm all in favor of reducing intermediate steps, and I'm all in favor of clear interfaces. Sometimes those two pull in opposite directions ;) Happy to discuss possible enhancements.

What I meant is that: From an interface perspective it doesn't matter where the metadata required does come from (beside maybe some kind of "informational section"). The interface requires metadata keys to contain specific values. That's it.

gehoern (approved):

PR #5118, opened 2026-08-04
feat: reset `/etc` overlay on version update and add etc setup hooks

### LINKED ISSUE #5121: [ER] inplace update: have a hooks location that is not deleted during major update
### Garden Linux Version

2150

### Enhancement Type

New Feature

### Proposal

Inplace update already fullfills the feature of doing minor updates. For major updates we need to re-initialize the /etc directory. But in that case we loose all startup scripts and all details users have registered towards GL. Also the first init script of cloud providers might not be re-executed. So please introdure a hook-directory that remains even after /etc revamp.

`/var/lib/gardenlinux/etc-setup-hooks
`

A major update can be detected in the root disk if the os-release does not match that on the overlay to be mounted over /etc resets the overlay. So no extra caution needs to be taken in the update scripts etc. GL can detect this automatically (TODO: make sure the /etc/os-release is not only the symlink from /usr/share)

It is then supposed to be run all the installed hooks whenever the /etc overlay got wiped. The idea is that the initial user data script should install hooks there, rather than expect that whatever it modifies on /etc will survived a version bump.

TODO: We'll need to consider how the first update from a version that does not yet use these hooks to one that does will work. I.e. how a node that was originally provisioned with a user data script that did not install hooks can gain the hooks prior to updating to the new version.

### Motivation for Improvement

_No response_

### Code of Conduct

- [x] I confirm that I read and agree to this project's Code of Conduct.

nkraetzschmar:

When the `os-release` on the root disk does not match that on the overlay to be mounted over `/etc` resets the overlay.
This is to enable major version upgrades via in-place updates, but also just to make regular in-place updates more stable.

The `/var/lib/gardenlinux/etc-setup-hooks` is then supposed to be used to install hooks that will run whenever the `/etc` overlay got wiped. The idea is that the initial user data script should install hooks there, rather than expect that whatever it modifies on `/etc` will survived a version bump.

**TODO:** We'll need to consider how the first update from a version that does not yet use these hooks to one that does will work. I.e. how a node that was originally provisioned with a user data script that did not install hooks can gain the hooks prior to updating to the new version.


gehoern (approved):
run-etc-setup-hooks i would use run-parts to run executables in e directory

gehoern:
refer #5121

acumino:
>We'll need to consider how the first update from a version that does not yet use these hooks to one that does will work. I.e. how a node that was originally provisioned with a user data script that did not install hooks can gain the hooks prior to updating to the new version.

Can we not solve this by putting the condition that any new version can only be updated from a version which supports hooks?
For example, if 1877.10 supports hooks and 1877.9 doesn't, we can put the condition that 1877.11 can only be updated from 1877.10. This way we don't need to ensure that the old version somehow gets the hook.

nkraetzschmar:
> > We'll need to consider how the first update from a version that does not yet use these hooks to one that does will work. I.e. how a node that was originally provisioned with a user data script that did not install hooks can gain the hooks prior to updating to the new version.
> 
> Can we not solve this by putting the condition that any new version can only be updated from a version which supports hooks? For example, if 1877.10 supports hooks and 1877.9 doesn't, we can put the condition that 1877.11 can only be updated from 1877.10. This way we don't need to ensure that the old version somehow gets the hook.

The main question would be how to go from 1877.9 (no hooks) to 1877.10 (using hooks). Since 1877.10 would then on boot detect that the `/etc/os-release` in the overlay does not match => reset it and attempt to run the hooks, but there'll be no hooks present since that instance was originally brought up with a user-data script that did not install hooks. Or is the idea that the update 1877.9 => 1877.10 would have to be a non in-place one?

A better solution might be to have the gardener node agent install the hooks before triggering the in-place update to a version requiring hooks. Since just placing them on disk in a non hook version does nothing, but then it would be prepared for the next update.

PR #5129, opened 2026-08-06
fix: collapse multiline module docstrings in disastig tests for diki report

### LINKED ISSUE #5130: remove the newline \n in docstring headers for all test_disastig_NNNNN.py for correct issue generation
As discussed today with diki colleagues, the newline in our docstrings is causing the issue generation to be problematic as below

Image

saubhikdattagithub:
## Summary

- Module-level docstrings in `test_disastig_NNNNN.py` files were multiline, causing the diki **Rule Name** field to break out of the report table layout due to embedded newlines
- Fix: collapse all 113 module-level docstrings to a single line

## Before
```python
"""
Ref: SRG-OS-000043-GPOS-00022

Verify the operating system alerts the ISSO and SA (at a minimum) in the event
of an audit processing failure.
"""
```

## After
```python
"""Ref: SRG-OS-000043-GPOS-00022 

Verify the operating system alerts the ISSO and SA (at a minimum) in the event of an audit processing failure."""
```

Fixes #5130


tests/integration/security/compliance/test_disastig_00216.py
```diff
@@ -1,6 +1,4 @@
-"""
-Ref: SRG-OS-000471-GPOS-00216
-"""
+"""Ref: SRG-OS-000471-GPOS-00216"""
```
Akendo: Don't we miss here something?

tests/integration/security/compliance/test_disastig_00152.py
```diff
@@ -1,6 +1,4 @@
-"""
-Ref: SRG-OS-000365-GPOS-00152
-"""
+"""Ref: SRG-OS-000365-GPOS-00152"""
```
Akendo: Is here also something missing?

Akendo (changes requested):
We might miss some description in two tests. @saubhikdattagithub can you check on that if my assumption here is incorrect?

tests/integration/security/compliance/test_disastig_00216.py
saubhikdattagithub: Yes 00216.py now has the Title in the docstring (was missing)

tests/integration/security/compliance/test_disastig_00152.py
saubhikdattagithub: Yes 00152.py now has the Title in the docstring (was missing)

saubhikdattagithub:
> We might miss some description in two tests. @saubhikdattagithub can you check on that if my assumption here is incorrect?

Yes, thanks for the closer look. Both tests are now addressed @Akendo 
Also 00193.py was updated accordingly

Akendo (approved):

PR #5151, opened 2026-08-11
fix: publish to gardenlinux-fips

5kt:
Publish to gardenlinux-fips and not gardenlinux/gardenlinux-fips

Fixes #


5kt:
Closing, should in fact go under gardenlinux/gardenlinux/fips

PR #5153, opened 2026-08-11
fix: use gardenlinux/fips instead of gardenlinux/gardenlinux-fips

5kt:
Publish to gardenlinux/gardenlinux/fips instead of gardenlinux/gardenlinux-fips

Fixes #


NotTheEvilOne (approved):
LGTM. While I do not like the current not scaling list of flavors it's fine for the time being. If more flavors follow a new behavior and logic should be implemented.

PR #5210, opened 2026-08-20
Disable the issue creation on nightly failure

5kt:
The auto creation of an issue when the nightly fails should be disabled for the time being. It is not working as expected, the issue is created on first failure and most of the time on 2nd/3rd retry the nightly will be green.
This is creating a lot of issues that shouldn't be created.

Fixes #

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -128,36 +128,3 @@ jobs:
       actions: write # TODO: 3215 elevated permissions for included workflow
       id-token: write # TODO: 3215 elevated permissions for included workflow
     secrets: inherit
-  create_issue_on_failure:
-    name: Create issue on failure
-    needs:
-      - build
-      - test
-      - upload_workflow_data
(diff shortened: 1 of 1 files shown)


sh4sm (approved):

yeoldegrove (approved):
To make this if statement not even more confusing, maybe just comment out the whole step?

.github/workflows/nightly.yml
```diff
@@ -136,7 +136,11 @@ jobs:
       - upload_workflow_data
       - publish_ghcr
       - publish_s3
-    if: ${{ always() && failure() }}
```
NotTheEvilOne, line 139: As already discussed on [#5009](https://github.com/gardenlinux/gardenlinux/pull/5009#discussion_r3560954258) this has been the wrong place for integration anyway. I would recommend to remove it completely as well.

NotTheEvilOne (approved):

PR #5216, opened 2026-08-24
feat: enable extra fips flavors (azure usi arm64 and gcp fips usi)

5kt:
After checking with @tmang0ld @Akendo and @gehoern, these extra flavors should be enabled.


Akendo (approved):

Akendo:
Needs to be backport to 1877

PR #5275, opened 2026-09-09
Fix the flavors mess

Gchbg:
This fixes the fips flavors mess that was in flavors.yaml

**Definition of Done:**

**Release note**:

```bugfix user
Fix FIPS flavors matrix
```


Vincinator (approved):
ali and opesntack fips are dropped. Matches what we have on 1877 branch, so /lgtm

PR #5278, opened 2026-09-10
feat: create_nightly_issue.py: add test results

yeoldegrove:

This adds the actual test results to the nightly issue which is a nice information that was previously missing.
It also fixes a pagination issue while retrieving the job logs and makes sure the issue body does never hit GitHubs limit.

## tests

Tested and verified with https://github.com/gardenlinux/gardenlinux/issues/5277

### before

image

### after

image

### DIFF
--- a/hack/create_nightly_issue.py
+++ b/hack/create_nightly_issue.py
@@ -14,17 +14,158 @@ Usage:
 """
 
 import argparse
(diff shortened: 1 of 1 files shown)


vivus-ignis (approved):
awesome, thank you!

